PR #8, opened 2016-08-24
auto closing pair ` ` instead of ' '

innerlee:
`` ` ` `` are used to enclosing commands, while `` ' `` means transpose.


davidanthoff:
Thanks! Might take a while to release a new version of the extension, quite busy right now.

PR #13, opened 2016-10-01
Completion evaluation

novatena:
Features:
+ repl based autocompletion. Also the same as working with REPL.
+ execute julia code via command "julia: Execute Code" on command prompt. This can be used, for example to import modules for auto-completion and to added user defined types and functions into julia for autocompletion on user file to work.

Next:
- julia must be in PATH -> to be improved
- still no feedback when perform code execution -> to be added.
- goto definition may be implemented by using @less or @functionloc macros in the REPL.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
             "properties": {
                 "julia.validate.executablePath": {
                     "type": ["string", "null"],
-                    "default": "/home/nova/Softwares/julia-0.5/bin/julia",
+                    "default": "null",
                     "description": "Points to the julia executable."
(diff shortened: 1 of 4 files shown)


davidanthoff:
I think we are going to use the [language server protocol](https://github.com/Microsoft/language-server-protocol) going forward for things like a completion provider.

I hope to reactive the REPL code that I started creating in #2 soon, and that should provide the send to REPL command.

If you have more providers that could be part of our language server, any PR would be greatly appreciated!

PR #14, opened 2016-10-30
Add typescript providers

ZacLN:
Adds TS providers that talk to a single Julia process for:
- Hovers
- Documentation
- Completions
- Linting
- Function definitions

Only tested on Linux w/ vscode 1.6.1 and recent Julia version.

I've done it with a fairly even TS/Julia code mix as implementing the pure julia language-server  meant that one couldn't provide these features for unsaved files (i.e. the client-server interface will only transfer file location details not individual snippets of code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
     "name": "language-julia",
     "displayName": "julia",
     "description": "julia language support",
-    "version": "0.4.1",
+    "version": "0.4.2",
     "publisher": "julialang",
     "engines": {
         "vscode": "^1.0.0"


davidanthoff:
Oh, one more thing: are you the creator of this code? I saw some emails that suggested that someone copied the code from somewhere, but it was not clear to me whether you (@ZacLN) are the person who did the copying or the original author of the code.

ZacLN:
Ok, alot to reply to. I'm inexperienced with contributing so I thought I'd dump out what I had first so that you could decide how to incorporate it/h

davidanthoff:
Yeah, no worries, this is actually useful because it gives me a really good overview where this is going! Plus it is just fantastic how much you created here. Happy to work together to get this merged eventually.

ZacLN:
Sorry, pressed send by mistake, to continue:
 
- Re code origins: nova/novatena2015 appeared to have an updated fork which has since disappeared. This had the julia proxy back-end and code completion,hovers and definition goto. This build on the previous package.
- Re language server vs TS/julia mix: My main issue with the pure language-server approach was that working on unsaved files would not work and that working with unsaved changes would require some short term version control system on the Julia side. Though doable this seems to require quite a bit more work for little practical gain (I tried this initially because I thought it a more elegant solution). Client side requests for things like hover/completions just send selection position data.
- In terms of package setup: I don't have access to Mac/Windows systems so would need some OS specific help in how to do this across platforms. As an initial approach would you think a fixed path in the settings contribution, similar to that given for the package location,  appropriate?

Once you give me an idea of how you think it best to progress with all this I'll start again with smaller PRs

Ta, Zac

jl/server.jl
```diff
@@ -0,0 +1,48 @@
...
+    "symbols" => symbolrequesthandler,
+    "diagnostics" => diagnosticshandler
+)
+
+function run()
+    while true
+        request = nothing
+        ok = false
+        try
+            request = JSON.parse(readline())
```
ZacLN: This doesn't seem to matter: things printed to STDOUT not fitting the expected form of a ?Result don't get picked up by the TS side of things.

jl/utils.jl
```diff
@@ -0,0 +1,189 @@
...
+    str = code[1:pos]
+    return split(str, '\n')
+end
+
+function isimportcmd(cmd::AbstractString)
+    return startswith(cmd, "using") || startswith(cmd, "import")
+end
+
+function doimport!(cmdcache::Set{AbstractString}, filename::AbstractString)
+    lines = readfirstlines(filename, 20)
```
ZacLN: This is code I didn't write/change. I was going to change it to simply search for all instances of using/import/include to ensure all relevant modules get included into the namespace.

src/async.ts
```diff
@@ -1,187 +0,0 @@
-/*---------------------------------------------------------------------------------------------
```
ZacLN: This was used for the previous linter, not sure of it's exact role but it seems easier to keep all the components running on one process?

PR #15, opened 2016-10-31
Add server and hoverprovider

ZacLN:
This rewrites the server backend simplifying the interface somewhat and uses a named pipe rather than STDIO.

I've rewritten the server side to use a pipe as was suggested. This should prevent any possible interference of modules printing output on load and may allow some sort of direct interface with server to provide output in the future.

The interface between various providers and the server is now more modular, the server simply receives a request and sends back the result without doing any parsing of that result, leaving it to the (in this case) JuliaHoverProvider class to do what it will with the returned data.

There may be some drawback to this approach but I can't see them atm.

### DIFF
--- a/jl/server.jl
+++ b/jl/server.jl
@@ -1,0 +1,22 @@
+using JSON
+server = listen("juliaserver"*string(getpid()))
+
+hover(sock,request) = write(sock,JSON.json(Dict(
+                        "id"=>request["id"],
+                        "type"=>"hover",
+                        "doc"=>string(Docs.doc(getfield(Main,Symbol(request["params"]))))
+                    )))
+
+while true
+    sock = accept(server)
+    @async while isopen(sock)
+        msg = readline(sock)
+        try 
+            request = JSON.Parser.parse(msg)
+            if request["type"] == "hover"
+                hover(sock,request)
+            end
+        end
+    end
+end
+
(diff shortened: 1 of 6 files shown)


davidanthoff:
I played around with the language server protocol, and it seems super simple and straightforward. I have a very rough prototype of the linter using it here #16.

So I think what we should do is implement all of this as a language server. I see two major benefits: all of this would also work in Eclipse, and pretty much the whole code could stay in julia, which also makes things a lot easier. Does that sounds good to you?

What I need to do is clean up #16 a bit. I should actually be able to reuse the whole package installation, julia config stuff etc. that is already there. And then once I have that done, you could integrate all the cool features you coded up as additional things the language server provides?

davidanthoff:
I should say, I should have told you that I was looking into this. Sorry if this created even more work... But this weekend and today were just busy, and I was glad when I found a few minutes to actual try things out.

ZacLN:
No problemo, I'll hold off on continuing until we see where you are with this.

Incidentally, did you manage to get any of the features working with unsaved files?

davidanthoff:
Yes, that was the main thing I wanted to investigate. The server gets a new ``textDocument/didChange`` every time a user types anything, even if the file is not saved. The payload of that message is the editor buffer. Right now I use a pretty inefficient version of that where the whole file gets transmitted at every change, but it should be pretty simple to change that over to the incremental version. At the end of the day I just have a ``Dict`` in the server, where the key is the filename, and the value is the file content. And then I update that ``Dict`` whenever I get any of the messages that relate to content changes.

ZacLN:
Fantastic. I'll leave this alone now then.

PR #16, opened 2016-10-31
WIP Language server

davidanthoff:


ZacLN:
Ok, I couldn't get this up and running but can see the idea.

When I tried this before I had a structure for taking requests along the lines of: [https://gist.github.com/ZacLN/d84fd47540d83303ab6b34e10bab6525](url)

This means that adding methods just requires a new method type, parameter type and appropriately parameterised `Respond` function keeping things pretty clean and meaning you don't have to mess around with endless `Dicts` (and coincidentally making it much easier to understand when you have vscode showing you the structure).

Do you think this would be a decent way to incorporate the various providers?

davidanthoff:
@ZacLN Yes, that is definitely a better design than what I've done so far.

I was mainly trying to get the whole process management etc. up and running. I think I'm pretty close with that, and once that is merged you could start to a) change the server code itself to something that looks more like the gist you pointed to and b) incorporate more providers?

Could you try the current version on this branch? I think the linter should work now, plus the whole installation of the packages that are needed etc.

The one thing that is still missing a detection that a setting has changed, and then a restarting of the language server.

My suggestion right now would be this: if this version works on your machine, I'll merge it into master. At that point you can start to create PRs that modify the julia code, i.e. both use a better structure and add more providers. I can in parallel try to get the setting modification stuff more robust. I think those two things we could work on in parallel, because one should only change julia code and the other should only change the typescript code, right?

davidanthoff:
Oh, and I'm back to communicating via STDIN and STDOUT, otherwise one can't have the language client package in VS Code manage the process. But I managed to redirect the standard STDOUT, so that any julia code that just writes ``println`` actually goes somewhere else. If you want to print some debug info in the julia code, just use ``info("whatever")``, and that will show up in the output window of the running VS Code instance.

ZacLN:
Excellent, that sounds great.
I've got initialisation response, hovers, goto, completions and signature helpers setup with toy messages as input. If it's alright with you I'll work from the point that

ZacLN:
OK, the server seems to be up and running. Incidentally, is there any particular reason for limiting the Julia version below 0.6?

davidanthoff:
Ok, I've squashed and merged this, so now we can work off master. I also removed the limit re julia 0.6, no good reason for that. Very much looking forward to your PRs!

davidanthoff:
Oh, and I just added an integrated terminal. Was super easy, check it out!

PR #18, opened 2016-11-02
Add hover provider

ZacLN:
Adds the basic hover provider.

I've run into problems already with STDIO, it can't seem to send large messages (~1500 characters) without breaking so I've had to limit doc returns to the first 20 lines.

### DIFF
--- a/scripts/languageserver/transport.jl
+++ b/scripts/languageserver/transport.jl
@@ -18,7 +18,8 @@ function read_transport_layer(stream)
 end
 
 function write_transport_layer(stream, response)
-    n = length(response)
+    response_utf8 = transcode(UInt8, response)
+    n = length(response_utf8)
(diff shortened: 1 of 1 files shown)


davidanthoff:
Cool!

I think I just fixed the error with the large messages that you ran into [here](https://github.com/JuliaEditorSupport/julia-vscode/commit/c4ccc0573680b361b8801a83723e820cd3841935). Essentially I was incorrectly encoding the length of the messages. Can you try whether things work without truncating large messages now?

And it would be a lot easier if you could create a branch other than ``master`` for your commits :)

But at least the hover provider works on my machine!

davidanthoff:
And the completions also seem to work. I'm really excited about all of this, great work!

ZacLN:
Oh man, I did it again... I'll shift off master, appologies

PR #27, opened 2016-11-04
WIP JSON-RPC refactoring

davidanthoff:
@ZacLN This is the beginning of a slight refactoring of the JSON-RPC layer. Not done yet, but I wanted to give you a heads up and get any feedback you might have.

The general idea is to a) use ``Val{Symbol("nameofmethod")}`` to dispatch on different JSON-RPC messages and b) use multi dispatch for the initial parsing of incoming messages. Plus, we should be able to eventually move the basic JSON-RPC infrastructure into its own package.


ZacLN:
Yep, sounds great. I was thinking that I've actually built in one extra unneeded layer processing which this should remove?

davidanthoff:
Ok, the slight refactoring turned into a big one ;)

Here are some key points:
- I got rid of the global variables, there is now a ``LanguageServer`` type. This is mostly in prep of moving things into a package, and then it should be a lot easier to e.g. write tests etc.
- I've renamed ``Respond`` to ``process``, that seemed more appropriate, given that the server is actually not responding to some messages.
- I've removed the ``try`` and ``catch`` blocks in the ``process`` method. They were really hiding errors that we might make. I think we should not have such a catch-all thing, but rather only return an error response if we detect a well known problem. But if there is a bug in our code, the server should just crash, otherwise those bugs will just go undetected for a long time.
- The main flow is slightly different. ``process`` no longer returns anything, instead one can just send a response from within ``process``, or not.

davidanthoff:
There is one final general change I think we should make: I would prefer if the actual work to respond to some message all happened in the ``process`` method, and not in various constructor calls. I think that would make things easier and cleaner to understand. So essentially all the types that represent data structures from the protocol definition would be defined in ``protocol.jl``, and each type would have only two constructors: one that takes a ``Dict``, and another one that takes all the values of its fields. But any processing, logic etc. would not live in these constructors.

ZacLN:
Yep I think that all makes sense, I'll have a look tomorrow to get my head round it.

ZacLN:
For the latter part
```
ProtocolTypes = [Position
                 Range
                 Location
                 TextDocumentIdentifier
                 TextDocumentItem
                 etc..]

defs = Expr[]
for T in ProtocolTypes
    tname = Symbol(T)
    fn = fieldnames(T)
    ft = collect(T.types)
    args = Expr[]
    for (n,t) in zip(fn,ft)
        if in(t,ProtocolTypes)
            push!(args,Expr(:call,Symbol(t),Expr(:ref,:d,"$n")))
        else
            push!(args,Expr(:ref,:d,"$n"))
        end
    end
    
    f = Expr(:(=),Expr(:call,tname,Expr(:(::),:d,:Dict)),Expr(:call,tname,args...))
    eval(languageservermod,f)
end
```
should do the trick concisely

davidanthoff:
That is cool! I'll merge my stuff first, but then I'll add that.

PR #28, opened 2016-11-04
Incremental update

ZacLN:
@davidanthoff I've added a rudimentary partial update which _seems_  to work without problems. I couldn't think of any complicated corner cases beyond multiple cursor pasting to trick it up but you may want to have a test yourself before merging.

### DIFF
--- a/scripts/languageserver/provider_misc.jl
+++ b/scripts/languageserver/provider_misc.jl
@@ -34,7 +34,7 @@ type InitializeResult
 end
 
 const serverCapabilities = ServerCapabilities(
-                        TextDocumentSyncKind["Full"],
+                        TextDocumentSyncKind["Incremental"],
                         true, #hoverProvider
                         CompletionOptions(false,["."]),
                         true, #definitionProvider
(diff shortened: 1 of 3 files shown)


ZacLN:
Incidentally we lose a bit of speed converting from `SubString` to `Vector{UInt8}`  and back again so we should probably shift to holding documents simply as raw vectors. The immutability of `String`s causes all sorts of trouble

scripts/languageserver/provider_misc.jl
```diff
@@ -75,10 +75,36 @@ end
...
+        end
     try
-        documents[r.params.textDocument.uri] = split(r.params.contentChanges[1].text, r"\r\n?|\n")
         return
     catch err
         return Response{didChange,Exception}("2.0",r.id,err)
     end
 end
+
+function (c::TextDocumentContentChangeEvent)(vtdi::VersionedTextDocumentIdentifier)
```
davidanthoff: Could we make this a normal function? This just seems to complicate things...

scripts/languageserver/provider_misc.jl
```diff
@@ -75,10 +75,36 @@ end
...
+    d = join(documents[vtdi.uri],'\n').data
```
davidanthoff: I think this is not very efficient. Why put everything together into one large string, if we can just modify those lines that are actually changed? I think the idea would be that most of these change requests probably only affect only one line, so those change requests can be handled very efficiently because you only need to touch one element in the array that holds the strings for a document.

ZacLN: I think I was working under the misapprehension that split resulted in a vector of `SubString`s, I'll have another look

scripts/languageserver/provider_misc.jl
ZacLN: yep, sure. My whole repo has gone wonky so I'm going to close this and start afresh

scripts/languageserver/provider_misc.jl
davidanthoff: Ah, you are right! Well, but what we could do is simply start out with an array of ``SubString``s, and then as we receive change notifications, we could replace those ``SubString``s with real ``String``s for those lines that get modified?

scripts/languageserver/provider_misc.jl
davidanthoff: Sorry, that was probably partly my mistake by changing so much...

scripts/languageserver/provider_misc.jl
ZacLN: Actually, looking at it the type specification on the document Dict seems to force a conversion when we open a file

ZacLN: I don't know much about String handling tbh, I've always had the impression it's all quite slow in Julia?. I'm going to do this properly (tediously) and set up and compare 4 approaches (full sync, Vector{UInt},, Vector{String} and Vector{SubString})

davidanthoff: ``Vector{UInt}`` would probably be tricky, then we would have to do the whole UTF-8 handling ourselves... Do we know whether this part is actually a real bottleneck? I would be kind of surprised. Right now even the full sync performs pretty well on my system, so I would think if we were to move to a ``Vector{AbstractString}`` thing, it could only get better.

ZacLN: Ok I'll just do that and look into it later

PR #36, opened 2016-11-05
signatures run through arguments and don't crash

ZacLN:
fixes signatures and I uncovered a bug in get_word that caused hovers to crash when hovering brackets and other special characters


davidanthoff:
Great!

PR #40, opened 2016-11-07
Clean up hovers

ZacLN:
cleans up the hovers provider. Also replaces `get_word` which was crashing before on encountering unicode characters


ZacLN:
Pulled from wrong branch

PR #41, opened 2016-11-07
extend snippets

ZacLN:


snippets/snippets.json
```diff
@@ -8,5 +8,55 @@
 		],
 		"description": "For Loop",
 		"scope": "source.julia"
+	},
+	"While Loop": {
+		"prefix": "while",
+		"body": [
+			"while ($1)",
```
davidanthoff: Why parenthesis here?

snippets/snippets.json
```diff
@@ -8,5 +8,55 @@
...
+			"\t$2",
+			"end"
+		],
+		"description": "Immutable declaration",
+		"scope": "source.julia"
+	},
+	"Function Declaration": {
+		"prefix": "function",
+		"body": [
+			"function ($1::$2)",
```
davidanthoff: I don't think this should be the default, because a) there might be functions with no arguments and b) it is generally better in julia to actually not type function arguments.

snippets/snippets.json
```diff
@@ -8,5 +8,55 @@
...
+			"\t$3",
+			"end"
+		],
+		"description": "Function declaration",
+		"scope": "source.julia"
+	},
+	"Macro Declaration": {
+		"prefix": "macro",
+		"body": [
+			"macro ($1)",
```
davidanthoff: This seems the wrong syntax, shouldn't there be a field for the macro name first?

snippets/snippets.json
ZacLN: Yep, not the normal style

snippets/snippets.json
ZacLN: OK, and there was no placeholder for a name.

PR #44, opened 2016-11-11
Incremental doc update

ZacLN:
Will allow us to do partial updates of expensive static code analysis only on regions that have changed.


davidanthoff:
This now has a merge conflict, probably because of the other PRs I just merged. Can you resolve that?

Also, this does handle UTF8 stuff correctly, because you use ``read`` with the ``Char`` parameter, right? I.e. it will handle characters correctly that take more than on ``UInt8`` space? That is how I understood your good, just want to double check.

ZacLN:
Aye, that's right. I'm going to close/ open new PR.

PR #47, opened 2016-11-13
Wait until REPL is started in send code command

davidanthoff:
@ZacLN Could you try this on your system? The general idea is that you can send a line of code from the editor to the REPL via Ctrl+Enter. This PR here should make sure that this works even if you haven't started the julia REPL before you send the first bit of code to the REPL.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -218,19 +218,36 @@ function executeJuliaCodeInREPL() {
         }
     }
 
-    // TODO Somehow wait for the named pipe to be ready if we started the julia process here.
     startREPLCommand();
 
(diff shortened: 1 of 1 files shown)


ZacLN:
Nope this still fails to execute the first sent command I'm afraid

ZacLN:
I get that one

davidanthoff:
Ah, then it just takes longer on your system to start the julia process and get it ready to receive commands. I guess at 50 we are waiting for 5 seconds, which still seems ok, so I'll just change the limit to that. Thanks for testing!

PR #49, opened 2016-11-13
Only lint files in workspace

ZacLN:
Adds rootPath to LanguageServer type

### DIFF
--- a/scripts/languageserver/provider_misc.jl
+++ b/scripts/languageserver/provider_misc.jl
@@ -10,7 +10,7 @@ const serverCapabilities = ServerCapabilities(
                         SignatureHelpOptions(["("])) 
 
 function process(r::JSONRPC.Request{Val{Symbol("initialize")},Dict{String,Any}}, server)
-    server.rootPath=haskey(r.params,"rootPath") ? "file://"*r.params["rootPath"] : ""
+    server.rootPath=haskey(r.params,"rootPath") ? r.params["rootPath"] : ""
     response = Response(get(r.id),InitializeResult(serverCapabilities))
     send(response, server)
 end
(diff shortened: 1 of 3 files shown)


ZacLN:
@davidanthoff, I'm unsure of the rules we actually want for this. For example I'd want  to allow this for all files

scripts/languageserver/provider_misc.jl
```diff
@@ -10,6 +10,7 @@ const serverCapabilities = ServerCapabilities(
                         SignatureHelpOptions(["("])) 
 
 function process(r::JSONRPC.Request{Val{Symbol("initialize")},Dict{String,Any}}, server)
+    server.rootPath=haskey(r.params,"rootPath") ? "file://"*r.params["rootPath"] : ""
```
davidanthoff: Maybe not use the ``file://`` prefix here, after all this is not a file URI? Maybe just store the path as we get it?

scripts/languageserver/provider_misc.jl
```diff
@@ -21,7 +22,9 @@ end
 function process(r::Request{Val{Symbol("textDocument/didOpen")},DidOpenTextDocumentParams}, server)
     server.documents[r.params.textDocument.uri] = split(r.params.textDocument.text,r"\r\n?|\n")
     
-    process_diagnostics(r.params.textDocument.uri, server)
+    if r.params.textDocument.uri[1:min(length(r.params.textDocument.uri),length(server.rootPath))] == server.rootPath
```
davidanthoff: I'm wondering what the right behavior is when ``rootPath`` is ``""``... We probably still want to lint in that case? Maybe we actually need to rules: 1) if a folder is open, only lint files from that folder, 2) if no folder is open, lint any file, except files from the base library? I'm not sure, though...

ZacLN:
couldn't fix conflicts so I've opened a new one.

PR #55, opened 2016-11-16
return string for completion docs

ZacLN:
Documentation for completions need to a strings not markedstrings

### DIFF
--- a/scripts/languageserver/provider_completions.jl
+++ b/scripts/languageserver/provider_completions.jl
@@ -8,7 +8,9 @@ function process(r::Request{Val{Symbol("textDocument/completion")},TextDocumentP
         s = get_sym(i)
         d = ""
         try 
-            d = join(get_docs(s)[2:end],'\n')
+            d = get_docs(s)
+            d = isa(d,Vector{MarkedString}) ? (x->x.value).(d) : d
+            d = join(d[2:end],'\n')
         end
         kind = 6
         if isa(s,String)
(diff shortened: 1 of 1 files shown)


scripts/languageserver/provider_completions.jl
```diff
@@ -8,7 +8,9 @@ function process(r::Request{Val{Symbol("textDocument/completion")},TextDocumentP
         s = get_sym(i)
         d = ""
         try
```
davidanthoff: Actually, can we get rid of these ``try`` blocks? If there is an error, we should explicitly check for it and not rely on exception handling. And this kind of ``try`` block just hides bugs, imho.

ZacLN:
I've fixed this for now, if you spot any inconsistencies between hover docs and completion docs  let me know and I'll have a look.

scripts/languageserver/provider_completions.jl
```diff
@@ -7,9 +7,11 @@ function process(r::Request{Val{Symbol("textDocument/completion")},TextDocumentP
     CIs = map(comp) do i
         s = get_sym(i)
         d = ""
-        try 
-            d = join(get_docs(s)[2:end],'\n')
-        end
+        d = get_docs(s)
+        d = isa(d,Vector{MarkedString}) ? (x->x.value).(d) : d
+        d = join(d[2:end],'\n')
+        d = replace(d,'`',"")
```
davidanthoff: Hm, this doesn't seem very robust. Couldn't we use the functionality [here](https://github.com/JuliaLang/julia/blob/master/base/markdown/render/plain.jl) to render the markdown as plain text, and then return that?

ZacLN: This doesn't strip out the markdown punctuation as far as I know. I think `string` calls `plain` somewhere up the line

PR #61, opened 2016-11-16
Paste code to REPL

davidanthoff:
This uses a different strategy for sending code to the REPL. It still is triggered by Ctrl+Enter.

I'm leaving the old core around as a comment, in case we figure out how to solve the julia prompt issue, I think it would be a more elegant solution. But for a v0.5 release, this should do.

@ZacLN I'll no longer commit anything to master. I'll submit everything as PRs, and I'll wait until you have reviewed everything, ok? Seems better if we follow that workflow for now, four eyes spot more errors than two.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -179,12 +179,15 @@ function openPackageDirectoryCommand() {
 }
 
 function startREPLCommand() {
-    
+    startREPL();
+    REPLterminal.show();
+}
+
+function startREPL() {
     if (REPLterminal==null) {
         let args = path.join(extensionPath, 'scripts', 'terminalserver', 'terminalserver.jl')
         REPLterminal = vscode.window.createTerminal("julia", juliaExecutable, ['-q', '-i', args, process.pid.toString()]);
     }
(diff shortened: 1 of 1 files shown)


davidanthoff:
I think I'd actually prefer to keep them completely separate as two processes... If the language server is really a static analysis tool, then that should work, right?

Also, for example if someone starts a long running piece of code in the REPL, it would be odd if all the language server features stopped working etc., so my tendency is to just keep those two things separate.

ZacLN:
Aye, I completely agree I was thinking more of some possible inter process communication that updated the server about the 'state' of the REPL process but I'm not sure if thats possible or indeed - as you say - needed.

davidanthoff:
Ah, I hadn't thought about that... I think that is a good idea down the road, i.e. that the REPL process sends info to the language server so that it can provide better feedback. But probably something we could do way, way later, right? My sense is there is a lot we can do with static analysis for now.

ZacLN:
Definitely, let's come back to this

PR #63, opened 2016-11-16
Doc sync

ZacLN:
Ok, this implements incremental doc syncing. It's bunched together with a new `Document` type that at the moment stores the document as a `Vector{UInt8}` to allow mutability and stores the parsed document in different blocks (i.e. function defs, etc.). I've added this second bit as it makes it easier to then add the features that rely on parsing bit by bit.


scripts/languageserver/parse.jl
```diff
@@ -0,0 +1,243 @@
+type VarInfo
+    t
```
davidanthoff: Could you either use self-explanatory names, or add a comment in a case like this that says what this field stores?

scripts/languageserver/parse.jl
```diff
@@ -0,0 +1,243 @@
...
+end
+
+function Block(utd, ex, r::Range)
+    t, name, doc, lvars = classify_expr(ex)
+    ctx = LintContext()
+    ctx.lineabs = r.start.line+1
+    dl = r.end.line-r.start.line-ctx.line
+    Lint.lintexpr(ex, ctx)
+    diags = map(ctx.messages) do l
+        return Diagnostic(Range(Position(r.start.line+l.line+dl-1, 0), Position(r.start.line+l.line+dl-1, 100)),
```
davidanthoff: The columns here are incorrect. #60 fixes that in the current version, maybe you can just adapt that code here?

scripts/languageserver/parse.jl
```diff
@@ -0,0 +1,243 @@
+type VarInfo
+    t
+    doc::String
+end
+
+type Block
```
davidanthoff: Could you add a comment that describes what a ``Block`` is? It is not clear to me right now :)

scripts/languageserver/parse.jl
```diff
@@ -0,0 +1,243 @@
...
+                        string(l.code),
+                        "Lint.jl",
+                        l.message) 
+    end
+    v = VarInfo(t, doc)
+
+    return Block(utd, ex, r, name,v, lvars, diags)
+end
+
+function Base.parse(uri::String, server::LanguageServer, updateall=false)
```
davidanthoff: I think we should add a method to ``Base.parse`` here, because ``Base.parse`` is assumed to return something. So maybe just find some other name for this?

scripts/languageserver/parse.jl
```diff
@@ -0,0 +1,243 @@
...
+    doc = String(server.documents[uri].data)
```
davidanthoff: Might make sense to factor this out into a helper function? I bet getting a specific doc as a string will be used a lot... Not urgent, though.

scripts/languageserver/parse.jl
```diff
@@ -0,0 +1,243 @@
...
+        i0 = i1 = 1
+        p0 = p1 = Position(0, 0)
+        out = Block[]
+        i4 = 0
+    else
+        i4 = findnext(b->b.uptodate, server.documents[uri].blocks,i)
+        p0 = p1 = server.documents[uri].blocks[i].range.start
+        i0 = i1 = linebreaks[p0.line+1]+p0.character+1
+        out = server.documents[uri].blocks[1:i-1]
+    end
```
davidanthoff: Maybe add some comments here? Really difficult to understand what exactly is going on... Would probably be enough to have one comment at the top of the block that gives a hint what this is about, and then maybe comments for the various vars that describe what they mean.

scripts/languageserver/parse.jl
ZacLN: Good call.. I'm going to with parseblocks for now but I'm willing tp losing the whole `Block` name!

scripts/languageserver/parse.jl
ZacLN: The lines are wrong as well... :( I'm going to disable this for now until I can get it all lined up as it's not publishing these diagnostics anyway .

scripts/languageserver/parse.jl
ZacLN: I'm going to leave this as is for now

PR #65, opened 2016-11-17
Add debug mode option to language server

davidanthoff:
This should hide the raw JSON traffic if the extension is not running under the debugger.


davidanthoff:
Yes, I think any error should still appear in the julia language server output channel in VS Code, this option just makes that output less verbose.

We might think about logging things to a file at some point, so that we can more easily diagnose error reports that we might get from users.

PR #69, opened 2016-11-17
Rename Range fieldname

ZacLN:
This is ... slightly dodgy but I offer it as a suggestion for fixing the linting issue.

Maybe the lesser of two evils (the other being the `eval` of modified Range type definition).

### DIFF
--- a/scripts/languageserver/protocol.jl
+++ b/scripts/languageserver/protocol.jl
@@ -12,6 +12,14 @@ type Range
     stop::Position # mismatch between vscode-ls protocol naming (should be 'end')
 end
 
+function JSON._writejson(io::IO, state::JSON.State, a::Range)
+    Base.print(io,"{\"start\":")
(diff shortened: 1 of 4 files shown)


scripts/languageserver/protocol.jl
```diff
@@ -7,13 +7,17 @@ Position(d::Dict) = Position(d["line"], d["character"])
...
-    ex.args[3].args=ex.args[3].args[[2;4]]
-    ex.args[3].args[2].args[1]=Symbol("end")
-    eval(ex)
+type Range
+    start::Position
+    stop::Position # mismatch between vscode-ls protocol naming (should be 'end')
+end
+
+function JSON._writejson(io::IO, state::JSON.State, a::Range)
+    Base.print(io,"{\"start\":")
```
davidanthoff: We don't need these ``Base.`` parts here, right?

PR #79, opened 2016-11-18
remove JuliaParser dependency

ZacLN:
I started to have a look at forking JuliaParser to have it store state and line/column info at intermediate stages of parsing and think it can be done but for now we may as well strip these dependencies?

I've also fixed `get_word` to handle cases such as `321sin(x)`.

### DIFF
--- a/scripts/languageserver/main.jl
+++ b/scripts/languageserver/main.jl
@@ -7,12 +7,10 @@ end
 
 include("dependencies.jl")
 use_and_install_dependencies([
-    ("AbstractTrees", v"0.0.4"),
     ("Compat", v"0.9.4"),
     ("JSON", v"0.8.0"),
     ("Lint", v"0.2.5"),
-    ("URIParser", v"0.1.6"),
-    ("JuliaParser",v"0.7.4")])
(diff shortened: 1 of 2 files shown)


davidanthoff:
Yes, lets take out for now. I've been reading up more and more about the C# Roslyn design and I think that is an almost ideal model for how to handle things. Ideally we could improve JuliaParser into that direction without forking it... Once we have the LS server repo (I'll try to create it today), I'll post some links about the Roslyn design and the data structures they use, which all seems really clever.

ZacLN:
I'll have a read up on the Roslyn framework sometime over the weekend. From my quick summary look, JuliaParser already has a structure that we can work with without creating a whole new `Expr` type holding location specific information. I think we can get what we want just by extending the `ParserState`type and finding the points in parsing that we are interested in, e.g.: on hitting this [this](https://github.com/JuliaLang/JuliaParser.jl/blob/master/src/parser.jl#L962) we capture all  [declarations](https://github.com/JuliaLang/JuliaParser.jl/blob/master/src/parser.jl#L363), etc. until that expression closes.

...but it seems to be quite complicated

Issue #58, opened 2016-11-16
v0.5 release planning


davidanthoff:
@ZacLN I think it would be good if we released a v0.5 soonish, and then moved the whole language server code into its own package for a VS Code extension v0.6 release.

One big reason for moving the language server into a package is that we can then start to setup a whole testing infrastructure for it, which seems really key.

So I'm trying to triage right now what should go into v0.5, and what we should do in later releases. Maybe we can use this issue here to discuss that.

The one PR that I'm unsure about is #54. Right now I can't really test it because #57 is blocking things for me right now. But you also seemed to think that #54 is maybe a bit flaky? If so, should we wait with it until v0.6, or do you think it is just a small extra work to get it into a stable state?

I'm also still fighting with #31... If I can't find a good solution soon, I might just remove that functionality for now...

davidanthoff:
@ZacLN I just gave you write access to this repo, so that you can review PRs formally.

I would suggest we both still keep our branches in our own forks, i.e. we both never push any branch to this repo here, but instead push to our individual forks and then open PRs.

I've also configured things such that at least either you or me needs to approve each PR before we can merge it.

ZacLN:
That's great @davidanthoff, thanks. 

As far as the release: the incremental document sync is stable and working as far as I can tell. I've had to redo the PR    (#63) to fix merge issues (I will stop doing this) and will add in some features that rely on the block parsing in the next few days when I clean up the code a bit. 

I'm happy with either having a release before or after I make these additions as I think we've added enough to warrant it! Also: the parsing stuff may get a bit complicated.

As far as #31 goes... I'm not that bothered by this at the moment, I just use the build in `workbench.action.terminal.runSelectedText` at the moment and I'm not sure of the advantage of this without any features built around it?

davidanthoff:
So the only advantage we have with the custom send to REPL thing is that you don't have to select the text, and the cursor automatically jumps to the next line. So you can run a file by just repeatedly hitting Ctrl+Enter.

davidanthoff:
@ZacLN I think once we have merged the remaining PRs, we should try to stabilize the release (i.e. only bug fixes) and then release.

Also, once that is done I would create a new repo for the language server. So if you have major additions right now planned, it might make sense to wait until we have the new repo, otherwise you would have to redo all of these. And I assume I'll have to do some file reorganization for the new repo.

davidanthoff:
How about this as next steps:
[...]

PR #84, opened 2016-11-18
WIP Use LanguageServer package

davidanthoff:
Don't merge this! This is work in progress that we can merge once we have released v0.5.0. This uses the LanguageServer package.

I've also changed things so that in debug mode the language server is just using the packages from the system package location, not a private installation. In release mode it is still using its own copy of things. That should make dev for us simpler.


ZacLN:
Fantastic stuff! So, for development I should push all PR's to LanguageServer and push bug fixes to this and the LanguageServer, right?

davidanthoff:
Yes, exactly. Hopefully we can keep the churn rate on master here low for 0.5.0, and then move onto 0.6.0.

ZacLN:
Also, I think a file restructuring might be nice such that all document handling like didOpen methods and parsing/linting are all in one file. If you agree I'll submit a PR tomororw

davidanthoff:
Yes, that makes sense to do this in the LanguageServer repo, right?

ZacLN:
Yes

davidanthoff:
I've just merged the split directly onto ``master``, mainly because there was some submodule stuff going on that was not super trivial. See #58 for the plan for the 0.5 release.

Issue #100, opened 2016-11-23
Run tests command


davidanthoff:
This would add a command "julia: Run tests" that would run the file ``test/runtests.jl`` if present.

PR #104, opened 2016-11-26
Add test runner command

davidanthoff:
Implements #100


davidanthoff:
It is a stop-gate until Microsoft/vscode#15179 is implemented, i.e. native VS Code UI for running tests.

My typical dev cycle is that I edit code, immediately run the unit tests to see whether anything broke, edit code etc. So this makes that a little easier.

ZacLN:
Ah ok

ZacLN:
On a side note, do you know if it's possible to setup the `.vscode` settings for the LanguageServer so that debugging automatically launches julia-vscode package?

davidanthoff:
Don't know, I now always have one window open with the julia-vscode folder and one with the LS, and then I hit F5 in the julia-vscode window, which starts everything as I want.

Issue #103, opened 2016-11-25
'print(\mu)' don't change to 'print(μ)' while pressing the tab key


sunqi1993:
![image](https://cloud.githubusercontent.com/assets/12682690/20625071/c541c5f6-b34b-11e6-8903-9316374549b1.png)

bthlgd:
I think this is a bug of vscode snippet.
You can try to add a space before \mu, then it works as expected.

davidanthoff:
I think one option would be to move these latex completions into the language server. I think @ZacLN had suggested that at some point? I'm not sure that would fix this, but it might be worth a try.

We could also get rid of [this](https://github.com/JuliaEditorSupport/julia-vscode/blob/master/snippets/generate_latex_snippets.jl) script then, and it would be less work to keep the julia base list of latex completions in sync with this extension.

ZacLN:
I'm going to do this this weekend. It just needs a function on the server file to locate words including the escape character then to provide a TextEdit rather than a simple word so that the initial backslash is removed

PR #106, opened 2016-11-26
fix latex completions

ZacLN:
This should get things working better for #103. Completions shortcuts for super/sub scripts are still slight wonky but can be accessed by searching for completions on `\^` or `\_` respectively.


davidanthoff:
Can you make this a PR against the LS repo? v0.5 is published, so any change to the LS code should now be added to the LS repo. I've approved #107, and that is all we need to do in this repo, correct?

ZacLN:
Ah ok, I didn't realise it had gone out, I'll close this.

davidanthoff:
Yep, v0.5 is in the wild :)

PR #110, opened 2016-11-27
executeJuliaCodeInREPL ctrl+Enter -> alt+Enter

innerlee:
``ctrl+Enter`` is a very useful key binding in vscode to go to a newline below current line. I suggest not override this command. It seems ``alt+Enter`` is available for our use.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
         ],
         "keybindings": [{
             "command": "language-julia.executeJuliaCodeInREPL",
-            "key": "ctrl+Enter",
+            "key": "alt+Enter",
             "when": "editorTextFocus"
         }],
         "configuration": {


ZacLN:
Personally, I prefer `ctrl+Enter` so maybe this should be set as a configurable setting?

davidanthoff:
One can probably already change this via [this](https://code.visualstudio.com/docs/customization/keybindings#_customizing-shortcuts)?

My original idea was that we would use ``Ctrl+Enter`` to execute the current block and then move the cursor to the next block, use ``Alt+Enter`` to execute the current block but not move the cursor, and then maybe one more to execute something like a cell a la RStudio markdown down the road.

innerlee:
@davidanthoff yes, you are right. Feel free to close this if the original arrangements of key bindings are better.

davidanthoff:
@innerlee Thanks nevertheless! Hope you continue to contribute, we need all hands :)

PR #115, opened 2016-12-03
Add command to run current file in a new terminal process

ZacLN:


davidanthoff:
I've been working on a debug adapter. That would allow us to hook into the normal VS Code start/run commands, including Ctrl+F5. I'm not yet done, and there are some tricky issues...

I would suggest we leave this PR open. If I'm done with the debug adapter before the next release, great. If not, we can merge this PR instead. Does that sound reasonable to you?

ZacLN:
That's fine, I have no attachment to or use case for this command myself

PR #121, opened 2016-12-21
Add weave command

davidanthoff:
This adds a weave command for [Weave.jl](https://github.com/mpastell/Weave.jl).

~~Right now I only support script weave files. It would be nice to add support for jmd files down the road, but I think before we do that we would need to figure out things like syntax highlighting and how that integrates with our LS etc.~~

FYI @mpastell

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -20,6 +20,16 @@ let g_context: vscode.ExtensionContext = null;
 let testOutputChannel: vscode.OutputChannel = null;
 let testChildProcess: ChildProcess = null;
 let testStatusBarItem: vscode.StatusBarItem = null;
+let weaveProvider: vscode.TextDocumentContentProvider = null;
+let lastWeaveContent: string = null;
+let weaveOutputChannel: vscode.OutputChannel = null;
+let weaveChildProcess: ChildProcess = null;
+
+export class WeaveDocumentContentProvider implements vscode.TextDocumentContentProvider {
+    public provideTextDocumentContent(uri: vscode.Uri): string {
+        return lastWeaveContent;
+    }
+}
 
 // this method is called when your extension is activated
 // your extension is activated the very first time the command is executed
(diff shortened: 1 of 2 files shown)


davidanthoff:
Actually, I changed my mind and just added support for jmd files as well.

src/extension.ts
```diff
@@ -207,6 +224,62 @@ async function openPackageDirectoryCommand() {
...
+        weaveChildProcess.on('close', async function (code) {
+            weaveChildProcess = null;
+
+            if (code==0) {
+                weaveOutputChannel.hide();
+
+                lastWeaveContent = await fs.readFile(output_filename, "utf-8")
+
+                let uri = vscode.Uri.parse('jlweave://nothing.html');
+                let success = await vscode.commands.executeCommand('vscode.previewHtml', uri, vscode.ViewColumn.One, "julia weave output");
```
ZacLN: Is it possible/desirable to have the preview split the editor?

mpastell:
Nice, looks like I need to try vscode.

davidanthoff:
> This seems to work well on first running the command but subsequent updates don't seem to pass through to the preview? (though the .html file is updated)

Hm, I can't reproduce that, my preview always updates... What are the exact steps you do to get that error?

src/extension.ts
```diff
@@ -207,6 +224,62 @@ async function openPackageDirectoryCommand() {
     }
 }
 
+async function weaveCommand() {
+    if (vscode.window.activeTextEditor === undefined) {
+        vscode.window.showErrorMessage('Please open a document before you execute the weave command.');
+    }
+    else if (vscode.window.activeTextEditor.document.isDirty || vscode.window.activeTextEditor.document.isUntitled) {
+        vscode.window.showErrorMessage('Please safe the current file before you weave it.');
```
davidanthoff: Fixed.

src/extension.ts
```diff
@@ -207,6 +224,62 @@ async function openPackageDirectoryCommand() {
...
+    }
+    else if (vscode.window.activeTextEditor.document.languageId!='julia' && vscode.window.activeTextEditor.document.languageId!='juliamarkdown') {
+        vscode.window.showErrorMessage('Only julia (.jl) files can be weaved.')
```
davidanthoff: Fixed.

src/extension.ts
davidanthoff: I thought about it, but I think it would set up an expectation about a live update of the preview, which we don't have... We could make it an option down the road.

ZacLN:
1. Load some file
2. run `julia: Weave document`
3. change file
4. close preview 
5. run `julia: Weave document`

though not closing the preview makes no difference

davidanthoff:
I can't reproduce this, but maybe the latest commit I just pushed will fix this. Could you try again?

ZacLN:
That does the trick

Issue #123, opened 2016-12-22
Should we use the atom julia grammar?


davidanthoff:
I just figured out how we can use the [Atom julia grammar](https://github.com/JuliaEditorSupport/atom-language-julia/blob/master/grammars/julia.cson) in this extension. All we have to do is convert the cson file into a json file, and then we can use that file in the extension here. Converting is simple using [this](https://atom.io/packages/cson).

The real question is, should we do that? My sense is that the grammar file over there probably gets a lot more attention than our version here. @be5invis what is your take on that? I know you added a few things recently to our grammar file, do you know whether those things are taken care of in the atom grammar already?

davidanthoff:
@be5invis We now are using the atom grammar, let me know if you have any concerns about that!

PR #124, opened 2016-12-22
WIP Use the grammar file from Atom

davidanthoff:
This is an implementation of #123. Not sure we should actually do this, but probably worth checking out.


ZacLN:
Sorry I merged the wrong PR, though I do think it's a good idea. How can I reverse out of this elegantly?

davidanthoff:
Lets leave it for now. If @be5invis thinks that we should use the original instead we can still undo it later.

PR #126, opened 2016-12-22
Various weave features

davidanthoff:
Lots of stuff here:

- At every weave, pre-start a julia process for the next weave. This speeds things up quit a bit.
- Have two weave preview commands, one for column one, one for side-by-side.
- Add a save to file command that shows all the supported output formats as an option and then saves the result to a file.
- Weave preview works without saving the source file and no longer generates any files in the workspace.
- Generally suppress some annoying console output messages about Weave overriding various methods.
- Use a template for the preview that doesn't show a link to the source file, and open the weave homepage in a different window.
- Drop support for script input files, so only .jmd files are supported going forward.

I think that pretty much covers what we need for now.

FYI @mpastell. Also, would be great if you could tag a new Weave release that has some of the additional fixes we added, so that this here works with the latest registered Weave version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,9 @@
         "onCommand:language-julia.startREPL",
         "onCommand:language-julia.executeJuliaCodeInREPL",
         "onCommand:language-julia.runTests",
-        "onCommand:language-julia.weave",
+        "onCommand:language-julia.weave-open-preview",
+        "onCommand:language-julia.weave-open-preview-side",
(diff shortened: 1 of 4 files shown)


mpastell:
Very exiting to see this work! It will certainly make Weave more usable.

I will try tag a new version on time but I think current master still needs some more testing. When are you going to release this?

davidanthoff:
No rush, will be a while before we are ready to release a new version.

PR #146, opened 2017-03-22
Simplify pkg handling

davidanthoff:
This enables us to have only one copy of the packages that we need in the VS Extension, not one copy per julia version. I've also updated the versions of the packages that we use. I think it might be easiest to just merge this, and if there is still a problem, fix this on ``master``, because this is a bit annoying to do on a branch.


davidanthoff:
Actually, I know that this breaks with ``master`` of LanguageServer right now, but it would still be easier to merge this here first and then I'll fix the issues over in LanguageServer.

Issue #37, opened 2016-11-06
Writing `` end`` should auto-unindent


davidanthoff:

ZacLN:
[ref](https://github.com/Microsoft/vscode/issues/2272)

davidanthoff:
There is some new info about this at Microsoft/vscode#2272. I don't fully understand it, but does this mean we can actually solve this right now?

tkelman:
It looks like they might at least make it so hitting enter after typing the end will de indent it?

davidanthoff:
Yeah, that is my impression. Microsoft/vscode#17868 has a lot of text, but it is still not entirely clear to me what the current status is, to be honest... But it does seem to be on the January 2017 milestone, so I assume something is going to improve with the next version.

davidanthoff:
Just an update here: things now work when one hits enter, i.e. then the line gets properly indented. I would still much prefer that things auto unindent when the word ``end`` has been typed, but we'll have to wait for upstream support for that.

I believe the upstream issue is Microsoft/vscode#2272.

davidanthoff:
Solved with VS Code 1.14. You need to set the ``editor.autoIndent`` user setting to ``true`` for this to work.

PR #148, opened 2017-03-27
Update indentation rules

davidanthoff:
This copies the indentation rule from [here](https://github.com/JuliaEditorSupport/atom-language-julia/blob/master/settings/language-julia.cson).

With current VS Code versions things should now unindent if you press Enter after the ``end``. Still not ideal, but I think we have to wait until Microsoft/vscode#2272 gets properly fixed for unindent when the word ``end`` is just typed.

This makes some progress on #37.


davidanthoff:
> I've got the new parser set up to correct these automatically now as well.

Ha, cool! I'll merge this for now.

PR #155, opened 2017-04-24
Parse options passed into the language server -- at least a little

oxinabox:
This makes flags passed in be handled as flags; with defaults and error messages

And use the same defaults for Package Dir not provided as LanguageServer.jl does.

It could probably be done nicer with a command line argument parsing library, but that would add a dependency.

### DIFF
--- a/scripts/languageserver/main.jl
+++ b/scripts/languageserver/main.jl
@@ -2,21 +2,35 @@ if VERSION < v"0.5"
     error("VS Code julia language server only works with julia 0.5 or newer.")
 end
 
-if length(Base.ARGS)!=2
-    error("Invalid number of arguments passed to julia language server.")
+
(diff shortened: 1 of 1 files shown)


davidanthoff:
Hm, why do we need to make this more general and robust? I always thought of this ``main.jl`` here as tightly coupled with the code [here](https://github.com/JuliaEditorSupport/julia-vscode/blob/master/src/extension.ts#L158), i.e. an internal implementation detail of the vscode extension. It really should never be called by anything else?

oxinabox:
Because it isn't really tightly coupled to the code here.
It is a good idea about having own package directory separate from .juliarc.
And I am leveraging it for my own purposes.

davidanthoff:
I think I would almost prefer if you would just make a copy of this in your repository for vim. I don't really want folks to take a dependency on this particular code, we might want to change it on a whim etc. Would that work? And then we would say that the stuff in LanguageServer.jl is actually properly supported for others as a dependency etc.

PR #158, opened 2017-05-01
Update protocol version, shift to julia v0.6

ZacLN:


scripts/languageserver/packages/AbstractTrees
```diff
@@ -1 +0,0 @@
-Subproject commit 74695d4634c2286e20c75accb2739d07344fe56a
```
davidanthoff: Something is wrong here with the submodules, this shouldn't reference local paths from your machine... Also, when I checkout this branch all the submodules don't update correctly, which I assume is due to whatever is going on here...

ZacLN: Hmm, yep that's all wrong. I'll have a fiddle

ZacLN: OK I've fixed it back to proper-submodules. I'll set the specific commits after the LS PR is merged/ the Julia release

ZacLN:
This is a mess, I can't get the submodules sorted so I'm going to close this open a new one.

PR #159, opened 2017-05-14
Update language server protocol to 3.0.0

ZacLN:
This bumps the julia requirement up to v0.6.0-  and the language server protocol to 3.0.0.

I'm having such problems removing submodules I'm going to do this in two steps, this adds CSTParser and the code change a second will remove the unneeded submodules.

### DIFF
--- a/scripts/languageserver/main.jl
+++ b/scripts/languageserver/main.jl
@@ -1,4 +1,4 @@
-if VERSION < v"0.6.0-"
+if VERSION < v"0.6.0-rc1"
     error("VS Code julia language server only works with julia 0.5 or newer.")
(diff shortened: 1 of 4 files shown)


scripts/languageserver/main.jl
```diff
@@ -1,21 +1,21 @@
-if VERSION < v"0.5"
+if VERSION < v"0.6.0-"
```
musm: I'm not sure if that is right, maybe safest to do  `0.6.0-rc1`   or `0.6-`

ZacLN: cheers

Issue #164, opened 2017-05-26
Support for displaying the output plots within the VsCode editor


abhi18av:
Hello julia-vscode team!

This is a feature request so that the entire code-feedback cycle within the VsCode editor speeds up. Could you provide a functionality like the [pythonVSCode](https://github.com/DonJayamanne/pythonVSCode) plugin ?

This would be awesome, since VsCode does have a better startup time etc. going for itself :)

davidanthoff:
Yes, this is on the radar, but not sure when we'll get to this.

davidanthoff:
I just published version 0.6.0-alpha.3 [here](https://github.com/JuliaEditorSupport/julia-vscode/releases/tag/v0.6.0-alpha.3) that has a plot pane. If anyone wants to test it and report back it would be great!

KristofferC:
Works ok:

![image](https://cloud.githubusercontent.com/assets/1282691/26745274/166023c8-47ea-11e7-891b-7ae70e511b01.png)

Running it again opens up a new plot window though:

![image](https://cloud.githubusercontent.com/assets/1282691/26745284/1d5a9c58-47ea-11e7-9b68-2de2fdc429fb.png)

davidanthoff:
@KristofferC Ah, thanks, that is a good bug report! The problem is that I always opened the plot pane in column one, so if you had moved it over to column two, then any new plot would open it again in column 1. In general both of these views then should display all the same plots and you can move through all the plots with the arrows icons etc. and both views will be synced.

[This](https://github.com/JuliaEditorSupport/julia-vscode/commit/b0bb2864ea0f1aec416cf57d15cf309f5f7acefc) commit fixes this, so that a normal plot command no longer opens a new plot view.

I think I'll wait with a new alpha release until we have some more new stuff, if that is ok. For now you can just close one of the views whenever a second one opens up, that should be harmless.

Another thing: right now I first look for a svg representation and use that if available, and the fallback is png. I've seen some discussion over in other plotting packages/IJulia etc. where people seem to prefer png as the better format. Do you have any opinion about that? It is easily changed...

KristofferC:
To me SVG should clearly be prioritized since it supports vector graphics and can support interactivity.

davidanthoff:
Ok, closing this because this is done :)

abhi18av:
Yup, and this works really well too. 👍

PR #168, opened 2017-06-01
Plot pane

davidanthoff:
Not perfect, and the UI needs some work, but I think this can be merged for now. Addresses #164.

It seems to work with Gadfly, PyPlot, VegaLite, Plots (with backends GR, PlotlyJS, some others don't work, but I think that might be more of a Plots.jl with julia 0.6 problem right now).

### DIFF
--- a/scripts/terminalserver/terminalserver.jl
+++ b/scripts/terminalserver/terminalserver.jl
@@ -1,5 +1,9 @@
 module _vscodeserver
 
+import Base: display, redisplay
+
+immutable InlineDisplay <: Display end
+
 pid = Base.ARGS[1]
 
 if is_windows()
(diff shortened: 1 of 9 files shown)


ZacLN:
I'm getting a 'can't' connect: no such file' error, I can't get much more detail as the terminal closes instantly. I seem to remember a difference in pipes between linux/windows?

ZacLN:
Scratch that, on linux the naming/positioning of the pipe seems to be identical to how it is on windows

scripts/terminalserver/terminalserver.jl
```diff
@@ -1,5 +1,9 @@
 module _vscodeserver
 
+import Base: display, redisplay
+
+immutable InlineDisplay <: Display end
+
 pid = Base.ARGS[1]
 
 if is_windows()
```
ZacLN: change to `if is_windows() || is_unix()` and drop the following if branch

davidanthoff:
@ZacLN Can you try again? I think it now should all work, plus UI is all finished as well.

ZacLN:
I'm just on a mobile 'til Monday I'm afraid so can't test things

davidanthoff:
Ah, ok. Ok if I just merge it anyways, without a review? I tested it on Ubuntu and it all seems to work. If there is a bug we can still fix it later?

ZacLN:
Yep for sure

KristofferC:
Maybe tag a new alpha release? I want to try with  https://github.com/KristofferC/PGFPlotsX

davidanthoff:
@KristofferC https://github.com/JuliaEditorSupport/julia-vscode/releases/tag/v0.6.0-alpha.3

PR #186, opened 2017-06-16
Updated readme for Windows path to executable.

tuckermcclure:
On Windows 10, using \\ for the path to the Julia executable on Windows did not work, whereas using forward slashes worked correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,17 @@ This [VS Code](https://code.visualstudio.com) extension provides support for the
 
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


davidanthoff:
Did you use double backlash, i.e. ``\\``, as the path separator? That should work.

tuckermcclure:
I did. It did not work. I also tried escaping the space and the ancient `C:\\PROGRA~!\\whatever` method. Only the forward slashes worked, and those did not require an escaped space.

pfitzseb:
FWIW, something like `"julia.executablePath": "C:\\Program Files\\julia\\Julia-0.6.0\\bin\\julia.exe"` works great for me.

davidanthoff:
Yes, the format that @pfitzseb showed is the right one and should definitely work on Win10.

tuckermcclure:
Ok, I will try again when I get home from the conference. Perhaps I had some other typo in there.

tuckermcclure:
After updating some things, `\\` works instead of `/`. Closing the pull request. Thanks.

davidanthoff:
Great.

PR #211, opened 2017-07-20
add reload-modules command

ZacLN:
Add menu option to force reload of (edited) modules imported in user code. 

This may allow slightly faster editting of LanguageServer as long as the LanguageServerInstance struct isn't changed? 

Goes w/ https://github.com/JuliaEditorSupport/LanguageServer.jl/pull/124

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
         "onCommand:language-julia.weave-save",
         "onCommand:language-julia.lint-package",
         "onCommand:language-julia.show-plotpane",
+        "onCommand:language-julia.reload-modules",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -116,6 +116,10 @@
                 "title": "Julia: Lint Package"
             },
             {
+                "command": "language-julia.reload-modules",
+                "title": "Julia: Reload imported modules"
+            },
+            {
```
davidanthoff: I think this command should be added to the ``activationEvents`` section in package.json.

ZacLN: yup, will do

PR #260, opened 2017-08-21
Add command to change REPL module

ZacLN:


davidanthoff:
Cool! I added two commits, I think they are pretty self-explanatory.

I guess one question is UI: It would be nice if one could see somewhere what the "current" module is, right? Could that be incorporated into the REPL prompt, maybe? Or we make it part of the status bar?

I assume you already planned to show a list of modules that one can pick from, instead of the free textform right now, right?

Overall this is really great, much better than what I had planned to do where we send the code via the socket. This strategy here, i.e. switching REPL module via a socket but still sending the code just via STDIN is way better, I think.

ZacLN:
I've changed it to a quick pick menu and added a minimal LSP style request type system to the interactions. It needs a way to handle:
- A call to change module when no REPL is open
- Calls to change module when multiple REPL's have been opened

ZacLN:
Ok, that handles the above two issues

ZacLN:
LanguageServer PR is merged so this can go in now

PR #283, opened 2017-10-01
split weave/repl functions into seperate files

ZacLN:
Creates classes for weave/repl functionality so that they can be moved into seperate files.
 I don't know much about typescript/OOP design so a quick review over the style is probably prudent.

I've added some infrastructure to have a REPL variables tree view but can't quite work out how to get the tree to update. I think it should work so that
1) hitting enter/a REPL eval sends an update of what variables are in the current module [here](https://github.com/ZacLN/julia-vscode/blob/refactor/scripts/terminalserver/terminalserver.jl#L27)
2) This message is picked up on the TS side [here](https://github.com/ZacLN/julia-vscode/blob/refactor/src/repl.ts#L145)
3) This should update the TreeDataProvider (this I haven't worked out)

After this is all working we should be able to set it up to click on variables and either access fields (i.e. type instances and modules) or get a display of the variable (i.e. a summary display of the object).

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,6 +96,10 @@
                 "title": "Julia: Execute Code"
             },
             {
+                "command": "language-julia.executeJuliaFileInREPL",
+                "title": "Julia: Execute File"
+            },
+            {
                 "command": "language-julia.toggleLinter",
                 "title": "Julia: Toggle Linter"
             },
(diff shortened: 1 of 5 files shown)


davidanthoff:
I've added this to the v0.9.0 milestone for now, right?

davidanthoff:
Ah, sorry, just saw the discussion on Slack, yes, lets include this in v0.8.0!

PR #289, opened 2017-10-06
Support multiple executablePath

innerlee:
And use the first path which works. 
Example:
```json
"julia.executablePath": [
    "/home/me/julia/bin/julia",
    "C:\\Users\\me\\AppData\\Local\\Julia-0.6.0\\bin\\julia.exe"],
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,23 +207,23 @@
             "commandPalette": [
                 {
                     "when": "false",
-                    "command": "language-julia.plotpane-delete"                    
+                    "command": "language-julia.plotpane-delete"
                 },
                 {
                     "when": "false",
-                    "command": "language-julia.plotpane-next"                    
+                    "command": "language-julia.plotpane-next"
                 },
                 {
                     "when": "false",
-                    "command": "language-julia.plotpane-first"                    
+                    "command": "language-julia.plotpane-first"
                 },
                 {
                     "when": "false",
-                    "command": "language-julia.plotpane-last"                    
+                    "command": "language-julia.plotpane-last"
                 },
                 {
                     "when": "false",
-                    "command": "language-julia.plotpane-previous"                    
+                    "command": "language-julia.plotpane-previous"
                 }
             ]
         },
@@ -282,14 +282,14 @@
             "command": "language-julia.plotpane-delete",
             "key": "delete",
             "when": "resourceScheme == jlplotpane"
-            }            
+            }
         ],
         "configuration": {
             "type": "object",
             "title": "julia configuration",
             "properties": {
                 "julia.executablePath": {
-                    "type": ["string", "null"],
+                    "type": ["array", "string", "null"],
                     "default": null,
                     "description": "Points to the julia executable."
                 },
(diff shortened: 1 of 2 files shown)


davidanthoff:
@innerlee Thanks for this, but I think I'd like to handle this in a different way. Also, we rearranged all the TS code, so this won't merge and I'm going to close this PR for now.

Here is what I plan to do about this: I don't want to support multiple locations for users, I think that makes things too complicated. What is your use case for that?

What I do want to support is this: if the setting is ``null`` (i.e. the default), I want the extension to look in a number of default locations where julia would end up being installed if someone just used the binary installers from julialang.org. This should work on Windows and on Mac, I don't really know about Linux. I think that would mean that for the vast majority of users, they wouldn't have to configure anything to get the extension running. More expert users can still edit the setting manually.

Do let me know if you have a specific use-case for supporting multiple paths in the setting, maybe I'm just missing something there!

innerlee:
I use both Windows and Ubuntu, and use the `Settings Sync` plugin to sync configurations across all machines. Basically, this is a feature request for handling binary location per OS (and a bad implementation/hack for that purpose). I usually set Julia binary path explicitly on Ubuntu. Unfortunately that will fail the Windows machine.

davidanthoff:
Ah, interesting... My gut feeling is that this too much of an edge case for us to support.

I can think of two potential solutions to this: I think if you set the setting to just ``julia`` and then make sure ``julia`` is on the ``PATH`` on both systems, things might work. If not, then we should fix that. Another option might be that we could also have the extension search some default locations on Linux if the setting is ``null``. I just don't know where that would be on Linux, but maybe there is a canonical way where we should look?

innerlee:
The first one is good. I do not think there are any default location on Ubuntu.

innerlee:
There are some subtlety I'm not quite sure. PATH is set on some shell config files like `.bashrc` or `.zshrc` etc. Would vscode aware of those shell config settings? If not, then julia would not be on the PATH unless we put them on somewhere like `/usr/bin`, which is not good by itself and not possible if you do not have sudo permission.

innerlee:
Just did a quick experiment using julia-vscode 0.7.0 on Ubuntu. Note that `julia` is on the PATH of zsh shell. And it is not on any system binary location. 

Set path to `julia` leads to `Couldn't start client julia Language Server` error, and do not set path leads to `Could not start the julia language server. Make sure the configuration setting julia.executablePath points to ...` error. 

So, I can think of two possible ways of fixing. One is support search path per OS. Or just like in this pr, support multiple search paths regardless of OS, and use the first working one. What do you think?

innerlee:
Here is the proposal:
have three settings `julia.executablePath.windows`, `julia.executablePath.osx`, and `julia.executablePath.linux`. If not set, find binary on system PATH. otherwise use the user setting.

This is inline with how [vscode does](https://code.visualstudio.com/docs/getstarted/settings#_copy-of-default-settings)
default

I can make a pr if this is okay :)

innerlee:
done in https://github.com/JuliaEditorSupport/julia-vscode/pull/344.

PR #291, opened 2017-10-07
add execute block command, return to active editor after any execution

ZacLN:
The new command executes the toplevel block that the start of any selection sits in. We may want to give options to determine code eval behaviour. e.g.:
- if there's a selection, eval it
- if no selection do we eval the line or block?

Also switches context back to the active editor after any eval


KristofferC:
If there is a section, eval it, sounds very reasonable to me.

PR #292, opened 2017-10-08
Support folding regions

### LINKED ISSUE #282: Support region markers
It will be supported in the next release.
See https://github.com/Microsoft/vscode-docs/blob/vnext/release-notes/v1_17.md#folding-regions

innerlee:
`#region` and `#endregion`.
fixes #282.

### DIFF
--- a/julia.configuration.json
+++ b/julia.configuration.json
@@ -26,5 +26,11 @@
         ["(", ")"],
         ["\"", "\""],
         ["`", "`"]
-    ]
+    ],
+    "folding": {
+        "markers": {
+            "start": "^\\s*#region",
+            "end": "^\\s*#endregion"
+        }
+    }
 }


KristofferC:
Can we in addition also add `#==*=#`? I think it is nice if you could visually block the regions:

```
#========================#
Hshsjshshsjsjsjsjhsyxyxtsnzb
Jsjshzhzhbsbsbsbsbsjjss
#========================#
```

innerlee:
Sure, I need to prepare two things:
* Find the gramma to write two styles of markers.
* Make sure start marker and end marker can be the same without causing pairing problem.

Or could you directly update this pr if you wish?

innerlee:
Ahh the first point can be implemented by writing clever regex, e.g., [typescript example](https://github.com/Microsoft/vscode/blob/45251543b93dd3f8d350f84ce34bf9f0727306e4/extensions/typescript/language-configuration.json#L30)

innerlee:
For the second point, I took the lazy path by asking. ref https://github.com/Microsoft/vscode/issues/35823

KristofferC:
Ah, I didn't realise the problem with common start and end markers. I don't want to hold up this PR so my suggestion can always be added later.

PR #295, opened 2017-10-11
add execute block command

ZacLN:


ZacLN:
There are separate commands (`language-julia.executeJuliaCodeInREPL` and `language-julia.executeJuliaBlockInREPL`) for each so can be bound as you like

PR #320, opened 2017-10-25
Fix typescript rearrangement

davidanthoff:
This should fix the issue that the rearrangement of the typescript files broke the plotting stuff. I actually think it also broke a number of other things :)

This PR ends up being quite large because I did a whole bunch of things:
- split more things out into their own files
- only use classes where we need to, for now it seems easier to just stick with globals in modules, we can still change that later
- I introduced a general pattern how new features can live in their own file

My gut sense is that we can merge this soonish, but then we need to make sure we actually test all the features we have with the new version to make sure we 1) fixed the things that were broken and 2) I didn't introduce new bugs.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -6,178 +6,108 @@ import * as fs from 'async-file';
 import * as path from 'path'
 import * as net from 'net';
 import * as os from 'os';
-var kill = require('async-child-process').kill;
-var exec = require('child-process-promise').exec;
-var tempfs = require('promised-temp').track();
 import { spawn, ChildProcess } from 'child_process';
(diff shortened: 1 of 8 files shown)


ZacLN:
Reorg looks good but breaks 'change module' and the repl variable tree view

davidanthoff:
I'm on it!

davidanthoff:
I think this should do the trick, both are working for me again. Thanks for being careful and finding this :)

ZacLN:
Yep thats done it

ZacLN:
Can we make the html/svg displayer communicate through the same pipe as the new REPL functions?

davidanthoff:
Yeah, I don't see why not. Maybe a v0.9.0 project?

ZacLN:
Yeah, no rush/end user impact

PR #339, opened 2017-11-01
Fix error in plot message encoding

### LINKED ISSUE #314: Plot pane gets stuck?
The plot pane seems to work for me for small, simple plots. After a few plots with a little more to them, the plot pane stops updating. I have to delete its plots and then make _two_ small plots before it's responsive again.

E.g., this usually works the first time:

```
    # Choose a plotting package.
    # pyplot() # Crashes after update on 2017-10-19.
    plotlyjs()

    # Define each plot that we'll need.
    p1 = plot(t, q_BI.',
              label  = ["q1" "q2" "q3" "q4"],
              ylabel = "Attitude Quaternion")
    p2 = plot(t, 180/π * ω_BI_B.',
              label  = ["ω1" "ω2" "ω3"],
              ylabel = "Rotation Rate (deg/s)")

    # Display those plots as subplots using the default "2" layout. Put any
    # common descriptors here.
    display(plot(p1, p2,
                 layout = 2,
                 xlabel = "Time (s)"))
```

However, when I change those variables and re-run it, it won't update the plot. If I do this:

```
plot(randn(10,10))
```

then nothing happens. If I do it _again_, then I see 10 random lines, each 10 samples long.

Now, if I re-run my plot code above, I see the plot.

I'm on Window 10, Julia 0.6, and VS Code 1.17.1, julia-vscode 0.7.0.

Anyone else seeing this issue?

davidanthoff:
Fixes #314.

Problem was that we need to send the length in terms of bytes, but we were sending the length in terms of characters. As soon as there is any Unicode character in an SVG file, things break :) This now fixes things.

### DIFF
--- a/scripts/terminalserver/terminalserver.jl
+++ b/scripts/terminalserver/terminalserver.jl
@@ -155,7 +155,7 @@ conn = connect(global_lock_socket_name)
 
 function display(d::InlineDisplay, ::MIME{Symbol("image/png")}, x)
     payload = stringmime(MIME("image/png"), x)
-    print(conn, "image/png", ":", length(payload), ";")
+    print(conn, "image/png", ":", endof(payload), ";")
     print(conn, payload)
 end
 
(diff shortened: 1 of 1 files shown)


ZacLN:
good catch (i hate these byte/character length bugs)

davidanthoff:
Yep, they are annoying. I think we had the same type of error previously with the LS communication...

PR #344, opened 2017-11-02
find julia exe

innerlee:
This pr find the default julia executable whenever the julia command is available on the vscode terminal.

Tested on Ubuntu 17.04 and Windows 10.

### DIFF
--- a/package.json
+++ b/package.json
@@ -341,6 +341,7 @@
         "child-process-promise": "^v2.2.1",
         "async-file": "^v2.0.2",
         "async-child-process": "^v1.1.1",
+        "promise-timeout": "^1.1.1",
         "promised-temp": "^v0.1.0"
     },
     "devDependencies": {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -26,12 +26,12 @@ let g_serverstatus: vscode.StatusBarItem = null;
 let g_serverBusyNotification = new rpc.NotificationType<string, void>('window/setStatusBusy');
 let g_serverReadyNotification = new rpc.NotificationType<string, void>('window/setStatusReady');
 
-export function activate(context: vscode.ExtensionContext) {
+export async function activate(context: vscode.ExtensionContext) {
     g_context = context;
 
     console.log('Activating extension language-julia');
 
-    g_settings = settings.loadSettings();
+    g_settings = await settings.loadSettings();
 
     // Status bar
     g_serverstatus = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left);
@@ -77,8 +77,8 @@ function setLanguageClient(languageClient: vslc.LanguageClient) {
     openpackagedirectory.onNewLanguageClient(g_languageClient);
 }
 
-function configChanged(params) {
-    let newSettings = settings.loadSettings();
+async function configChanged(params) {
+    let newSettings = await settings.loadSettings();
 
     repl.onDidChangeConfiguration(newSettings);
     weave.onDidChangeConfiguration(newSettings);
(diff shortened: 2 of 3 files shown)


davidanthoff:
Cool. I had also started working on this yesterday, see #345. I think both of these are v0.9.0, stuff, lets first get v0.8.0 out, and then I'll take another look at the PR here and we can decide how to merge those two efforts, ok?

innerlee:
Okay 💃

davidanthoff:
I've revived #345, which I think does everything that this PR does, right? So I'll close this one here, but please ping me if you think that this PR has something that the other one is missing.

innerlee:
no, they are different. Executing julia in js process is different from executing in a shell. Shells have confiure files. The configire file typically adds local binaries to PATH. So #345 will fail in this case.

davidanthoff:
I think the ``exec`` function I'm using will spawn a subshell and then execute the command in there, so wouldn't it pick up any configure file?

innerlee:
it may not pick the right configure. child-process-promise is a wrapper of child-process. The exec of child-process pick bash as default [see](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback) People seldom use bash. Mostly other shells.

davidanthoff:
Ah, so we should read the ``terminal.integrated.shell.linux`` setting (or the Mac one if on Mac) and then just pass that as the ``shell`` option to ``exec``? Then we should get the same shell that you would have gotten in this PR here, right?

innerlee:
that's plausible, unfortunately user can change system default shell and vscode respect this without need to configure. in this case, js exec will still use sh...

innerlee:
Currently #345 do not solve the problem mentioned in https://github.com/JuliaEditorSupport/julia-vscode/pull/289#issuecomment-341314709. That is, one set of settings that work on both windows and ubuntu. Note that both #289 and this pr can handle the situation.

Let me describe the origin of the problem. On ubuntu, people use other shells like zsh as default shell. Environment variables are set on the shell configure file like `.zshrc`. The configure file is shell-dependent. If one want to install julia, just extract the files to anywhere and add the path to `PATH` in the configure file.
Vscode terminal uses the default shell, in this case, the zsh. So in vscode terminal, you can type `julia` and launch the program. If you run js exec, it will use the `sh` as shell, so no `julia` can be found. the `terminal.integrated.shell.linux` isn't helpful because we do not need to configure this item, so js exec will still use `sh`, not zsh.

The benefit of the approach in this pr is that, you will get the julia the same as the one in the vscode terminal.
 
This pr is not a super set of #345, however it does complement #345. They can work together.

davidanthoff:
Yeah, I completely agree, I didn't understand that initially, but I now see the issue. First, thanks for clarifying this!

So I'm completely on board to try to find julia in the default shell, but I would like to ideally not open a terminal window (like this PR here does), but instead do it behind the scene like in #345.

Here is what I don't understand: what is the VS Code logic for picking up the system default shell? [This](https://code.visualstudio.com/docs/editor/integrated-terminal#_configuration) seems to suggest that it uses the content of ``$SHELL``, but the default VS Code config setting looks like ``"terminal.integrated.shell.linux": "sh"``, which to me looks like a hard-coded shell? After reading the doc, I would have expected that if ``terminal.integrated.shell.linux`` is ``null``, it would use the content of ``$SHELL``, and otherwise the content of ``terminal.integrated.shell.linux``, but that doesn't seem to be the case, right?

innerlee:
Thanks! I will try & test later to see if there's a reliable way to do this.

PR #350, opened 2017-11-04
Fix repl window focus

davidanthoff:
@ZacLN I think this should fix the problem that the current editor is opened in column one if one sends any code from the editor in column two to the REPL. It actually had nothing to do with plots, it happened for all code.

### DIFF
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -257,7 +257,6 @@ function executeSelection() {
         }
     }
     executeCode(text)
-    vscode.window.showTextDocument(editor.document);
 }
 
 function executeFile() {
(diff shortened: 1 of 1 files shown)


src/repl.ts
```diff
@@ -302,7 +299,7 @@ export interface TextDocumentPositionParams {
     position: vscode.Position
 }
 
-let getBlockText = new rpc.RequestType<TextDocumentPositionParams, void, void, void>('julia/getCurrentBlockText')
+let getBlockText = new rpc.RequestType<TextDocumentPositionParams, void, void, void>('julia/getCurrentBlockOffsetRange')
```
davidanthoff: Could we actually delete the whole line? ``getBlockText`` doesn't seem to be used anywhere. If so, lets delete it for v0.9.0, doesn't do any harm for now.

PR #353, opened 2017-11-04
Provide multi-root support for tasks

davidanthoff:
We should only merge this once we have released v0.8.0 final.


davidanthoff:
@ZacLN this could be merged now, right? Can you add a review?

PR #368, opened 2017-11-07
dont steal focus by showing output

### LINKED ISSUE #367: Output window opens while typing
The output window opens on its own while I'm typing with "julia Language Server" selected. It appears to happen each time there is an error in the language server. This behavior is quite distracting.

mac OS 10.13.1
VSCode 1.17.2
Julia VSCode 0.8.0-beta.7

A very similar issue has been reported here for another VSCode plug in:
https://github.com/angular/vscode-ng-language-service/issues/128

KristofferC:
Should fix #367 but I don't know how to test it.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -134,7 +134,8 @@ async function startLanguageServer() {
         synchronize: {
             configurationSection: ['julia.runlinter', 'julia.lintIgnoreList'],
             fileEvents: vscode.workspace.createFileSystemWatcher('**/*.jl')
-        }
+        },
+        revealOutputChannelOn: RevealOutputChannelOn.never
     }
 
     // Create the language client and start the client.


davidanthoff:
Hm, do we want this? I fear we just won’t get as many helpful error reports anymore because folks don’t see the LS output anymore when there is an error.

davidanthoff:
I'm moving this to v0.9.0 in any case, I want to get v0.8.0 out and this is not critical :)

innerlee:
Status bar is already good at indicating errors and warnings, my opinion 
screenshot_2017-11-08_09-57-18

davidanthoff:
This is about errors in the language server code itself, not errors in user code.

davidanthoff:
@ZacLN What do you think of this? I kind of feel that we could merge this, now that we have telemetry for LS server errors? Yes, users won't see these logs as easily if we merge this, but we should get the stack traces in any case automatically, so we probably no longer need to rely on users to report LS errors, right?

ZacLN:
Yep that makes sense to me (+ whatever change to vscode that made our logs disappear on error)

davidanthoff:
> (+ whatever change to vscode that made our logs disappear on error)

I complained and they removed some ``terminal.clear()`` call somewhere in VS Code. I'm not sure whether that was the thing that was at fault in our situation, but it was the only line that I had found that seemed to clear the console, so that might be a thing of the past. But still, lets merge this, we can always go back later.

Issue #362, opened 2017-11-06
Error in execute block code


davidanthoff:
Reported by @malmaud.

Start with this file:
````julia
module Hi
x=1
end

````
Position cursor at the end of ``end``, press Ctrl+Enter.

I get the following error log, but it then quickly disappears because some process goes away and VS Code deletes the log at that point:
````
INFO: 
ERROR: LoadError: BoundsError: attempt to access "module Hi\r\nx=1\r\nend\r\n\r\n"
  at index [0]
Stacktrace:
 [1] next at .\strings\string.jl:197 [inlined]
 [2] getindex(::String, ::Int64) at .\strings\basic.jl:32
 [3] ind2chr(::String, ::Int64) at .\strings\basic.jl:269
 [4] process(::LanguageServer.JSONRPC.Request{Val{Symbol("julia/getCurrentBlockOffsetRange")},LanguageServer.TextDocumentPositionParams}, ::LanguageServer.LanguageServerInstance) at c:\Users\david\source\julia-vscode\scripts\languageserver\packages\LanguageServer\src\provider_misc.jl:420
 [5] run(::LanguageServer.LanguageServerInstance) at c:\Users\david\source\julia-vscode\scripts\languageserver\packages\LanguageServer\src\languageserverinstance.jl:35
 [6] include_from_node1(::String) at .\loading.jl:569
 [7] include(::String) at .\sysimg.jl:14
 [8] process_options(::Base.JLOptions) at .\client.jl:305
 [9] _start() at .\client.jl:371
while loading c:\Users\david\source\julia-vscode\scripts\languageserver\main.jl, in expression starting on line 27
````

PR #370, opened 2017-11-07
fix offset

ZacLN:


davidanthoff:
Does this fix #362?

ZacLN:
In my tests it does...

PR #373, opened 2017-11-07
Use different keyboard shortcut for run block

davidanthoff:
Alright, I just miss Ctrl+Enter for "execute current line" too much to not give it a default binding :) Plus, Alt+Enter is free, so we can use that for execute block.

Plus, it means we don't have any breaking change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -231,11 +231,11 @@
             {
                 "command": "language-julia.executeJuliaCodeInREPL",
                 "key": "ctrl+Enter",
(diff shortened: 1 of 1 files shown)


KristofferC:
I'm a little bit sad by this (not much since I can just rebind the key). But Ctrl + Enter is block execution in Juno, IJulia (and MATLAB!) and is what a new user will typically default to. And I would argue that executing a block is in ~90% what you want to do.

ZacLN:
I'd be inclined to switch the default behaviour myself

PR #380, opened 2017-11-09
snippets: enhance function's snippet

iblis17:
peek 2017-11-09 20-42

### DIFF
--- a/snippets/snippets.json
+++ b/snippets/snippets.json
@@ -52,9 +52,9 @@
 	"Macro Declaration": {
 		"prefix": "macro",
 		"body": [
-			"macro $1($2)",
-			"\t$3",
-			"end"
+			"macro ${1:name}($2)",
+			"\t$0",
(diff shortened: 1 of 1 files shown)


davidanthoff:
I'm not a fan of this, to be honest. I feel it imposes a specific coding style on a large user base that is not really an official julia thing. I think folks that would like this kind of snippet can just add this (or some other favorite style) to their own installation with these instructions: https://code.visualstudio.com/docs/editor/userdefinedsnippets.

iblis17:
I agree that everyone has different coding style.
Can we create a snippet gallery or something like that to collect various contributed snippet?

davidanthoff:
Yes, that is a great idea! Maybe just a wiki page for now?

iblis17:
I added them here: https://github.com/JuliaEditorSupport/julia-vscode/wiki/Snippets#gallery

davidanthoff:
Great, thanks!

PR #390, opened 2017-11-13
handle workspaceFolders

ZacLN:


src/extension.ts
```diff
@@ -159,4 +159,7 @@ async function startLanguageServer() {
             g_serverstatus.hide();
         })
     })
+    vscode.workspace.onDidChangeWorkspaceFolders((e) => {
+        g_languageClient.sendRequest('workspace/didChangeWorkspaceFolders', e.added)
+    })
```
davidanthoff: Don't we have to handle removing workspace folders as well?

davidanthoff: Oh, and I kind of had assumed the languageClient would handle this stuff automatically? Seems weird that we need to listen to the ``onDidChangeWorkspaceFolder`` event here and then rerout it to the LS...

package.json
```diff
@@ -1,353 +1,358 @@
...
+		"mocha": "^4.0.1",
+		"@types/node": "^8.0.47",
+		"@types/mocha": "^2.2.44"
+	},
+	"__metadata": {
+		"id": "038dd41f-4a41-48e3-87de-ab2ea4310780",
+		"publisherDisplayName": "julialang",
+		"publisherId": "205aeff7-3cad-4caf-9f66-11fb2a0cbc74"
+	}
+}
```
davidanthoff: This commit shouldn't have any edits to this file either :)

package-lock.json
```diff
@@ -2738,27 +2887,17 @@
...
             "integrity": "sha1-pcbVMr5lbiPbgg77lDofBJmNY68=",
             "dev": true
         },
-        "yallist": {
-            "version": "https://registry.npmjs.org/yallist/-/yallist-2.1.2.tgz",
-            "integrity": "sha1-HBH5IY8HYImkfdUS+TxmmaaoHVI="
-        },
         "yauzl": {
             "version": "2.9.1",
             "resolved": "https://registry.npmjs.org/yauzl/-/yauzl-2.9.1.tgz",
```
davidanthoff: This commit shouldn't have any edits to this file :)

src/repl.ts
```diff
@@ -266,7 +266,7 @@ function executeFile() {
     if (!editor) {
         return;
     }
-
+
```
davidanthoff: Not important, but there also doesn't need to be an edit here, right?

innerlee: I suggest do a pr removing all the training spaces in source files.
Reasons are:
1. Some plugins highlight trailing spaces. Keeping them is not good looking.
2. Some configurations can remove trailing spaces when saving. It will be a pain to retain thoes spaces.

src/extension.ts
ZacLN: It's supposed to automatically send a notification on the above event when both server and client have `workspaceFolders`cababilities but doesn't in practice

davidanthoff:
For now I've just committed the line ``g_languageClient.registerProposedFeatures();`` on ``master`` (without a PR), we certainly need that. I'm still hoping that we can figure out how to get the notifications automatically without rolling our own version.

PR #397, opened 2017-11-18
WIP Add juliavscode/html mime type

davidanthoff:


davidanthoff:
This allows a type to spit out some custom HTML to be shown in the plot pane in VS Code. Right now we don't capture output as HTML for the plot pane because that would capture more than just plots (for example it would capture DataFrame output into the plot pane, which seems not ideal). With this one here a plot package can create some specific display logic for VS Code. The neat thing is that the package doesn't need to take a dependency on anything VS Code specific, it just adds support for a custom mime type, and done. I'm experimenting with that over in VegaLite.jl right now

davidanthoff:
This was still WIP, i.e. I was not really sure whether we should merge it :) Having thought about it now, probably ok to just have it in there, so no worries.

But maybe we'll follow the convention that for PRs from the two of us, we'll always let the person who opened the PR merge that PR? After the other one has approved? I know I haven't followed that always, but that might make sense? And for PRs from others, we only merge once both of us have approved the PR?

ZacLN:
Yep sorry about that, it is the convention we've been following I got mixed up in a flurry of merges

aaronsheldon:
oh man I'm really going back and forth on this one. Having data frames, help and methods() outputs, etc... render in a separate pane can be very handy. But I see the point - in that there is a tension between the default behavior of VSCode, and the default behavior of individual packages. In particular one generally expects what happens in the REPL to stay in the REPL, unless you use a function that one explicitly expects to step out of that IO stream.

Perhaps a configuration option can be added to enable/disable REPL html output to a separate pane, with the default configuration to disable? Maybe call the option "EnableRichOutput"?

PR #406, opened 2017-11-22
Update README.md

scls19fr:


innerlee:
We can find this command at the `File` menu, but the command itself does not belong to `File`, I think

scls19fr:
being a new user of Visual Studio Code I spent some time to find this command...

davidanthoff:
Thanks, this is a good suggestion. I just merged and pushed a slightly different formulation that does the same thing (seemed easier than go back and forth on this PR). Thanks for the inspiration!

PR #414, opened 2017-11-26
Add dependants test Task

ZacLN:


src/tasks.ts
```diff
@@ -44,7 +44,16 @@ async function provideJuliaTasksForFolder(folder: vscode.WorkspaceFolder): Promi
             let testTask = new vscode.Task({ type: 'julia', command: 'test' }, folder, `Run tests`, 'julia', new vscode.ProcessExecution(g_settings.juliaExePath, ['--color=yes', '-e', 'Pkg.test(Base.ARGS[1])', rootPath]), "");
             testTask.group = vscode.TaskGroup.Test;
             testTask.presentationOptions = { echo: false };
-            result.push(testTask);
+            result.push(testTask);   
+        }
+
+        if (await fs.exists(path.join(rootPath, 'REQUIRE'))) {
```
davidanthoff: That is probably a bit too lose, someone might have opened just the ``test`` folder of a package, and that can also contain a ``REQUIRE`` file. I think a more robust version would be to simply test whether the folder is a child folder of ``.julia/v0.x``, right? [This](https://github.com/JuliaEditorSupport/julia-vscode/blob/master/src/openpackagedirectory.ts#L19) line would give you the package root folder, and then we can just check whether the current folder is a direct (!) child of that, right?

src/tasks.ts
```diff
@@ -44,7 +44,16 @@ async function provideJuliaTasksForFolder(folder: vscode.WorkspaceFolder): Promi
...
+            let scriptpath = path.join(g_context.extensionPath, 'scripts', 'test_deps.jl')
```
davidanthoff: Can we move the ``test_deps.jl`` file into a folder ``tasks``?

src/tasks.ts
```diff
@@ -44,7 +44,16 @@ async function provideJuliaTasksForFolder(folder: vscode.WorkspaceFolder): Promi
...
+            let depTestTask = new vscode.Task({ type: 'julia', command: 'test' }, folder, `Run tests on dependents`, 'julia', new vscode.ProcessExecution(g_settings.juliaExePath, ['--color=yes', '-e', 'include("' + scriptpath + '")', path.basename(rootPath)]), "");
```
davidanthoff: I believe we need to use a different string for ``command`` because ``test`` is used above already. Maybe just ``testdependencies``?

Can we make the description ``Run tests of dependent packages``? Or something like that?

I think we can skip the whole ``-e 'include(...)'`` stuff and just pass ``scriptpath`` as the second element in that array?

scripts/test_deps.jl
```diff
@@ -0,0 +1,20 @@
+function get_deps(pkg, deps = Set{String}())
+    for d in Pkg.dependents(pkg)
+        push!(deps, d)
+        get_deps(d, deps)
```
davidanthoff: Hm, maybe we should just test direct dependencies? I'm not sure, but this might take a very long time...

PR #429, opened 2017-12-24
Add telemetry support

davidanthoff:
This will give us telemetry reports on how users actually use the extension. We can do all sorts of crazy things with this, for example detect performance problems in the wild etc.

I think there is also a way to automatically get crash reports with stack traces whenever anything goes wrong on any client system. If so, that should help us a lot with improving the quality of the extension.

We'll have to think about user opt in. I think right now it just uses the global VS Code user setting on that, but we might need our own.

Also, this is far from ready. And I need to figure out how I can give @ZacLN access to my Azure Application Insights account.

TODO
- [x] Crash reports from extension
- [x] Crash reports from LS
- [x] Add user opt-out option
- [x] Give @ZacLN access to Azure App Insights
- [x] Remove user part of paths in stack trace
- [x] Write privacy policy
- [x] Instrument more code (optional)
- [ ] Add support for LS instrumentation (optional)


ZacLN:
Stack traces filling up my inbox would be a true Christmas present! 😃

davidanthoff:
@ZacLN can you take a look at https://github.com/JuliaEditorSupport/julia-vscode/wiki/Telemetry. Does that sound reasonable to you? I'll also ask on the slack channel for opinions.

davidanthoff:
@ZacLN Now ready for review! I changed the whole opt-in thing again. Essentially we now never send anything unless a user has explicitly agreed to that. BUT, I've made it easier to agree than to opt-out :)

PR #435, opened 2017-12-29
set default repl editor to vscode

ZacLN:


src/repl.ts
```diff
@@ -132,7 +132,7 @@ function startREPL(preserveFocus: boolean) {
         startREPLConn()
         startPlotDisplayServer()
         let args = path.join(g_context.extensionPath, 'scripts', 'terminalserver', 'terminalserver.jl')
-        g_terminal = vscode.window.createTerminal("julia", g_settings.juliaExePath, ['-q', '-i', args, process.pid.toString()]);
+        g_terminal = vscode.window.createTerminal("julia", g_settings.juliaExePath, ['-q', '-i', args, process.pid.toString(), process.execPath]);
```
davidanthoff: I think it would be easier to use the other ``createTerminal`` method here: ``createTerminal(options: TerminalOptions): Terminal``.  ``TerminalOptions`` (https://code.visualstudio.com/docs/extensionAPI/vscode-api#TerminalOptions) allows us to directly set env vars in the child process, so we don't have to pass things around via command line args and don't have to make any change to ``terminalserver.jl`` at all.

ZacLN: Oops, I had a search and couldn't find this- will change it.

ZacLN: I remember the problem now- I can't seem to import `TerminalOptions` from `vscode`

davidanthoff: Maybe you are on a really old version of the ``vscode`` package? Have you tried a ``npm install`` in the root folder for the extension?

ZacLN: Hmm can't seem to get it up, even from a clean install

davidanthoff: I don't think you need to import that, you just write:
````
g_terminal = vscode.window.createTerminal({
    name: "julia",
    shellPath: g_settings.juliaExePath,
    shellArgs: ['-q', '-i', args, process.pid.toString()],
    env: WHATEVERWENEEDHERE});
````

ZacLN: Nope thats not it: 'env does not exist in type TerminalOptions'

Issue #176, opened 2017-06-09
Add a grid view for tables


davidanthoff:
It can definitely be done, see https://github.com/microsoft/vscode-mssql and https://github.com/microsoft/angular2-slickgrid. Probably a big project, this issue is mainly to remind me of these links for later.

davidanthoff:
Has been shipping for a while.

PR #438, opened 2018-01-02
WIP Add table view option

davidanthoff:
Very much work in progress.

I'm using https://www.datatables.net/ for the actual UI, mostly because it seemed really simple to integrate. Probably worth checking whether that is a good choice or not.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
         "onCommand:language-julia.lint-package",
         "onCommand:language-julia.show-plotpane",
         "onCommand:language-julia.reload-modules",
+        "onCommand:language-julia.showTableView",
         "onLanguage:julia",
         "onLanguage:juliamarkdown",
         "workspaceContains:deps/build.jl",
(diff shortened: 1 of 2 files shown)


xiaodaigh:
I had some experience using the https://github.com/bvaughn/react-virtualized to show only part of the table; so it would be an awesome addition as performance can serious degradation if not careful with the implementation especially when dealing with large datasets.

datatables.net is fine if you implementation the lazy load feature. I find react-virtualized easier to use but then you have to use react but I guess it's quite popular now.

Ken-B:
see #176, a reminder from the past ;-)

davidanthoff:
Haha, yes, this is not new. But now I have a bit more of an idea how to do it and actually think it is not really that big of a project. At least to get something basic going. Big datasets will be a different story...

davidanthoff:
I'm closing this attempt because the whole architecture for custom web views in VS Code has changed in the meantime. I still plan to do this, but there is nothing in this PR that would be useful at this point.

PR #446, opened 2018-01-14
use full exec path

innerlee:
using `julia` might trigger error in some case. Use full path works fine.

screenshot_2018-01-14_14-37-12

### DIFF
--- a/src/juliaexepath.ts
+++ b/src/juliaexepath.ts
@@ -35,7 +35,7 @@ export async function getJuliaExePath() {
             let foundJulia = false;
             for (let p of pathsToSearch) {
                 try {
-                    var res = await exec(`"${p}" -e "println(VERSION < v\\"0.7-\\" ? JULIA_HOME : Sys.BINDIR)"`);
+                    var res = await exec(`"${p}" --startup-file=no --history-file=no -e "println(VERSION < v\\"0.7-\\" ? JULIA_HOME : Sys.BINDIR)"`);
                     if (p == 'julia' || p == "julia.exe") {
                         // use full path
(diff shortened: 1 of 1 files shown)


src/juliaexepath.ts
```diff
@@ -5,6 +5,7 @@ import * as os from 'os';
 import * as path from 'path';
 import * as process from 'process';
 var exec = require('child-process-promise').exec;
+var which = require('which')
```
davidanthoff: Add a ``;`` at the end.

src/juliaexepath.ts
```diff
@@ -36,10 +37,12 @@ export async function getJuliaExePath() {
             for (let path of pathsToSearch) {
                 try {
                     var res = await exec(`"${path}" -v`);
-
                     actualJuliaExePath = path;
+                    if (actualJuliaExePath == 'julia') {
+                        actualJuliaExePath = which.sync('julia')
```
davidanthoff: Could you change this to use ``await``? You'll need to promisfy ``which`` to do that. I think [this](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original) describes quite well how to promisfy a function like ``which`` so that it can be used with ``await``.

innerlee:
Tested on Ubuntu and Win 10. I do not have Mac.

davidanthoff:
Lets just wing it with Mac. I've never tested this on Mac in the first place, plus we'll have lots of beta testers with Mac, so if there is a problem, we can figure it out then. Thank!

@ZacLN ok with you? Just waiting for your review, and then we can merge.

davidanthoff:
Thanks, @innerlee!

PR #458, opened 2018-01-26
Use KaTeX instead of MathJax

ChristianKurz:
[KaTeX](https://github.com/Khan/KaTeX) is a lot faster than MathJax. As of recently, [functionality is on par](https://khan.github.io/KaTeX/function-support.html).

Same as my [PR for Weave.jl](https://github.com/mpastell/Weave.jl/pull/119), so perhaps it's wise to wait for @mpastell's comments.

### DIFF
--- a/scripts/weave/preview.tpl
+++ b/scripts/weave/preview.tpl
@@ -6,13 +6,24 @@
   {{#:title}}<title>{{:title}}</title>{{/:title}}
   {{{ :header_script }}}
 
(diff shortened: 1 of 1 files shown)


iblis17:
I think KaTex is still missing some font support like `mathsrc` for me ... etc :/

davidanthoff:
So right now it appears that this is not being merged in Weave.jl, so I'll hold off here until there is a decision upstream.

PR #485, opened 2018-04-10
[Discussion] [Live Share] Restrict language services and tasks to local workspaces

lostintangent:
In preparation for [Visual Studio Live Share](https://aka.ms/vsls) adding support for "guests" to receive remote language services and tasks for Julia, this PR simply updates the current `DocumentSelector` to be limited to `file` and `untitled` (unsaved) files. This way, when someone has the Julia extension installed, and joins a Live Share session (where files use the `vsls:` scheme), their language services and tasks will be entirely derived from the remote/host side, which provides a more accurate and project-wide experience (guests in Live Share don't have local file access to the project they're collaborating with).

*Note: As an example, the TypeScript/JavaScript language services that come in-box with VS Code [already have this scheme restriction](https://github.com/Microsoft/vscode/blob/master/extensions/typescript-language-features/src/utils/fileSchemes.ts#L12), and so this PR replicates that behavior.*

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -62,10 +62,13 @@ export async function activate(context: vscode.ExtensionContext) {
     juliaexepath.activate(context, g_settings);
     repl.activate(context, g_settings);
     weave.activate(context, g_settings);
-    tasks.activate(context, g_settings);
     smallcommands.activate(context, g_settings);
     packagepath.activate(context, g_settings);
     openpackagedirectory.activate(context, g_settings);
+
+    if (vscode.workspace.rootPath) {
+        tasks.activate(context, g_settings);
+    }
 
     // Start language server
     startLanguageServer();
@@ -152,7 +155,10 @@ async function startLanguageServer() {
     };
 
     let clientOptions: LanguageClientOptions = {
-        documentSelector: ['julia', 'juliamarkdown'],
+        documentSelector: [
+            { language: 'julia', scheme: 'file' },
+            { language: 'juliamarkdown', scheme: 'file' },
+        ],
         synchronize: {
             configurationSection: ['julia.runlinter', 'julia.lintIgnoreList'],
             fileEvents: vscode.workspace.createFileSystemWatcher('**/*.jl')


lostintangent:
@davidanthoff Is there anything else I could do to help with this PR?

davidanthoff:
Thanks! I'm just coming out of a very busy non-julia phase, but should be able to look at this soon.

I think my plan would be to release v0.10 first, and then tackle this.

src/extension.ts
```diff
@@ -62,11 +62,14 @@ export async function activate(context: vscode.ExtensionContext) {
...
     weave.activate(context, g_settings);
-    tasks.activate(context, g_settings);
     smallcommands.activate(context, g_settings);
     packagepath.activate(context, g_settings);
     openpackagedirectory.activate(context, g_settings);
 
+    if (vscode.workspace.rootPath) {
+        tasks.activate(context, g_settings);
+    }
+
```
davidanthoff: This seems unrelated? If this fixes a bug, can you please move it into a separate PR?

src/extension.ts
```diff
@@ -152,7 +155,10 @@ async function startLanguageServer() {
     };
 
     let clientOptions: LanguageClientOptions = {
-        documentSelector: ['julia', 'juliamarkdown'],
+        documentSelector: [
+            { language: 'julia', scheme: 'file' },
+            { language: 'juliamarkdown', scheme: 'file' },
+        ],
```
davidanthoff: This should also include the ``untitled`` scheme, right?

davidanthoff:
@lostintangent, bump, would probably be good if we could merge this soon. Are the two comments I raised valid? If so, any chance you might be able to address them?

davidanthoff:
@lostintangent Do you think you might have time soon to take a look at this? Just trying to plan a bit.

davidanthoff:
Moving this to the backlog. Not clear to me what we need to do hear, and it shouldn't hold up our next release.

PR #495, opened 2018-05-14
WIP Start work on new terminal communication

davidanthoff:
This is the start of a rewrite of the whole communication stuff with the REPL. The goal is to make things like ``@__FILE__`` and all relative paths from ``include`` work automatically. It also should make the whole communication more robust in general.

Way to WIP to look at, though :)

### DIFF
--- a/scripts/terminalserver/terminalserver.jl
+++ b/scripts/terminalserver/terminalserver.jl
@@ -34,76 +34,11 @@ else
 end
 
 import Base: display, redisplay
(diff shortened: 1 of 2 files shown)


davidanthoff:
Note to self: there is a lot of stuff we can copy from https://github.com/JunoLab/Atom.jl/blob/master/src/repl.jl and https://github.com/JunoLab/Atom.jl/blob/master/src/eval.jl.

PR #522, opened 2018-07-06
Add config for LS tracing

davidanthoff:
VS Code now has a native tracing for LS communication, this enables that. One can then take the native output, upload it [here](https://microsoft.github.io/language-server-protocol/inspector/) and get a nice UI to analyze the traffic.

We should probably also find out how we can save the log to a file automatically...


davidanthoff:
Note to self: wait until `traceOutputChannel` is available with this.

davidanthoff:
Yes, this just works and outputs all the trace information in a new output channel.

PR #553, opened 2018-10-11
Add support for environments

davidanthoff:
This actually does a lot :) I goes hand in hand with https://github.com/JuliaEditorSupport/LanguageServer.jl/pull/290 and https://github.com/ZacLN/StaticLint.jl/pull/7.

First how the symbol server now works: the cache is no longer built from the extension, instead that is all handled from the LS side. LS loads SymbolServer.jl, and that handles/manages the cache. If SymbolServer realizes that it needs info for a user package, it will start a new julia process that collects that info, and then stores that stuff in a cache. So this is essentially the same design @ZacLN had before, with the only difference that the symbol cache now is created automatically with no user interaction, fully from the julia side of things. This now also (finally!) fixes the situation with LS: the LS julia process is now completely isolated from anything in the ``.julia`` folder, and then only the sub process that SymbolServer.jl starts will actually interact with the standard depot again. I think that will (hopefully) greatly increase the stability of everything.

This also has the start of a UI for switching environments.

Some things still need to be done:
- [x] The whole thing doesn't really work right now... Well, it does create the cache etc., but then no doc strings appear, so something is still off... I don't think this is going to be a big thing, probably just some stupid mistake somewhere...
- [ ] We need to make the cache version aware, i.e. cache per git sha1 of packages, like the precompiled versions.
- [ ] The LS needs to handle the change of environment custom message that I introduced.


davidanthoff:
Alright, I think this should now be functionally equivalent to what is happening on ``master``, but with the improvements mentioned above.

I think it would be easiest if we merge this (and the two other PRs) as is for now and then go on fixing things from there.

I looked into the whole new package manager/environment story a bit more, and it is complicated... I think we'll probably need some help from Stefan or someone like that to get it right... I'll open an issue for that separately, but maybe @ZacLN you could review this and we try to get it merged rather sooner than later, so that we don't diverge too much?

PR #588, opened 2018-11-15
use bracketed paste to send text to the terminal, fixes indentation problems

KristofferC:
Fixes indentation problems.

I have no idea how to test this. 

Ref: https://discourse.julialang.org/t/vs-code-extension-v0-11-0-beta-release/17541/9

### DIFF
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -293,7 +293,7 @@ async function executeCode(text) {
     var lines = text.split(/\r?\n/);
     lines = lines.filter(line => line != '');
     text = lines.join('\n');
-    g_terminal.sendText(text + '\n', false);
+    g_terminal.sendText('\e[200~' + text + '\n' + '\e[201~', false);
 }
 
 function executeSelection() {


KristofferC:
Well, did you try it out somehow and see that it works ok? :P

davidanthoff:
Na, I trust you. Plus we are still in the beta phase, so hopefully problems will be reported back once we ship a new beta.

davidanthoff:
Haha, ok, maybe I will check it out locally then :) I had interpreted your comment as "I don't know how to write an automated test".

KristofferC:
No, it is: "I don't know how to get this to stick in my local extension." :)

ZacLN:
Nope I tried this approach before, it won't work (on linux at least)

davidanthoff:
Ok, I'll revert.

KristofferC:
Why would #495 solve this. Also, can someone tell me how to test this. I don't see a reason why it shouldn't work.

davidanthoff:
The goal with #495 is to in general no longer send code to the REPL as a string a la copy-paste, but instead have a background green thread in the REPL process that listens on a named pipe for messages from the extension (we already have that background server), and then code to be executed would be send via this pipe to the REPL process.

davidanthoff:
Easiest way to get a working dev environment for this is to do step 1 & 2 from [Option 3](https://github.com/JuliaEditorSupport/julia-vscode/wiki/Updating-to-the-lastest-julia-vscode#option-3). At that point you can open the julia-vscode folder in VS Code, and run the extension with F5 in a debug environment.

KristofferC:
Okay, FWIW just to make it explicit, this is how things are right now:

nov-15-2018 13-46-02

This keeps going for minutes.

PR #622, opened 2018-12-16
add option to disable wrapping of REPL code (also bump submodules)

ZacLN:


davidanthoff:
Ah, now I see that the submodule update is on purpose. I already bumped submodules on ``master``... In any case, I have another PR that I think addresses this also, but without a new option, which seems simpler to me.

PR #625, opened 2018-12-23
Fix sendText call on Windows

### LINKED ISSUE #610: Strange error in VS code input
Hello.

I've been using Julia on VS code for several days but today it doesn't work. I don't know why.
Maybe it's because now I'm using it with language-julia-0.11.0-beta.2.vsix 

Whatever I try to run on the main editor it adds it a prefix "200~" and generates an error.

For example:
`using DataFrames, GLM`

produces:
```
200~using DataFrames, GLM
ERROR: MethodError: no method matching ~(::Int64, ::Int64)
Closest candidates are:
  ~(::Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8}) at int.jl:274
  ~(::Integer) at int.jl:138
Stacktrace:
 [1] top-level scope at none:0
```

Has anyone had the same problem?
Do you think is a problem wit VS Code, the extension or with Julia?
running the code directly on the REPL works well.
I've tried reinstalling the extension but the problem is still there.
I've also tried reinstalling a different version of VS Code.

PD:  I've reinstalled the beta1 and it works but beta2 doesn't.

davidanthoff:
Fixes #610.

This would replace #622. I think I would prefer not to add an option for this, instead I think we should use this on platforms where it works and not where it doesn't work, but really we should figure that out.

### DIFF
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -294,7 +294,12 @@ async function executeCode(text) {
     var lines = text.split(/\r?\n/);
     lines = lines.filter(line => line != '');
     text = lines.join('\n');
-    g_terminal.sendText('\u001B[200~' + text + '\n' + '\u001B[201~', false);
+    if (process.platform == "win32") {
+        g_terminal.sendText(text + '\n', false);
+    }
+    else {
+        g_terminal.sendText('\u001B[200~' + text + '\n' + '\u001B[201~', false);
+    }
 }
 
 function executeSelection() {


davidanthoff:
We can revisit if we see a bug report on Unix?

ZacLN:
Yep

KristofferC:
This bug is weird because it isn't the terminal that handles the escape sequence, it is the Julia REPL itself. The terminal only decides whether it should put in the escape sequence on a copy paste.

PR #641, opened 2018-12-27
Preserve webview panel column

EricForgy:
I was having an issue with the plot panel. I wanted to write code in one editor column and have the plot panel in another editor column. When I manually moved the plot panel to a second column and redraw the chart, the column disappeared and the plot panel covered my code. 

This PR attempts to fix this so we can leave a plot panel open in another column. Plotting again plots to the same column wherever you put it.

image

Note: I am getting some errors messages that the extension has crashed (it might just be on my side), so let's have a close look and test it out before merging, but wanted to submit to see what you think.

### DIFF
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -49,7 +49,7 @@ function showPlotPane() {
         }, null, g_context.subscriptions);
 
         g_plotPanel.onDidChangeViewState(({ webviewPanel }) => {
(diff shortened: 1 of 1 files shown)


EricForgy:
This last commit ☝️ seems to fix the focus issue. I still get some extension errors, but I don't think those are related to this PR.

EricForgy:
Yay! Thanks 😊

PR #642, opened 2018-12-27
Intellisense suggested changing Languages -> Progamming Languages

EricForgy:
This is a minor change to package.json suggested by Intellisense. Seems reasonable 😊 

I can add this to my other PR (#641) if you prefer.


EricForgy:
Alternatively, I could rename this PR to "Cosmetics" or something and go through and clean up a lot of Intellisense issues. What do you think?

PR #644, opened 2018-12-30
Release keyboard bindings after plot pane is closed

davidanthoff:
@EricForgy I think that should fix the bug you mentioned on Slack?


davidanthoff:
I'm just going to merge this and create a new RC, I'm 99% sure this fixes a bad bug, so better to get it out sooner :)

PR #663, opened 2019-01-15
Fix how JULIA_EDITOR is passed to REPL

davidanthoff:
Fixes https://discourse.julialang.org/t/error-ioerror-could-not-spawn-c-program-filesmicrosoft-vs-codecode-exe/19677.


skanskan:
David, thank you very much for your work and help.

PR #688, opened 2019-03-13
Search for `julia.executablePath` in PATH

### LINKED ISSUE #687: Allow specifying julia executable without full path
It would be great if `julia.executablePath` would search `$PATH` for the julia binary when the setting is not an absolute path, but just a binary name (i.e. `"julia.executablePath": "julia-1.0"`).

Currently the full path has to be specified, making it hard to sync configuration across machines, where julia might be installed in different places.

tribut:
This allows specifying the julia binary in `julia.executablePath` without full path by searching the PATH environment variable.

Fixes #687

Note that I did not update `package-lock.json` because `npm install` removes all the caret symbols from version specifications for me, probably because I'm [using a different npm version](https://github.com/npm/npm/issues/20434).

### DIFF
--- a/src/juliaexepath.ts
+++ b/src/juliaexepath.ts
@@ -4,7 +4,10 @@ import * as vslc from 'vscode-languageclient';
 import * as os from 'os';
 import * as path from 'path';
 import * as process from 'process';
+import * as util from 'util';
+import * as which from 'which';
 var exec = require('child-process-promise').exec;
+const whichAsync = util.promisify(which);
 
 let g_context: vscode.ExtensionContext = null;
 let g_settings: settings.ISettings = null;
@@ -58,7 +61,12 @@ export async function getJuliaExePath() {
             }
         }
         else {
-            actualJuliaExePath = g_settings.juliaExePath;
+            if (g_settings.juliaExePath.includes(path.sep)) {
+                actualJuliaExePath = g_settings.juliaExePath;
+            } else {
+                // resolve full path
+                actualJuliaExePath = await whichAsync(g_settings.juliaExePath);
+            }
         }
     }
     return actualJuliaExePath;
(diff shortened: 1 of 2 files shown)


davidanthoff:
Do we actually really need this? For example, on my Ubuntu machine I have the setting as just `julia`, and it seems to work just fine?

tribut:
Well, it doesn't work for me. That's why I created the PR. I've set `"julia.executablePath": "julia-1.0",` and `julia-1.0` is in the path obviously (and it works with the PR applied). The result:

Screenshot from 2019-03-15 23-25-16

Not sure why it works for you. Same result with `julia`, btw.

davidanthoff:
I found this: https://github.com/Microsoft/vscode-languageserver-node/issues/358. Are your `julia-1.0` and `julia` commands shell aliases or something like that?

tribut:
No, they are both symlinks to the respective julia binaries. Not sure if I understand the bug report correctly, but there is no shell script involved here.

~~~
[felix:~] $ readlink -e $(which julia-1.0)
/home/felix/.local/opt/julia-1.0/bin/julia
~~~

src/juliaexepath.ts
```diff
@@ -58,7 +59,12 @@ export async function getJuliaExePath() {
             }
         }
         else {
-            actualJuliaExePath = g_settings.juliaExePath;
+            if (g_settings.juliaExePath.includes(path.sep)) {
+                actualJuliaExePath = g_settings.juliaExePath;
+            } else {
+                // resolve full path
+                actualJuliaExePath = which.sync(g_settings.juliaExePath, {nothrow: true});
```
davidanthoff: Do you think we could use [this](http://2ality.com/2017/05/util-promisify.html#using-an-async-function) technique to actually use the async version here? I've tried pretty hard so far to not use sync versions of anything throughout the extension.

tribut: Sure, I've updated the PR accordingly.

PR #690, opened 2019-03-13
Use environment from current working directory

### LINKED ISSUE #661: show current working directory's env?
The environment only shows what's in `.julia`, but it would be convenient if we could show (or just use it as default) when a standard julia project is detected, e.g

when open a `JuliaProject` folder which contains `Project.toml`

```sh
code JuliaProject
```

The environment is changed to this project's env by default, and will be show in the menu instead of pick a folder.

tribut:
Started working on this before realizing that LanguageServer's `julia/activateenvironment` [is a stub](https://github.com/JuliaEditorSupport/LanguageServer.jl/blob/master/src/requests/misc.jl#L107-L111).

Apart from the fact that - because of this - it really doesn't do much, the code works. The correct list of environments is offered and the environment on startup is sensible.

That said, this is a breaking change in case there are some people out there who want to always start in the default environment. If you think this is necessary, I'll add a config option for it.

Fixes #661


tribut:
Sorry for the force-pushes, I've cleaned up the variables names a bit to be more in line with the rest of the code.

davidanthoff:
Cool!

So the main tricky thing is that I don't actually know what the best behavior is... Here is my current thinking:
- if I open a folder that is a project (i.e. not a package) and it has a `Project.toml` and `Manifest.toml` in it, I definitely think it should become the active project.
- But if I open a package that is deved, I actually _don't_ want to activate an environment based on that packages `Project.toml`, I think. In that case I'd want the active project to be an environment that includes the package that I'm editing. The tricky thing of course is that it is not clear what that environment is, and in fact it might not exist.

Maybe the best option would be to narrow this PR so that it only applies in situations where there is _also_ a `Manifest.toml` around? That should exclude the package situation, but cover the project situation.

For the package story, I'm not sure what to do... Maybe it should actually just be a workspace setting?

tribut:
Well, the reason I built it the way it is now was because it is what julia does (with the `--project` flag). The way it's implemented in this PR, it will always behave exactly like julia would. To me at least this satisfies the principle of least surprise[tm].

Also, it does not seem to be necessarily the case that dev'd packages only have `Project.toml` checked in: https://discourse.julialang.org/t/does-manifest-toml-belong-in-the-repository/12029

Lastly, it is trivial to switch to a parent repository using the dropdown menu.

tribut:
Even after reading #661 again, I still don't really understand the motivation of *not* activating a project when I explicitly open it in VSCode, but instead work inside the parent project.

My thinking is this: When I open `code myapp/dev/somepackage` I want to work on `somepackage`, maybe because it is missing a feature or maybe because I found a bug. Would I not want `Pkg.test()` to run the tests of that package then (as opposed to the tests of my app) to check if I successfully repaired the bug? Would I not want to have the dependencies of that package in scope to implement my new feature?

Sure, after fixing the bug/adding the feature I might want to go back to my app, but would I not want to have my files available so I can run the script that formally failed or even add an integration test? That is, would that not happen in the context of `code myapp`?

I feel like I'm missing something.

davidanthoff:
Let me ponder this a bit more. I might well be wrong with my reasoning here.

PR #695, opened 2019-03-31
Wrap terminalserver in a monitor process. Fixes #653.

### LINKED ISSUE #653: Leave terminal window open (or at least capture the stacktrace) when julia has a fatal error.
The julia terminal closes when the julia process crashes. It would be useful to be able to see the stacktrace that gets dumped when this happens.

twadleigh:
The terminal server is wrapped in a monitor process. When the server process exits with an error code, the user is asked whether to restart it. This allows the user to see the output of a failing process before the terminal window closes (addressing #653).

Optionally, and by default, the user is prompted for a restart even on a successful termination of the terminal server process. A setting has been added, however, to allow the user to opt out of the prompt on successful termination, yielding a behavior of the terminal more like that before the addition of the monitor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -326,6 +326,11 @@
                     "type": "boolean",
                     "default": true,
                     "description": "Display plots within vscode."
+                },
+                "julia.closeTerminalOnSuccess": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "When enabled, the Julia terminal will close without prompting for a restart if the Julia process terminated with success."
                 }
             }
         },
--- a/scripts/terminalserver/terminalserver_monitor.jl
+++ b/scripts/terminalserver/terminalserver_monitor.jl
@@ -1,0 +1,14 @@
+#=  Launches the terminal server in a child process. If the child process errors
+out, it prompts the user to dismiss the terminal, thereby preserving any output.
+=#
+try
+    opts = Base.JLOptions()
+    bin = unsafe_string(opts.julia_bin)
+    prj = unsafe_string(opts.project)
+    script = joinpath(@__DIR__, "terminalserver.jl")
+    run(`$bin -q -i --project=$prj $script $ARGS`, wait = true)
+catch e
+    @error e
+    println("\nPress ENTER to dismiss terminal.")
+    readline()
+end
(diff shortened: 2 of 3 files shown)


davidanthoff:
Hm, I have to admit that I find this a bit heavy handed... I also worry that it adds quite a bit of startup time, because now we are starting two julia instances, which at least on Windows is a problem. If this was for a very common scenario, I think the situation would be different, but this essentially only covers cases where there is a real bug in julia itself, right? So hopefully this would be really a corner case situation?

I could see two alternatives: I think the best one would be to actually lobby for/add an option to VS Code itself to not close the terminal UI when the terminal process closes. That seems like the best option to me.

Alternatively, I guess we could have a user option that enables this behavior, off by default, so that by default users don't get this monitor process. But let me think about that a bit more, I am a bit worried about the added code complexity for something that seems really more like a corner case...

twadleigh:
More than just bugs in the julia implementation can cause julia to crash. Bugs in external shared libraries that julia code might call into can, as well. I do a lot of development between C++ and Julia, so I see this quite a lot. Not really a corner case for me.

Using julia to monitor julia is, granted, potentially a bit heavyweight. The monitor process doesn't do anything particularly julian. It could be rewritten as a shell script, for instance. That would mean, though, maintaining parallel implementations across platforms. I went with julia mostly because it is automatically portable. I'm not super familiar with typescript, but, if it were capable of doing this job (which I would guess it probably is) that would be a good candidate for reimplementing the monitor.

I would also note, though, that the added query-for-restart logic amortizes that additional startup time over an arbitrary number of restarts. 

Also, even without crashes I still frequently kill and restart my julia REPL, though, of course, Revise has certainly lessened the pressure to do so. The restart logic means you can do that from the terminal command line without having to kill and reopen it. That is a pattern that Juno uses, and I have found it convenient. This PR would bring the same to julia-vscode.

twadleigh:
I will see about rewriting the monitor implementation in typescript.

davidanthoff:
Could we first investigate whether there is a chance to address this at the VS Code level? I think just opening an issue on their repo and asking would be great.

twadleigh:
I'm going to close this for now, pending https://github.com/Microsoft/vscode/issues/70444.

davidanthoff:
Perfect! If upstream drags out, we can revisit this fix here!

PR #709, opened 2019-04-11
Better parsing of triple quoted strings

haberdashPI:
This change ensures triple quoted  standard and non-stadard string literals are parsed properly. Before this change """bla bla bla""" would be parsed as three separate string literals: [""]["bla bla bla"][""] rather than as one string literal ["""bla bla bla"""].

This also changes non-standard string literals, highlighting the prefix with the support.function.macro color.

I ran into this problem while looking to create a small vscode plugin to allow the injection of other languages within Julia (e.g. to properly parse mat"x + 1" as matlab code and py"x + 1" as python code, etc...). Without this change it is quite hard to avoid parsing errors when injecting the grammar (because it is hard to find the triple quoted sections, as they are labeled as three different, ungrouped entities).

### DIFF
--- a/syntaxes/julia.json
+++ b/syntaxes/julia.json
@@ -493,7 +493,7 @@
           ]
         },
         {
-          "begin": "\"\"\"",
+          "begin": "(?<![[:word:]⁺-ₜ!′∇])\"\"\"",
           "beginCaptures": {
             "0": {
               "name": "punctuation.definition.string.multiline.begin.julia"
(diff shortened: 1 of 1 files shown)


davidanthoff:
Great, thanks!!

There is one caveat: could you make these edits over in https://github.com/JuliaEditorSupport/atom-language-julia/tree/master/grammars? We regularly just copy the file from there into VS Code here (via a CSON->JSON converter), so generally the workflow is that any changes should be made in atom-language-julia, and then they will show up in VS Code as well eventually, once we copy the file.

haberdashPI:
The changes have merged in the atom repository. Can you just let me know when they are available in this repository? Thanks!

PR #712, opened 2019-04-15
add delete all plots

### LINKED ISSUE #174: Add option to clear all plots
Just add a button that deletes all plots.

clouds56:
fixes #174

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,6 +150,14 @@
                 }
             },
             {
+                "command": "language-julia.plotpane-delete-all",
+                "title": "Julia: Delete all plots",
+                "icon": {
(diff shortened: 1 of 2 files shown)


clouds56:
The icon is now reusing "delete.svg" since I couldn't manage to draw a new one.

davidanthoff:
@clouds56: Thanks, looking good!

I made a few adjustments:
- Merged `master` in so that we are up to date
- Removed the icon (I don't think we should have the same icon appear twice next to each other)
- Moved the command into the menu that opens if one clicks `...`
- Enabled telemetry

I'm going to merge this now, if you think we should do more or refine, please just open another PR!

PR #720, opened 2019-05-12
Streamline communication with REPL

davidanthoff:
Only use one named pipe for all messages from julia to the extension and make the communication protocol more robust. Replaces #495.

@ZacLN I'm probably going to merge pretty soon because this is only the first step in a series of things I want to change. In particular, I want to finally change our code execution model so that we no longer just send the text to the REPL, but send the code via the pipe and can then execute it with more context information (so that things like `@__DIR__` etc work).


ZacLN:
Looks great, feel free to strip out anything that's not been used in a while (e.g. "repl/variables", etc.)

PR #726, opened 2019-05-16
Grid view

davidanthoff:
This one actually works :) Pretty much any source from the [Queryverse](https://www.queryverse.org/) should already show in a table.

Still some stuff to do:

- [x] Compat with all TableTraits.jl sources
- [x] Proper scrolling 
- [x] Add a proper API and don't hook into the default display stack for this

EDIT: Alright, this is ready for folks to try out. Any reports of success or failure with this branch would be great. The general pattern to use this is:
```julia
table |> vscodedisplay()
# or
vscodedisplay(table)
```
`table` can be pretty much anything that is a table in julia (in particular anything that implements the [TableTraits.jl](https://github.com/queryverse/TableTraits.jl) interface, which should be pretty much everything). So try it out with DataFrames, Query, QueryTables, IndexedTables, Pandas, the whole queryverse file IO story etc!


ZacLN:
Nice! Works without any problems. Is it possible to match themes?

davidanthoff:
Probably somehow :) I think the grid does somehow support themes, but I haven't looked into the details. https://github.com/JuliaEditorSupport/julia-vscode/issues/734.

PR #729, opened 2019-05-20
Add environmentPath configuration setting

### LINKED ISSUE #728: Setting for specifying path to Environment to be activated
Using the Command Palette for Julia, we can change the current environment which is a very useful feature when working with multiple workspaces, each having it's own Julia environment created as per https://julialang.github.io/Pkg.jl/v1/environments/ .
But currently, there is no way to specify which environment to load by default for which workspace. It would be great if there is a Setting provided where we can specify the path to the environment to be loaded for the current User / Workspace / Folder.
This Setting could then be used internally to start Julia with the required environment using the `--project` flag on running F5.

davidanthoff:
Fixes #728.

@tribut This takes the next step from your PR we recently changed. The main difference now is that I no longer activate auto activate an environment if there is no `Manifest.toml` present. I played with the implementation that did for a while, and I think one ends up with a sub-par experience. But, one can still pretty easily opt-in to opening a package and automatically getting that `Project.toml` activated as an environment: one just needs to instantiate the env once (thus creating a `Manifest.toml`) and then it will get auto activated.

But the main feature of this PR is that it adds a workspace level setting that stores the env that one selects via the UI, so that this setting gets persisted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -348,6 +348,12 @@
                     "type": "array",
                     "default": [],
                     "description": "Additional julia arguments."
+                },
+                "julia.environmentPath": {
+                    "type": ["string", "null"],
+                    "default": null,
+                    "description": "Path to a julia environment.",
+                    "scope": "window"
                 }
             }
         },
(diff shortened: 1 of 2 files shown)


tribut:
Looks fine to me. Still don't fully understand the reason behind the "Manifest behavior", but this should work for everyone.

PR #747, opened 2019-06-12
Fix code loading

davidanthoff:
This should generally be more standard and work with packages that only have a `Project.toml`.


davidanthoff:
Merging for now, given that `master` is broken no matter what without this.

ZacLN:
Yep seems to work for me

PR #755, opened 2019-06-14
Make sure all background in weave preview is white

### LINKED ISSUE #737: Weave preview shows margin bars
The new WebView #736 for Weave produces wide bars along the left and right margins that cannot seem to be removed by  resizing the panel.

An image makes the problem clearer:

https://pasteboard.co/IgEwLt3.png

davidanthoff:
Fixes #737.

### DIFF
--- a/scripts/weave/preview.tpl
+++ b/scripts/weave/preview.tpl
@@ -21,7 +21,7 @@
   {{{ :highlightcss }}}
 
 </HEAD>
-  <BODY>
+  <BODY style="background-color: white">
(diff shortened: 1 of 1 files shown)


pablosanjose:
Does this simply change the color or the margins? Wouldn't it be better to remove those extra margins altogether?

davidanthoff:
Yes, this just changes the color :)

I think I copied the template directly over from Weave at some point? A PR that improves the layout situation would be welcome!

PR #759, opened 2019-06-16
Add highlighting for several languages embedded as strings

tshort:
Adds support for non-standard string literals for `js`, `R`, `py`, and `md`. This catches us up with additions to the grammar for atom-language-julia.

In addition to highlighting, snippets and language-appropriate commenting work as expected for the embedded language.

Pinging @haberdashPI as he has several extensions that add most of these separately. Given that Atom is including them by default, I thought maybe we should, too. If there's a consensus, I can also add `mat"` (one that @haberdashPI has that is not implemented here). David, you used more complicated regex's. If those fix problems over the basic versions included here, please bring that up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,14 @@
             {
                 "language": "julia",
                 "scopeName": "source.julia",
-                "path": "./syntaxes/julia.json"
+                "path": "./syntaxes/julia.json",
+                "embeddedLanguages": {
+                    "meta.embedded.block.cpp": "cpp",
+                    "meta.embedded.block.js": "javascript",
+                    "meta.embedded.block.markdown": "markdown",
+                    "meta.embedded.block.python": "python",
+                    "meta.embedded.block.r": "r"
+                }
             },
             {
                 "language": "juliamarkdown",
--- a/syntaxes/julia.json
+++ b/syntaxes/julia.json
@@ -408,7 +408,7 @@
             }
           },
           "name": "embed.cxx.julia",
-          "contentName": "source.cpp",
+          "contentName": "meta.embedded.block.cpp",
           "patterns": [
             {
               "include": "source.cpp"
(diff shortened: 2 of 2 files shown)


haberdashPI:
For each of my plugins, I tried the simpler, more obvious regex's, and they did not work for me.

However, things may have improved with newer versions of a particular language plugin or with recent changes to the Julia syntax definition; I would just be careful to test each plugin with at least these cases:

   - A full file where the embedded language is multiline inside of a function 
   - A full file where the embedded language is single-line inside of a function
   - A full file where the embedded language is multiline at the top-level
   - A full file where the embedded language is single-line at the top-level 

The amount of indentation can matter for whether the form works or not. The problems I ran into were evident in some but not all of those cases. In some cases (python) I could not get all cases to work, and selected a regex that ensured *some* of the cases worked, without getting "bleed-out": a case where the syntax highlighting interprets the closing token as a part of the embedded language.

tshort:
Thanks, @haberdashPI. That's good advice. Here's an example I made to test some of those out:

https://gist.github.com/tshort/183d3d2095ae0945b937f660c9df4808

At the top level, everything worked pretty well. The only one that seems a bit messed up to me is the Cxx one (the one already implemented).

Inside of the function, most work fine. Multiline for Python is the one that has an irritation where the trailing triple quotes are not right (bleed out to following code) unless they are outdented to the far left. I'll try to look at that more. 

Multiline for Markdown inside a function is a little wonky in that indentation messes it up. That's probably to be expected, though. It also has a bit of a problem with the final triple quotes.

tshort:
Also, commenting and snippets work right in each of the multi-line blocks.

davidanthoff:
I think I'm in favor, but one problem is that I regularly just copy the syntax file from the atom language extension to the repo here, so any custom changes we make in `julia.json` would ideally be made in https://github.com/JuliaEditorSupport/atom-language-julia, and then we can just copy that over here. Or is that what happened for these changes already?

tshort:
That’s the case. `atom-language-julia` added these in the last couple of months.

PR #761, opened 2019-06-18
Support for custom sysimg

davidanthoff:
This adds experimental support for custom sysimages. There are two parts to this:
- A task that builds a custom sysimage for the current environment and stores that as `JuliaSysimage.dll` alongside the `Project.toml`.
- A configuration setting (off by default) that enables a logic where we start new REPLs with a custom sysimage if there is a file `JuliaSysimage.dll` in the same folder as the `Project.toml`.

This all only works if PackageCompiler.jl is installed. I might look into ways to cut that dependency at some point in the future, but I think for now this is ok, given that all of this is labeled experimental.

@KristofferC if you have thoughts on this, I'd be really interested to hear them!

### DIFF
--- a/package.json
+++ b/package.json
@@ -357,6 +357,12 @@
                     "default": null,
                     "description": "Path to a julia environment.",
                     "scope": "window"
+                },
+                "julia.useCustomSysimage": {
+                    "type": "boolean",
+                    "default": "false",
+                    "description": "Use an existing custom sysimage when starting the REPL",
+                    "scope": "application"
                 }
             }
         },
--- a/scripts/tasks/task_compileenv.jl
+++ b/scripts/tasks/task_compileenv.jl
@@ -1,0 +1,52 @@
+using Pkg, Libdl
+import PackageCompiler
+
+pkg_ctx = Pkg.Types.Context()
+
+mktempdir() do temp_dir
+    precompile_temp_file_name = joinpath(temp_dir, "temp_file_for_compile.jl")
+
+    used_packages = Set(keys(pkg_ctx.env.project.deps))
+
+    usings = """
+        using $(join(used_packages, ','))
+        for Mod in [$used_packages]
+            isdefined(Mod, :__init__) && Mod.__init__()
+        end
+        """
+
+    open(precompile_temp_file_name, "w") do out_file
+        println(out_file, """
+            # We need to use all used packages in the precompile file for maximum
+            # usage of the precompile statements.
+            # Since this can be any recursive dependency of the package we AOT compile,
+            # we decided to just use them without installing them. An added
+            # benefit is, that we can call __init__ this way more easily, since
+            # incremental sysimage compilation won't call __init__ on `using`
+            # https://github.com/JuliaLang/julia/issues/22910
+            $usings
+            # bring recursive dependencies of used packages and standard libraries into namespace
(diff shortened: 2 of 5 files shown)


davidanthoff:
I'm merging for now. This stuff is all behind a feature flag, and I think we are better off putting this into the beta and then iterating based on feedback.

tkf:
Hi, I just found it via PackageCompiler.jl issue.

> * A configuration setting (off by default) that enables a logic where we start new REPLs with a custom sysimage if there is a file `JuliaSysimage.dll` in the same folder as the `Project.toml`.

This is very similar to what I'm doing with `jlm`:

* [ANN: a solution to the precompilation problem: JuliaManager.jl / jlm CLI, a system image manger for Julia - Community / Package announcements - JuliaLang](https://discourse.julialang.org/t/ann-a-solution-to-the-precompilation-problem-juliamanager-jl-jlm-cli-a-system-image-manger-for-julia/22849)
* [jlm: System image manager for Julia — jlm 0.2.0-DEV documentation](https://jlm.readthedocs.io/en/latest/)

There is one indirection in `jlm` as it stores the path to a system image in `.jlm/data.json`.  Maybe we can combine the effort here, so that the same system image setup can be used outside vscode as well?  I'm open to rename `.jlm` directory name and change the schema of `.jlm/data.json`.  Currently, the content is something like

```json
{
  "name": "jlm.LocalStore",
  "jlm_version": "0.1.2-DEV",
  "config": {
    "runtime": {
      "/home/takafumi/repos/watch/julia/usr/bin/julia": {
        "sysimage": "/home/takafumi/repos/hack/julia-sysimage-recipes/all2/build/1.2.0/sys.so"
      },
      "/home/takafumi/opt/julia/julia-1.1.1/bin/julia": {
        "sysimage": "/home/takafumi/repos/hack/julia-sysimage-recipes/all2/build/1.1.1/sys.so"
      }
    },
    "default": "/home/takafumi/opt/julia/julia-1.1.1/bin/julia"
  }
}
```

tkf:
You might also be interested in the "gorilla patch" I'm using to avoid recurring precompilations when using different system images: https://github.com/JuliaLang/julia/issues/27418#issuecomment-499306707

davidanthoff:
I think I don't really understand that, so if there is something that we could do here, I would probably need some help to understand the situation better first :)

tkf:
`jlm` is just a command line entry point to launch `julia` (plus some utilities).  `jlm run ...` is equivalent to `julia -J PATH/TO/sys.so ...`.  It seems that you are doing something very similar.  I thought it would be nice to create a common Julia launcher spec that works anywhere, including VS code.

davidanthoff:
Ah, yes! Agreed, it would be nice if we could align. Are you coming to juliacon? Maybe this is something we can sort out together there?

tkf:
Yes, I'm coming to juliacon. Do you want to setup a quick meeting using Discourse DM or email?  Or maybe we'll just come across during juliacon anyway at some point (I'm thinking to go to your talk).

davidanthoff:
Lets try to just bump into each other :) I'm only arriving Tue late evening, so I'll be there Wed-Fri (including the hackathon).

tkf:
Sounds good! I'll be there except workshop (until hackathon).

PR #762, opened 2019-06-19
Fixed greedy highlighting of embedded C++.

twadleigh:
Also, updated the syntax file from atom-language-julia.

PR #763, opened 2019-06-19
Copy latest syntax from Atom

davidanthoff:
@twadleigh, this is what I get if I just convert the latest version of https://github.com/JuliaEditorSupport/atom-language-julia/blob/master/grammars/julia.cson to JSON. It is different than what you had put into https://github.com/julia-vscode/julia-vscode/pull/762. So, am I right to assume that you a) copied the syntax from atom, and then b) also made some manual modifications?

We generally can't carry manual modifications here in the repo. I want to be able to just periodically copy the latest version from atom, without any manual review. So I think if we want these manual modifications, then ideally you would submit them as a PR against the original, upstream https://github.com/JuliaEditorSupport/atom-language-julia/blob/master/grammars/julia.cson, and then they will flow back to here when we copy the syntax file over again.

### DIFF
--- a/syntaxes/julia.json
+++ b/syntaxes/julia.json
@@ -407,10 +407,11 @@
               "name": "punctuation.definition.string.end.julia"
             }
           },
-          "contentName": "meta.embedded.block.cpp",
+          "name": "embed.cxx.julia",
+          "contentName": "source.cpp",
           "patterns": [
             {
-              "include": "source.cpp#root_context"
+              "include": "source.cpp"
             },
             {
               "include": "#string_dollar_sign_interpolate"
@@ -428,15 +429,16 @@
             }
           },
           "end": "\"",
+          "name": "embed.cxx.julia",
           "endCaptures": {
             "0": {
               "name": "punctuation.definition.string.end.julia"
             }
           },
-          "contentName": "meta.embedded.block.cpp",
+          "contentName": "source.cpp",
           "patterns": [
             {
-              "include": "source.cpp#root_context"
+              "include": "source.cpp"
             },
             {
               "include": "#string_dollar_sign_interpolate"
@@ -459,7 +461,8 @@
               "name": "punctuation.definition.string.end.julia"
             }
           },
-          "contentName": "meta.embedded.block.python",
+          "name": "embed.python.julia",
+          "contentName": "source.python",
           "patterns": [
             {
               "include": "source.python"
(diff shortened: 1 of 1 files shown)


twadleigh:
I chose these hand-editted mods because:
1. they appeared to be more consistent with the master version of the syntax file;
2. they also appeared to be more consistent with the current `package.json`.

I will happily redo this PR with the unmodified output of `cson2json`.

davidanthoff:
Ok, I already replaced things with the unmodified version from cson2json and that is what we have right now on `master`.

Can you elaborate a bit more on

> they also appeared to be more consistent with the current `package.json`

Do we have a problem there right now?

twadleigh:
I was trying to stick with the use of `meta.embedded.block.*` that was present in the old syntax and `package.json`. My experimentations, though, suggested that that particular convention provided no special benefit.

davidanthoff:
Ok, great! I have to admit I don't really understand that part of the extension, I'm just copying things over generally, so any help is very much appreciated!

twadleigh:
BTW, the single most important edit didn't make it into master. The `"include": source.cpp` lines need to be changed to `"include": source.cpp#root_context`. This is because the vscode version of `source.cpp` has a top-level production called `#source_wrapper` which purposely matches the entire input string (running past the embedding delimiter).

Cf. https://github.com/jeff-hykin/cpp-textmate-grammar, from which I believe vscode's `source.cpp` is derived. The author claims that the `#source_wrapper` production is actually to make embedded C++ _better_, at least for markdown. I could see that maybe it could be useful in the case of markdown, whose syntax is best expressed with `begin/while` rules, but breaks things for grammars with real delimiters that are expressible in terms of `begin/end` rules.

The fix is to bypass the greedy match of `#source_wrapper` and go straight to `#root_context`.

twadleigh:
There is no problem with `package.json` _per se_, but the transliterated syntax obviates the `embeddedLanguages` clause.

tshort:
The following in `package.json` fixes most of the changes to use the raw conversion.

```json
                "embeddedLanguages": {
                    "source.cpp": "cpp",
                    "source.js": "javascript",
                    "source.gfm": "markdown",
                    "source.python": "python",
                    "source.r": "r"
                },
                "tokenTypes": {
                    "source.cpp": "other",
                    "source.js": "other",
                    "source.gfm": "other",
                    "source.python": "other",
                    "source.r": "other"
                }
```

The markdown support is broken because Atom uses `source.gfm`, and VSCode uses `text.html.markdown`. I tried including both into Atom and VSCode, and it seemed to work, but I'm not sure if that'd mess up other applications.

twadleigh:
@tshort I'll test that out on my end.

twadleigh:
@tshort C++ embedding still suffers from the greedy match.

Also, I looked at the grammar in `atom-language-c`. It doesn't contain a single top-level production, let alone one named `#root_context`, so there is no patch to be made upstream at `atom-language-julia`.

I might look in to writing a conversion script or creating a grammar that embeds the transliterated grammar from `atom-language-juila` (or both).

PR #766, opened 2019-06-19
Adds automatic transformation of atom-language-julia grammar.

### LINKED ISSUE #765: Embedded C++ breaks syntax highlighting.
I have captured an example of the failure in https://github.com/twadleigh/julia-syntax-test-cases/blob/master/embedded_code.jl. From line 43 on the code is highlighted as though it is mostly a C++ string.

For the record, https://github.com/twadleigh/julia-syntax-test-cases/blob/master/corner_cases.jl also shows another leaky failure.

### LINKED ISSUE #770: Get the markdown injection grammar to apply to markdown embedded in julia.
It currently works on `*.md` files, but apparently not yet in `md"..."` blocks in `*.jl` files.

twadleigh:
- Updates the syntax using the new script.
 - Fixes #765.
 - Fixes #770.
 - Adds an injection grammar for highlighting julia code blocks in markdown.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,11 +70,11 @@
                 "scopeName": "source.julia",
                 "path": "./syntaxes/julia.json",
                 "embeddedLanguages": {
-                    "meta.embedded.block.cpp": "cpp",
-                    "meta.embedded.block.javascript": "javascript",
-                    "meta.embedded.block.markdown": "juliamarkdown",
-                    "meta.embedded.block.python": "python",
-                    "meta.embedded.block.r": "r"
+                    "meta.embedded.inline.cpp": "cpp",
+                    "meta.embedded.inline.javascript": "javascript",
+                    "meta.embedded.inline.markdown": "juliamarkdown",
+                    "meta.embedded.inline.python": "python",
+                    "meta.embedded.inline.r": "r"
                 }
             },
             {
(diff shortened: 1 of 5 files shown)


twadleigh:
I'm planning to add a little more syntax-related work to this before I'll consider it ready for merging.

davidanthoff:
This is absolutely awesome, thanks!! Just ping me when it is ready and I'll merge. I'll add this to the v0.12.0 milestone, because it would be nice to include that in the next release.

twadleigh:
There are still a couple tweaks I'd like to do here, but this should be usable as is. I'll open up separate issues to track the follow-ons.

twadleigh:
OK, I'm now really happy with this. It captures every possible desirable feature related to syntax that I could think of.

I would note, though, that there is one sense in which the changes here could be called "breaking". The code fence tokens for Julia-flavored markdown are now named consistently with all the other languages that can embed in markdown. That means, though, that the formatting of the code fences may be different for `*.jmd` files after this is merged.

I would assert that this sort of "breakage" is acceptable. However, it might be worth mentioning in any release notes.

davidanthoff:
> I would assert that this sort of "breakage" is acceptable. However, it might be worth mentioning in any release notes.

Yeah, I'm not worried about that. I'll add a generic `Reworked syntax highlighting` to the CHANGELOG, my sense is that that is enough.

Issue #560, opened 2018-10-21
Execute code cell


sdewaele:
A feature that would allow execution of a code cell would be great to have. The feature is available in Matlab and Spyder. It means that the code block starting with a certain marker, e.g. `#%%` would be executed in the REPL. BTW - I am rather new to VS code, so perhaps this feature is already available.

dgleich:
Just a quick note on this -- I use this feature extensively in the atom editor. Would love to see it in julia-vscode too so I can better evaluate a potential move. Here's one of the initial issues on the atom-julia-client for it. 

https://github.com/JunoLab/atom-julia-client/issues/63

If I get an itch to do something new at some point, I may take a closer look. 

Standard cell delimiters are: 

##
#%%

If I run into any others, I'll post them.

davidanthoff:
Yes, very much on my radar! There are some changes in the pipeline over in VS Code itself that will make this scenario a heck of a lot more powerful, I think once that lands (hopefully soon) I'll take a stab at this.

sdewaele:
Thanks for implementing this! (same as #777)

PR #777, opened 2019-06-25
Add cell-wise evaluation.

twadleigh:
Addresses #560.

Possible TODOs:
~- Add more delimiters.~
~- Make delimiters configurable.~
- [x] Add default keybinding.
- [ ] Decide on right family of keybindings.
- [ ] Add echo in the REPL on code chunk execution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "activationEvents": [
         "onCommand:language-julia.openPackageDirectory",
         "onCommand:language-julia.startREPL",
+        "onCommand:language-julia.executeJuliaCellInREPL",
         "onCommand:language-julia.executeJuliaCodeInREPL",
         "onCommand:language-julia.weave-open-preview",
         "onCommand:language-julia.weave-open-preview-side",
@@ -179,6 +180,10 @@
                 "title": "Julia: Toggle server logs"
             },
             {
+                "command": "language-julia.executeJuliaCellInREPL",
+                "title": "Julia: Execute code cell"
+            },
+            {
                 "command": "language-julia.executeJuliaBlockInREPL",
                 "title": "Julia: Execute code block"
             },
@@ -250,6 +255,11 @@
             {
                 "command": "language-julia.executeJuliaCodeInREPL",
                 "key": "ctrl+Enter",
+                "when": "editorTextFocus && editorLangId == julia"
+            },
+            {
+                "command": "language-julia.executeJuliaCellInREPL",
+                "key": "shift+Enter",
                 "when": "editorTextFocus && editorLangId == julia"
             },
             {
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -612,6 +612,37 @@ async function selectJuliaBlock() {
     }
 }
 
+async function executeJuliaCellInRepl() {
+    telemetry.traceEvent('command-executejuliacellinrepl');
+
+    let ed = vscode.window.activeTextEditor;
+    let doc = ed.document;
+    let rx = new RegExp("^(##|#---|#%%|# %%)"); // atom-julia-client defaults
+    let curr = ed.selection.active.line;
+    var start = curr;
+    while (start >= 0) {
+        if (rx.test(doc.lineAt(start).text)) {
+            break;
+        } else {
+            start -= 1;
+        }
+    }
+    start += 1;
+    var end = start;
+    while (end < doc.lineCount) {
+        if (rx.test(doc.lineAt(end).text)) {
+            break;
+        } else {
+            end += 1;
+        }
+    }
+    end -= 1;
+    let startpos = new vscode.Position(start, 0);
+    let endpos = new vscode.Position(end, doc.lineAt(end).text.length);
+    let code = doc.getText(new vscode.Range(startpos, endpos));
+    executeCode(code, false)
+}
+
 async function executeJuliaBlockInRepl() {
     telemetry.traceEvent('command-executejuliablockinrepl');
 
@@ -681,6 +712,8 @@ export function activate(context: vscode.ExtensionContext, settings: settings.IS
     context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaCodeInREPL', executeSelection));
 
     context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaFileInREPL', executeFile));
+
+    context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaCellInREPL', executeJuliaCellInRepl));
 
     context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaBlockInREPL', executeJuliaBlockInRepl));
 
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -573,6 +573,12 @@ function executeSelection() {
     executeCode(text, selection.isEmpty)
 }
 
+async function executeInRepl(code: string, filename: string, start: vscode.Position) {
+    let msg = filename + '\n' + start.line.toString() + ':' +
+        start.character.toString() + '\n' + code
+    sendMessage('repl/runcode', msg)
+}
+
 async function executeFile() {
     telemetry.traceEvent('command-executejuliafileinrepl');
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -252,6 +257,11 @@
                 "key": "ctrl+Enter",
                 "when": "editorTextFocus && editorLangId == julia"
             },
+            {
+                "command": "language-julia.executeJuliaCellInREPL",
+                "key": "shift+Enter",
```
davidanthoff: I think one worry here is that this is already a valid keybinding in the VS Code editor...

src/repl.ts
```diff
@@ -612,6 +612,37 @@ async function selectJuliaBlock() {
...
+            break;
+        } else {
+            end += 1;
+        }
+    }
+    end -= 1;
+    let startpos = new vscode.Position(start, 0);
+    let endpos = new vscode.Position(end, doc.lineAt(end).text.length);
+    let code = doc.getText(new vscode.Range(startpos, endpos));
+    executeCode(code, false)
```
davidanthoff: I think we should use the `sendMessage('repl/runcode', msg_body)` pattern instead, that makes for a much more stable code execution story. It would probably make sense to create a new function a la `executeCode2` (name really doesn't matter much at this point) that wraps the logic to execute code that way. Right now there are two places in the code that use this ([here](https://github.com/julia-vscode/julia-vscode/blob/9a2a2fbd839b79ce67de02216fe184ba0208b72c/src/repl.ts#L588) and [here](https://github.com/julia-vscode/julia-vscode/blob/9a2a2fbd839b79ce67de02216fe184ba0208b72c/src/repl.ts#L637)), and they could then also use such a `executeCode2` function. All that function would do is essentially wrap [this](https://github.com/julia-vscode/julia-vscode/blob/9a2a2fbd839b79ce67de02216fe184ba0208b72c/src/repl.ts#L633) stuff.

The benefit of that approach is that the code executed that way has line numbers associated with it, and things like `@__DIR__` work.

pfitzseb:
> I would almost prefer that both Juno and VS Code just have one, common pattern for that, and that is it. @pfitzseb, does that make sense to you?

It does, but unfortunately Juno already a) has this user configurable and b) supports `##, #---, #%%, # %%` out-of-the-box.

Imho everyone should just standardize on `##`, but I feel like it's a bit too late for this now :)

davidanthoff:
> Imho everyone should just standardize on `##`, but I feel like it's a bit too late for this now :)

Well, we don't have backwards compat issues here in VS Code land (yet) on this topic, so we could just support that one option, and maybe it will lead to folks using that more, and in a few years Juno can drop support for the other options as well? And maybe the Juno doc could emphasize one option, and strongly recommend to use that and not configure others or use others? I'm completely on board that Juno doesn't want to drop support for things that work and have been around for a while, but we could try to use messaging towards one option.

package.json
davidanthoff: One other idea: we could re-use Alt+Enter for this? If you are in a cell, it executes a cell, if not, it executes a code block? That would also make it compatible with Juno, at least partially?

pfitzseb:
Seems like a good idea, yeah.

package.json
twadleigh: `Shift+Enter` has no conflicts in the Julia context. There are analogous bindings for python in the `editorTextFocus` scope, but they don't conflict because of the distinct `editorLangId`. The only other binding is in the `inSettingsSearch` scope.

In contrast, `Alt+Enter` _would_ conflict with block execution. There is always a cell that can be uniquely associated with the cursor position.

src/repl.ts
twadleigh: I pushed a version of this in my latest commit. This was the first thing I tried, but I wasn't happy without having a visible sign that the cell was executed. It would be nice to have something visible happen at the REPL.

package.json
twadleigh: That said, I'd prefer if the bindings could be as consistent as possible with Juno.

twadleigh:
@davidanthoff and @pfitzseb agree: `^##` is now the only pattern on which a cell boundary will match in `julia-vscode`.

package.json
davidanthoff: Ah, you are right, somehow I thought `Shift+Enter` did something different in the editor, so that isn't a problem.

> In contrast, `Alt+Enter` would conflict with block execution. There is always a cell that can be uniquely associated with the cursor position.

I guess this would amount to saying that if a file has even just a single `##` in it, then `Alt+Enter` executes cells, otherwise it executes code blocks. That doesn't seem like a great choice, though.

So maybe we just stick with the choice you made here for now, and then at some point in the distant future we can re-evaluate all the code execution keybindings.

src/repl.ts
davidanthoff: Yeah, I was unsure whether to echo code back or not. [This](https://github.com/julia-vscode/julia-vscode/blob/9a2a2fbd839b79ce67de02216fe184ba0208b72c/scripts/terminalserver/terminalserver.jl#L149) line does echo code back if it is not disabled.

I'm just a bit worried that it gets unwieldy when one executes a whole file or a lot of code... But maybe that would be ok. Or maybe we need some other visual identification that code is running.

@pfitzseb, are you always echoing all code back in the REPL that is being executed in Juno?

davidanthoff:
I've merged for now, we can iterate re echoing back stuff in separate PRs.

Thanks for the PR!

src/repl.ts
pfitzseb: I'm pretty sure we're *never* echoing executed code in the REPL.

> Or maybe we need some other visual identification that code is running.

VSCode does support custom status bar elements, right? Would maybe be useful to have something like Juno's [progress meter](http://docs.junolab.org/latest/man/juno_frontend/#Progress-Meters-1). Also, the REPL prompt is hidden when code is executed, so that should also be a good indicator that *something*'s happening.

davidanthoff: Yeah, I agree, a UI in the VS Code chrome would be nice. One thing, though: how could we hook into that if folks just type code into the REPL and run it? There are no callback hooks for the REPL a la "about to run user code", and "finished running user code", right?

pfitzseb: True, that's a bit messy.
Juno has [this abomination](https://github.com/JunoLab/Atom.jl/blob/92ad44109f24b80aa49f233dade4ae9579817f20/src/repl.jl#L139). I'm starting to think it might make sense to prototype some repl enhancements in a `REPLPlus.jl` package that could be used by Juno and VSCode and provides a bunch of hooks (and maybe even solves the whole async printing business properly).

davidanthoff: Ok, I looked at it, but I don't get how this helps with the user typing something in the REPL and pressing Enter. Does that code somehow hook into that?

> I'm starting to think it might make sense to prototype some repl enhancements...

Yes, that would be great! Or maybe we could even add these two simple hooks into julia base for a future version? If both VS Code and Juno could make use of such hooks, they shouldn't be too controversial?

pfitzseb: > Ok, I looked at it, but I don't get how this helps with the user typing something in the REPL and pressing Enter. Does that code somehow hook into that?

Yes, we're replacing the REPL `respond` function with that [here](https://github.com/JunoLab/Atom.jl/blob/92ad44109f24b80aa49f233dade4ae9579817f20/src/repl.jl#L187-L193).

> Yes, that would be great! Or maybe we could even add these two simple hooks into julia base for a future version? If both VS Code and Juno could make use of such hooks, they shouldn't be too controversial?

Sure, but Base/stdlibs are moving pretty slowly, so prototyping in an external package probabaly makes sense.

davidanthoff: Ah, ok, now I get it!

Yes, totally agreed, prototyping somewhere else probably is much easier.

twadleigh:
Two things I was mulling for this:
1. Maybe swap the keymappings for cell and block evaluation. I think that should be more consistent with Juno, which doesn't have a notion of block evaluation, but uses `Alt+Enter` for cell evaluation. Of course, that is a breaking change for `julia-vscode`.
2. About echo: Juno doesn't do it on the REPL, but it does give the result inline in the editor. Not having any visible feedback would degrade the value of that functionality for me. Or I would just have to get into the habit of putting `println`s at the end of all my cells. When [vscode let's us echo inline](https://github.com/microsoft/vscode/issues/3220) then we can do it the Juno way. Until then, I was hoping to at least have some feedback at the prompt.

davidanthoff:
I think aligning Juno und VS Code keymaps makes sense, so I think swapping bindings sounds fairly reasonable to me. I think maybe best if we open a PR for that and then ask for some feedback, and then based on that decide? But I think it is certainly worth exploring.

Re echo: ah, maybe the problem is that we not only not echo the code, but also don't show the result of the expression that is executed in the REPL? We could also make the echo thing a config setting?

Btw., I think the insider build already has support for arbitrary HTML output between lines, so I think we are getting closer to being able to do that kind of stuff!

PR #793, opened 2019-07-18
Add NUM_THREADS configuration option

ZacLN:


KristofferC:
The name "REPL threads" sounds a bit odd to me. Why include REPL at all?

ZacLN:
Because it's specific to the REPL (rather than the language server, test tasks, etc.). Happy to take other suggestions

KristofferC:
I think the implementation should be such that it is equivalent to having the environment variable. Otherwise, how would you eg run the tests of a package threaded?

davidanthoff:
I've assigned this to v0.13.0, ok? Lets get v0.12.0 out with the feature set we have first :)

davidanthoff:
So maybe call it just `NumThreads`, and then use it for more processes, I.e. test runs, doc build etc? The latter don’t have to be part of this PR, of course.

PR #795, opened 2019-07-18
Include the julia version in crash reports

davidanthoff:


scripts/languageserver/main.jl
```diff
@@ -33,6 +33,9 @@ catch e
 
         # Send error message
         temp_io = IOBuffer()
+        versioninfo(temp_io, verbose=true)
```
KristofferC: Is the verbose option really needed? It will send a bunch of extra info like env variables.

davidanthoff: I thought it wouldn't harm? The more info we have, the easier it is to diagnose what is going wrong.

KristofferC: > The more info we have, the easier it is to diagnose what is going wrong.

Sure, that's what they all say ;). I might have env vars called `JULIA_PASSWORD` which will now be sent. I think at least some kind of heads up to people about the additional info being now being sent in the change notes would make sense.

davidanthoff: Actually, the password story is pretty convincing, I'll make this non-verbose.

davidanthoff: https://github.com/julia-vscode/julia-vscode/commit/cc4fe5ecfb17397bf2f014cec2fd76ce88c206dc

PR #806, opened 2019-07-22
add "julia.format.xxx" config section

ZacLN:


davidanthoff:
Should the scope for these be Windows?

davidanthoff:
I'd say this is v0.13.0 stuff, after we put out a bug fix v0.12.1 release, ok?

davidanthoff:
I think we still need to sort out the scope question?

PR #818, opened 2019-07-30
WIP: documentation support

ZacLN:


src/smallcommands.ts
```diff
@@ -89,6 +89,29 @@ function toggleFileLint(arg) {
...
+        vscode.window.showErrorMessage('Error: Lanuage server is not yet running.');
+    }
+    else {
+        try {
+            g_languageClient.sendRequest("julia/help", searchString).then((v :string)=>{
+                vscode.window.showInformationMessage(v)
+            });
+        }
+        catch (ex) {
+            5
```
davidanthoff: Remove that line :)

src/smallcommands.ts
```diff
@@ -89,6 +89,29 @@ function toggleFileLint(arg) {
...
 
+async function openJuliaHelp() {
+    let searchString = await vscode.window.showInputBox()
+    if (g_languageClient == null) {
+        vscode.window.showErrorMessage('Error: Lanuage server is not yet running.');
+    }
+    else {
+        try {
+            g_languageClient.sendRequest("julia/help", searchString).then((v :string)=>{
+                vscode.window.showInformationMessage(v)
```
davidanthoff: We could also open this in a dedicated help pane?

ZacLN: Yep, as is this just spits out the documentation (markdown markings and all) into a tiny really un-useful panel - I assume a WebView pane would be able to handle markdown?

src/smallcommands.ts
ZacLN: this is an _essential_ part of the code

src/smallcommands.ts
davidanthoff: Hm, not sure, I think WebView only takes HTML... But there must be some facility to compile markdown into HTML in the VS Code code somewhere :)

NightMachinary:
I don't know if this is in the scope of the extension or VSCode, but

https://stackoverflow.com/questions/61324685/vscode-how-do-i-open-the-documentation-ala-show-hover-in-another-pane

> I like to be able to see the documentation of the symbol at cursor using a keyboard shortcut. I know of the command editor.action.showHover, but I can't scroll that popup's contents using the keyboard. In fact, I prefer to have a documentation pane open at the right side that just shows the docs for anything the cursor is on.

NightMachinary:
I have this in my `startup.jl`:

```
more(content) = more(repr("text/plain", content))
# using Markdown
# more(content::Markdown.MD) = more(Markdown.term(Core.CoreSTDOUT(), content))
function more(content::AbstractString)
    run(pipeline(`echo $(content)`, `less`))
    nothing
end
macro d(body)
    :(more(Core.@doc($(esc(body)))))
end
```

Is there a way I can add a shortcut that runs `@d <symbol-at-cursor>` in the REPL?

Issue #837, opened 2019-08-26
Make Alt-Enter work with selection


davidanthoff:
If the user has selected some code, just run that when Alt+enter is pressed.

PR #841, opened 2019-08-27
Change code execution keybindings

ZacLN:
When a range of text is selected Alt-Enter now executes that selection.


ZacLN:
https://github.com/julia-vscode/julia-vscode/issues/837

Issue #873, opened 2019-10-14
enum blockdeclaration (begin ... end) gives errors when no defaults are specified


radioflash:
This gives "missing reference" errors (both inside the declaration and wherever the literals are used):

```
@enum MyEnum begin
    ZERO
    ONE
end
```

When default values are specified, everything works as expected:
```
@enum MyEnum begin
    ZERO = 0
    ONE = 1
end
```

ZacLN:
Appears to be fixed by StaticLint updates

Issue #872, opened 2019-10-14
ccall keyword is not recognized


radioflash:
AFAIK, `ccall` is a keyword in Julia since v0.6.

The vscode extension does not respect it, though, and complains about "missing reference to ccall" wherever it is used.

Maybe this is just a matter of adding "ccall" to syntaxes/julia.json?
I will submit a pull request if someonce can confirm that the required change is simple (I lack insight into the codebase).

davidanthoff:
We just regularly copy over the syntax file from https://github.com/JuliaEditorSupport/atom-language-julia, so ideally you would make a PR against that repo, and then we would copy things over to here once that is done.

radioflash:
Some simple experiments led me to believe that the problem is with StaticLinter (not the syntax json file): Because the problem only shows after the file is open for some time (delay from linter?), while the syntax highlighting (powered by the syntax json?) kicks in immediately.

After looking into everything some more, it seems to me that both this issue and also the block enum one (#873) is already fixed in the newest StaticLinter version.

But from what I can tell the StaticLinter master branch only works with the SymbolServer master branch (that combination passes all of their respective tests).

The julia-vscode project is using an older StaticLinter and SymbolServer commits (SymbolServer also seems to have forked off?).

I've cloned the repo, changed both StaticLinter and SymbolServer to master branch and built the plugin file. This fixes this issue and also #873.

radioflash:
I think only these two commits would be lost by changing SymbolServer to master (everything else was just version bumping):

julia-vscode/SymbolServer.jl@5e403f5
julia-vscode/SymbolServer.jl@ae81128

Maybe the master branch makes these obsolete, anyway.

ZacLN:
Fixed https://github.com/julia-vscode/SymbolServer.jl/commit/d517c0c77edb3dd26325dbdc94afd912f05bc92d, will be in the next release

PR #875, opened 2019-10-15
Update StaticLint/SymbolServer versions

radioflash:
This lets the linter recognize the ccall keyword
and blockwise (`@enum MyEnum begin X1 X2 end`) enum definitions (among others).

See #872 and #873


davidanthoff:
Unfortunately we can't merge things right now... SymbolServer.jl has a bit of a versioning chaos right now: `master` has some stuff fixed, but then we also fixed things on a release branch, and before we can use `master` again we need to merge those two branches... Once we have consolidated that situation, we can then use the latest version here again.

PR #878, opened 2019-10-26
add "forceDebug" option to allow easy debugging for end users

runapp:


package.json
```diff
@@ -390,6 +390,11 @@
                     "default": "false",
                     "description": "Use an existing custom sysimage when starting the REPL",
                     "scope": "application"
+                },
+                "julia.forceDebug": {
```
davidanthoff: Would probably make sense to use a better name here that makes it clear that this is about the language server, right? This name here suggests to users that we have debugging support for julia in the extension, which we (unfortunately) don't.

src/extension.ts
```diff
@@ -75,7 +75,7 @@ export async function activate(context: vscode.ExtensionContext) {
     // Start language server
     startLanguageServer();
 
-    if (vscode.workspace.getConfiguration('julia').get<boolean>('enableTelemetry')===null) {
+    if (vscode.workspace.getConfiguration('julia').get<boolean>('enableTelemetry') === null) {
```
davidanthoff: There are a lot of these formatting changes in this PR, any chance you could get rid of those? Makes it easier to check what really is changed. If not, no worries.

runapp: Reverted. I just subconsciously pressed Format Document lol

package.json
runapp: My English is not good so I have no suggestions on tip writing, thought I totally agree with you. A naive idea is to rename it to `julia.extension.debugMode`?

davidanthoff:
@runapp I think what is still not clear to me is what this option actually does, i.e. I'm not familiar with the debug mode of the LanguageClient. What does that actually do?

runapp:
Sorry I really didn't get your exact point in the last comment.
Programatically it just passes an extra boolean to the constructure of `LanguageClient`.
As a result, the Julia part(`main.jl` in `scripts` folder) runs with `--debug=yes`, and communications between the JS part and Julia part (jsonrpc contents) is printed to "Output" pane in VSCode.
I think those text is helpful to me,cause it helps me finding that it's not `a.jl`, the only file opened, but another `.jl` file in the same folder causing the problem in issue #876 .

PR #902, opened 2019-12-14
Show eval return value in REPL

### LINKED ISSUE #833: Alt+Enter does not show result of expression
I've recently upgraded to 0.12.1, and found that Alt+Enter doesn't show the code block runned anymore, which seems to be good for using it on functions. However the result of the expression disappeared too. I wonder whether it's by design or a mistake.

If it's by design, I suggest that it should check the last token, and hide output only if it is `;`.

davidanthoff:
Fixes #833.

This does not yet work because the `julia/getCurrentBlockOffsetRange` message to the LS seems to return a block that does not include any trailing `;`, and that breaks the logic in this PR here that would decide whether to display a result or not. @ZacLN, could you look into the LS implementation of `julia/getCurrentBlockOffsetRange` and maybe change that so that it includes trailing `;`?

### DIFF
--- a/scripts/terminalserver/terminalserver.jl
+++ b/scripts/terminalserver/terminalserver.jl
@@ -116,6 +116,10 @@ function sendMsgToVscode(cmd, payload)
     print(conn, payload)
 end
 
+function ends_with_semicolon(x)
+    return REPL.ends_with_semicolon(split(x,'\n',keepempty = false)[end])
+end
+
 @async begin
     server = listen(pipename_torepl)
     global conn = connect(pipename_fromrepl)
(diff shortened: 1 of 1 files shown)


ZacLN:
I think it should be enough to change `p2` to `p3` in the second return argument at https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/requests/misc.jl#L74. I can't think off the top of my head whether that would interfere with logic elsewhere to probably worth testing a bit.

davidanthoff:
Lets do that in v0.13.1 :)

davidanthoff:
> I think it should be enough to change p2 to p3 in the second return argument at https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/requests/misc.jl#L74. I can't think off the top of my head whether that would interfere with logic elsewhere to probably worth testing a bit.

@ZacLN do you want to make that PR? Is probably safer than me dabbling in there :)

Issue #863, opened 2019-09-21
Move cursor to next cell when executing a code cell


davidanthoff:
Also, on my system execute cell does not move cursor to next cell. That would be great to have as well as a shortcut to move to cell up/down with cursor.

_Originally posted by @phlavenk in https://github.com/julia-vscode/julia-vscode/issues/858#issuecomment-533802024_

dgleich:
Would you be receptive to making this optional? I actually don't like to to advance to the next cell when running...

Using there are two commands: run and step, and just run. (Atom/Juno uses Alt-Enter for run, Alt-Shift-Enter for run and step to next...)

Seems like an easy change... :)

davidanthoff:
Yes, we should add different commands that execute and execute&move.

Finding key bindings for them will be much more difficult because they are scarce, but we can worry about that once we have the basic functionality.

dgleich:
Seems like this is here now, but there is no default keybinding... I guess I'm unusual in that I like running the same cell over and over...

- Alt+Enter
- Shift+Enter

currently do the same thing.

Is the goal to hold off on finding keycaps for these until some of the notebook stuff is done? I'm teaching a class soon and was checking to see if I could switch over my intro to vscode vs. Atom... having 50 students setup a new keybinding as the first step isn't always the most fun thing to do...

PR #908, opened 2019-12-17
Update repl.ts

ZacLN:
Address #863


davidanthoff:
I'm putting this on the v0.13.1 milestone, i.e. we can merge as soon as we ship v0.13.0.

ZacLN:
vscode seems to handle it well even if we're on the last line

PR #929, opened 2020-01-11
Adding NUM_THREADS to test environment

tro3:
I am unable to run the test suite in my environment, so I may have a TS syntax error here.  (I'll see if I can figure out how to fire it up.)  But the concept of adding the `env` option works - I tested by hacking my local js files in the extension.

Let me know if you'd rather have this packaged in an alternate branch, or what other contribution guidlines you'd like me to follow.


tro3:
Oops - added to the wrong Task...

PR #931, opened 2020-01-13
Add another doc format option

davidanthoff:
Required for https://github.com/julia-vscode/DocumentFormat.jl/pull/63.


package.json
```diff
@@ -454,6 +454,12 @@
                     "default": true,
                     "description": "Format inline documentation.",
                     "scope": "window"
+                },
+                "julia.format.kw": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Format kw.",
```
davidanthoff: @ZacLN What would be a good description? I'm not actually sure what this option does :)

PR #940, opened 2020-01-19
Handle telemetry notifications from the LS

davidanthoff:
This will only properly work once I have rewritten the LS message sending to be async, but we can merge this here now already, it works fine with the current version of LS as well.

My current plan is to only change the message sending of LS to a channel/async model for this release. That should be much easier, and is enough to enable sending of crash reports from the symbol server when the LS itself didn't crash. It will also allow us to use the progress reporting part of LSP.

This also enables us to send generic telemetry stuff from the LS, which I assume could be useful.


davidanthoff:
@ZacLN this should be ready for review and merge now.

davidanthoff:
I'm moving this to 0.14.1, no need to merge now.

PR #943, opened 2020-01-22
Update extension.ts

ZacLN:
Read server debug mode from settings


davidanthoff:
Maybe we should revive https://github.com/julia-vscode/julia-vscode/pull/522 instead?

ZacLN:
Ah, yes

Issue #820, opened 2019-08-09
Use new fs API


davidanthoff:
See [here](https://code.visualstudio.com/updates/v1_37#_vscodeworkspacefs).

DonJayamanne:
Based on my experience, this is required when dealing with user input or workspace files. As for opening files that are shipped with the extension we can continue using the node FS api.
E.g. if a user were to open a remote julia file, in that instance its best for extension to use the VSC FS Api to read the contents of the julia file (this is just an example, ideally one would use existing text document API).

I.e. IMHO I dont think we should udpate all of the code to use the VSC API (not assuming this is what the original intent was, but would like to ensure anyone picking this up is aware of that) - again this is my opinion.

pfitzseb:
Yeah I think we're doing fairly well in this regard, although there (unfortunately) are some cases where we need to read files outside of the workspace.

davidanthoff:
Great, that makes sense!

The space where I'm actually really confused is how our LS is supposed to handle that. There is lots of code in the LS that reads files from disc, and none of that works when a remote file is opened. I've never understood what the idea for that is in terms of the LSP :) Or to say it another way, I've always thought that the LSP really needed a part where a LS can open files via LSP, rather than using direct file access. But, that is probably for another day.

DonJayamanne:
> s files from disc, and none of that works when a remote file is opened. I've never understood what the idea for that is in

Generally with LSP, they get the contents of the current file. But not necessarily of other remote files.
I.e. if user is editing a remote file, then LSP gets the contents of that file.

PR #944, opened 2020-01-23
Add execute file to explorer context menu

ChrHorn:
Individual files can be selected with the explorer context menu and sent to the REPL.

### DIFF
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -614,15 +614,24 @@ async function executeInRepl(code: string, filename: string, start: vscode.Posit
     sendMessage('repl/runcode', msg)
 }
 
-async function executeFile() {
+async function executeFile(uri?: vscode.Uri) {
     telemetry.traceEvent('command-executejuliafileinrepl');
 
-    var editor = vscode.window.activeTextEditor;
-    if (!editor) {
-        return;
+    let path = ""
+    let code = ""
+    if (uri) {
+        path = uri.fsPath
+        code = await fs.readTextFile(path)
     }
-    let document = editor.document;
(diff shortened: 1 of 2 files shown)


ChrHorn:
Just saw https://github.com/julia-vscode/julia-vscode/issues/820 and switched to the new fs api.

src/repl.ts
```diff
@@ -759,7 +769,7 @@ export function activate(context: vscode.ExtensionContext, settings: settings.IS
 
     context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaCodeInREPL', executeSelection));
 
-    context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaFileInREPL', executeFile));
+    context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaFileInREPL', (uri?: vscode.Uri) => { executeFile(uri) }));
```
davidanthoff: Do we need that change here? I would have thought that we don't need to introduce the anonymous function here, but can just continue to pass `executeFile`?

ChrHorn: I think that's the only way to get the path to the file that was selected in the explorer. Got the answer from here: https://stackoverflow.com/questions/51961457/how-to-get-file-name-or-path-in-vscode-extension-when-user-right-click-on-file-i/

davidanthoff: It just seems to me that the expression `executeFile` and `(uri?: vscode.Uri) => { executeFile(uri) }` in both cases evaluates into a function that has the same signature? Could you try the other option? If it doesn't work, happy to go with what is here.

ChrHorn: You are right. I checked again and everything seems to work fine without the anonymous function.

PR #946, opened 2020-01-26
Add debug adapter

davidanthoff:
Nothing works, this is just my experimentation branch.


ZacLN:
Does it run `foo.jl` and connect to it? I was trying to set this up and really struggling. Supposedly you can point to a debug adapter just by specifying the script in the `package.json` but I got nowhere with that

davidanthoff:
No, it doesn't :) I opened an issue over at the main vscode repo asking for help.

It is weird, because I had that part running in an old branch somewhere else years ago, and I can't see what I'm doing different now...

ZacLN:
I had exactly the same experience. Anyway - I'm going to get DebugAdapter's interaction with JuliaInterpreter up and running so things are ready to test

davidanthoff:
It is actually not exactly clear to me right now whether we should implement the debug protocol in Julia or in TS... In particular, if we want F5 to run the current open file under the debugger (which I think should be the goal), then I'm not sure it is the right model to have the debug adapter and the debugee in the same process... So I think we should probably first try to get a good handle on the various start options before we spend a lot of time implementing the details of the protocol?

ZacLN:
> ... not sure it is the right model to have the debug adapter and the debugee in the same process... 
Why would this be necessary if it were implemented in Julia? Naively  it seems like it

davidanthoff:
>> ... not sure it is the right model to have the debug adapter and the debugee in the same process...

> Why would this be necessary if it were implemented in Julia?

I don't understand right now how the debugee is supposed to have a console if it also hosts the debug adapter. That wouldn't be a problem for debugging stuff in the REPL (because in that case we would attach the debugger via a port to the running process), but for launching a new debugee, I think we have to communicate via stdin/stdout, and that seems to mean that we couldn't at the same time have a console? I'm not sure, I might misunderstand that part right now. I also need to go back to my old implementation and see what I had done there.

> Naively it seems like it would ultimately be easier to control the debugee from another julia process than from TS?

Maybe :) On the flipside, we could use a fully implemented protocol implementation from TS that seems to exist. Not clear right now whether the debug protocol is actually based on JSON RPC? I think maybe not? So we would have to implement that as well.

I think right now I still need to play around more with stuff to get a sense what the best strategy here is.

weinand:
Basically you have two ways to architecture your Julia debugger extension:
- Implement the DAP in a debug adapter that runs as a separate process implemented in any language (e.g. Julia?). You can talk to this process via stdin/stdout or via a socket. Additional functionality (e.g. "f5" support without launch configs) can be implemented in the debugger extension in TS.
- Implement the DAP in JS or TS. With this approach you have the option to run the DA as a separate process or "inline" inside the extension.

Mock debug illustrates these variants here: https://github.com/microsoft/vscode-mock-debug/blob/28b4fde41228b0aabe93eab9064992e276859405/src/extension.ts#L91-L149

davidanthoff:
@weinand, cool that is very helpful! I think I am getting the picture slowly :)

And, progress: debug adapter now starts, so that step is taken.

I think the big question for us is whether we should have the debug adapter run in the debugee process, or whether we have one process for the debug adapter that then controls the debugee. For various reasons I think the latter is probably better, but I'll experiment around a bit more.

weinand:
BTW, another option is to run the debug adapter in the language server. So a single process implements both protocols LSP and DAP. 
The Java extension does this.

davidanthoff:
Just a quick update: I've played with yet another option, namely to implement the DA inline in the extension process in TS. That seems to only work in the insider build so far, but I assume this will ship soonish?

I think that is _by far_ the simplest way to implement this. Here are some major benefits:
- Startup time: if we implement the DA in Julia, every debug start will have to start two Julia processes (one for the DA, and one for the debuggee), with all the delays due to the Julia startup performance issues we have. With the inline implementation, we only have to start the Julia debugee process itself.
- Debugging: Way, way easier! The DA is written in TS, runs in the extension process, and so we can just use the excellent TS debugger on every single part of it.
- Amount of code: we can use all the pre-canned packages that exist in TS to implement a DA, and we can just start with the template from the mock debug adapter. In the end we'll have a solution where we have to write significantly less code than any other option.

The only downside I can think of right now is that our DA will only work with VS Code, and not other IDEs. I think for now I don't really care, I think if this becomes an issue later on, someone can still rewrite the DA in Julia.

I'll leave this PR open, but right now I'm working on a different branch that pushes this strategy that I just described.

davidanthoff:
@weinand quick question: I've prototyped a version now that uses the inline debug adapter option, i.e. where the debug adapter runs inside the extension. I'm very happy with that solution, it seems way easier than anything else. But, that API only exists in the Insider edition. Do you have a sense when that will ship in the full product?

ZacLN:
This all sounds great David, and def the right  approach.

weinand:
Every Insiders becomes Stable at the beginning of the next month: that's next week...

davidanthoff:
Awesome :) I can wait a week, and I certainly won't be done in a week.

davidanthoff:
Closing this PR, our debugger will come via https://github.com/julia-vscode/julia-vscode/pull/957 (which has a broadly functioning implementation at this point).

PR #952, opened 2020-01-27
Update the list of Julia versions we auto-detect

davidanthoff:


ZacLN:
We're not ready for 1.4, the `begin` in array refs will cause too many problems. I'll have a look tonight

davidanthoff:
Ok, removed 1.4 support for now.

PR #957, opened 2020-01-29
Add debug adapter

### LINKED ISSUE #125: Debugging support
Currently we are waiting for some more work upstream. See [this](https://discourse.julialang.org/t/debugger-wire-protocol/488/7) discussion.

### LINKED ISSUE #364: Don't bind something to F5
By binding to F5 we seem to disable the more general infrastructure for F5 in VS Code. There is probably some other way to hook into the run file story that is less disruptive. I think it had something to do with debug adapters, though...

davidanthoff:
This is broadly functional at this point.

Major open TODOs:
- [x] Deal with situations where we don't have source code
- [x] Figure global breakpoints out
- [x] Remove need to step twice over single location
- [x] Detect kind of breakpoint stop and return proper terminated event
- [x] Figure out whether we can either run the wrapper in node or get rid of it entirely
- [x] Differentiate between `@run` and `@enter` macro
- [x] Figure out why Julia segfault on Linux at end of session
- [x] Support cmd line arguments
- [x] eval support (watch etc.)
- [x] Compiled mode
- [x] Remove mock runtime
- [x] Clean up Julia code
- [x] Clean up TS code
- [x] Use global BP handling from JI https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/370
- [x] Show nice UI for exception breakpoints
- [x] Handle composite variables better
- [x] Add restart frame support
- [x] Add set variable support
- [x] Test macros and how they behave

Things for a future version:
- [ ] Add global variable support
- [ ] Investigate a debug REPL mode
- [ ] Support `env` setting
- [ ] Deal with sub env, in particular when debugging things like `test/runtests.jl`
- [ ] Alternative to integrated terminal startup
- [ ] Handle completion requests
- [ ] Handle breakpoint location request
- [ ] data breakpoints
- [ ] Handle cancel requests
- [ ] Send column in stack trace request
- [ ] Deal with kwargs wrapper stack, ideal hide it somehow

Fixes #364. Fixes #125.


ianfiske:
Just a note to request you to consider looking into Infiltrator.jl and/or MixedModeDebugger.jl for this adapter (https://discourse.julialang.org/t/improve-debugger-gui-to-help-adoption/33708/21).  They are the only debugging tools so far that are fast enough to actually be useable from my experience.

davidanthoff:
@weinand, can I ask for your help again? I'm having difficulty getting the nice exception thrown UI up and running. I've implemented
```
protected async exceptionInfoRequest(response: DebugProtocol.ExceptionInfoResponse, args: DebugProtocol.ExceptionInfoArguments) {

	response.body = {
		exceptionId: 'SOMEID',
		description: 'THE DESCRIPTION IN THE BODY',
		breakMode: 'userUnhandled',
		details: {
			message: 'THIS IS THE MESSAGE in the DETAILS',
			typeName: 'THETYPNAMEINTHDETAILS'
			// fullTypeName: 'THE FULL TYPENAME'
			// stackTrace: 'THE STACK TRACE'
		}
	}

	this.sendResponse(response);
}
```
But none of that info seems to appear anywhere in the UI. It is being called, I've checked that.

Any idea why this might not work? I would like a UI that looks like this show up:

image

davidanthoff:
@weinand: I actually figured it out, I had the same problem as https://github.com/microsoft/vscode/issues/28964. That is pretty tricky to figure out, though ;)

davidanthoff:
@ZacLN I think this is ready for review/merge. It really shouldn't impact any of the existing functionality, other than the change of what F5 does. So we might as well start shipping it?

There are a fair number of things I want to redesign as we go forward (in particular on the message side of things between the extension and the DA), but none of that is urgent and I think what we have can be shipped.

davidanthoff:
Cool. I'll hold off a couple of days until we've gotten rid of a couple more bugs that come in via crash reporting and then I'll merge.

PR #980, opened 2020-02-11
Add notebook controller implementation

davidanthoff:
This PR has _only_ a kernel implementation that can be used for Jupyter notebooks. The Jupyter extension provides all the file IO functionality, and we are only in charge of executing the code and returning results.

List of things that we still need to resolve on our end before we can ship a first version:
- [x] Support different Julia versions as kernels
- [x] Return MIME bundles rather than just one representation for results, like IJulia does
- [x] Add kernel lifetime management UI
- [x] Remove the use of one terminal per kernel, instead us child processes directly
- [x] Fix close/dispose/shutdown situation, right now a lot of stuff will just hang around

~~- [ ] What happens if a notebook is closed, kernel is still running, and the same file is opened again? The kernel should probably be reused?~~
~~- [ ] Check that all the MIME types we support in IJulia also work~~
~~- [ ] Figure out what we do about variable viewers. Could we hook the button in the toolbar to view the Julia workspace?~~
~~- [ ] Is the Julia extension going to be activated every time someone opens a Jupyter notebook, even if no Julia is used in that notebook? We might have to make sure that our extension does almost nothing in that scenario~~
~~- [ ] Generally go through the entire IJulia codebase and check whether there are features we need to have here as well~~
~~- [ ] Think about kernel status UI. JupyterLab has nice info like "Kernel idle", "Kernel starting" etc. Not sure where we could have that.~~
~~- [ ] Handle Julia environments~~

List of things that need to be resolved on the VS Code/Jupyter extension end before we can ship:
~~- [ ] I think general kernel management, although I might be wrong. Not clear to me right now whether the different kernels we register will now be automatically selected properly and whether kernelspec data is properly written to Jupyter notebook files.~~
- [x] Are things stable enough in VS Code and the Jupyter extension that we could ship this PR? My sense is yes, but would be good enough to double check.

UPDATE: Moved a lot of small issues into separate proper issues, we can resolve these after this PR has been merged.


aviks:
This is very exciting.  I was wondering about this `.jlnb` format -- would it make sense to do this with Weave `.jmd` files, and leverage weave for the actual rendering? (Pardon if this question is out of bounds)

davidanthoff:
> I was wondering about this .jlnb format -- would it make sense to do this with Weave .jmd files

This PR actually started out with `.jmd` files :) But then I changed it half way through. I'm still not super sure what the right call is.

My current thinking for a new file extension is roughly this:
- we have to register our notebook provider for a file extension. I kind of feel that if we register for a file type, we should support every feature of that file type 100%, i.e. it would be really a sub-par experience if we register for `.jmd` files, but then we end up bailing on some features that users have been using in those files. And we won't be able to support inline code, nor potentially some of the cell metadata. Also not sure about the document level metadata.
- I really want a distinction between a code cell that can be run, and a code block inside markdown that is just displayed in the markdown but that doesn't start a new cell. I don't think that distinction really exists in `.jmd` files, right? In my `.jlnb` format code cells are started by ` ```{julia}`, but ` ```julia` does _not_ start a code cell, it is just treated as content of the current markdown cell. I copied that from R Notebooks, and I kind of really like it.
- if we start with a fresh file extension, we can organically add features to the file format as we add features to the extension, without worrying too much about other stuff.

> leverage weave for the actual rendering

So I think for the core notebook editing and execution experience within VS Code we don't need Weave.jl. I think the Julia code that is in `notebook.jl` in this PR is pretty much all we need on the Julia side of things. Plus, Weave.jl has a lot of dependencies, and we generally try to make sure that the core features of the extension work out-of-the-box without the need to install any Julia packages, and with the current implementation that is the case.

BUT, if we do stick with the `.jlnb` extension, it would be awesome if Weave.jl could process them. In my mind that should be fairly trivial, at the end of the day they use strictly less features than `.jmd` files, so I would expect that Weave.jl could be the Julia package that one would use to programatically, from Julia, run and convert `.jlnb` files?

But this is all still very experimental, so there is certainly no firm decision to go with our own file format, happy to discuss all of this!

aviks:
I'm genuinely torn on this. On hand and, having a lightweight format for working in a notebook environment makes sense. For interactive explorations and teaching/sharing, it's perfect. 

However, I'm wary of throwing another format out into the world. In addition, I want to have a great authoring experience for jmd files, which I think this work can allow. 

cc:: @aviatesk and @pfitzseb who I know have thought about this.

cloudhan:
Let me link these issues
Notebook UX Explorations: https://github.com/microsoft/vscode/issues/95154
Notebook API: https://github.com/microsoft/vscode/issues/93265

jlperla:
When you use these vscode technologies for notebooks, do they require a python installation (as the "interactive terminal" and existing notebook does) or is it 100% typescript/javascript?

davidanthoff:
No Python involved at all, this is a pure TypeScript and Julia based implementation.

jlperla:
In addition. Have you played around with the interactive terminal for vscode? It is amazing and might change some of the design you could consider for the non-notebook interface.

davidanthoff:
You mean just the normal terminal in VS Code? Yes, the extension uses that for the Julia REPL, and this notebook PR here runs each notebook kernel inside such a terminal. Right now I'm showing stderr in the terminal, and stdout in the notebook, but I'm not sure whether that is actually a good design or not...

jlperla:
No, it is something different.  Looks like they may call it the "Python Interactive Window", not the "interactive terminal".  I think it is something you guys should examine carefully for the workflows, and not simply with the "notebook provider".   With python it is amazing and I find it far better than working with jupyter notebooks directly.  Try it out.


https://code.visualstudio.com/docs/python/jupyter-support-py#_python-interactive-window  and you can see that this is how they have their notebook debugging strategy: https://code.visualstudio.com/docs/python/jupyter-support#_debug-a-jupyter-notebook

You you can look at the rough notes I put down: https://github.com/ubcecon/tutorials/blob/master/python.md#interactive-work-with-python-files but I think all that is requierd to install is the https://code.visualstudio.com/docs/python/jupyter-support.

I know you are loath to add in python dependencies, but it is unavoidable if you start doing notebook stuff.  But if you used the `IJulia` conda installation, it could all be seamless.  Use an existign conda if it is installed, or do the `IJulia` installation in the setup process if not.  The user wouldn't even have to know.

*Speculation:*
As far as I can tell, it is just a jupyter kernel they run, so that code is sent to that terminal.  I see no reason that would be any different for Julia.  Moreover, seems very likely they will start integrating the backend kernel to run remotely in a seamless way, and there is no reason that the kernel can't be used to put results inline.  Then whatever mime-types you want to support in the IDE can just be implemented for the IJulia and you get it for free in vscode.

I think this is the future of jupyter.  The protocol is used to communicate piecemeal with a kernel in a highly interactive way, where what the kernel returns may be shown inline, in a different window, etc. depending on what the caller wants.

ToucheSir:
@jlperla have you tried using the Julia REPL integration? Much of the functionality from the Python interactive window is unnecessary because the Julia REPL is far more far-featured than the default Python one. `IJulia` merely adds a jupyter kernel layer on top of this functionality.

Some examples of functionality that require the Python Interactive Window but can be done with just the "default" repl in Julia VS Code:

- Inline results
- Rich data display (using `vscodedisplay`)
- Variables view
- Debugging: much like the Python extension, you can already write a script with special comments (`##` instead of `#%%`) to delimit code cells. End to end running and debugging thus works as usual. There isn't any per-cell debugging, but I've found that feature in the Python extension to be _so_ flaky and have so many caveats that it's not much of a win.

Also, it's important to note that the [Python extension](https://github.com/microsoft/vscode-python/issues/10496) is the number one consumer of and motivation for this new notebook provider! Both the Python Interactive Window and Jupyter Notebook editor are in this awkward position where they don't _quite_ work the same as `IPython`/Jupyter, yet also don't integrate well with any other part of VS Code (e.g. syntax highlighting). Given the amount of thought and effort that MS has put into the notebook provider APIs, I suspect the focus will be on developing this functionality instead of adding, say, inline results to the default Python editor.

jlperla:
> Much of the functionality from the Python interactive window is unnecessary because the Julia REPL is far more far-featured than the default Python one. IJulia merely adds a jupyter kernel layer

I have used all of these extensively. I don't think think the strategy of the interactive terminal to you is clear, as it isn't entirely to me. But for interactive stuff it is a far better experience than the current Julia repl and extensions in vscode for many things.

That said, I am not saying you should rip everything out and build around the interactive window, but just consider how it fits into the vscode notebook strategy. I believe it may be central to it. 

Mostly this would influence how you think about "notebook integration' for the vscode extension. I would understand the strategy there fully before designing your own notebook backend. Using a backend jupyter kernel infrastructure already there might be helpful

davidanthoff:
Ah, got it!

So the Python Interactive Window is a custom UI webview that they implement in the Python extension. My understanding is that they plan to replace that with a UI that is based on the same notebook API that I'm using in this PR here.

The question of whether one uses Jupyter for code execution or not is pretty orthogonal: there is the question of how one implements the front-end (VS Code notebook API or custom webview) and how one implements the code execution part (Jupyter protocol or not).

jlperla:
> So the Python Interactive Window is a custom UI webview that they implement in the Python extension. My understanding is that they plan to replace that with a UI that is based on the same notebook API that I'm using in this PR here.

Are you sure they would plan to replace it?  It seems a completely separate approach for those that really don't want to work in a notebook.  i.e. in effect it is a jupyter terminal that you can send python to instead of the existing terminal, which they would use for non-ipynb interactive code.

It also seems like it would be much easier for them to open up that functionality than building full ipynb editing support for mulitple languages.  Regardless, I would consider asking someone if they are thinking of opening that up in a way similar to the way they opened up the notebook UI.

More generally, consider that there are a lot of people who really like textual formats and do not like working in a notebook of any support.  The thing I like about `Rmd` and weave is specifically that it is not a notebook but can export them or create literate documents.

> Given the Python extension team hasn't seriously considered breaking out their Jupyter functionality into a language-agnostic extension, I think this may be an uphill battle. Agreed that IJulia would work well for the notebook backend if this were resolved.

I have no idea of what is happening on their side, just that it is worth asking someone what their medium-term intentions are.  It is definitely not something you would want to write on your own, but if they have plans to open up that particular feature it could be very useful.  

> The question of whether one uses Jupyter for code execution or not is pretty orthogonal: there is the question of how one implements the front-end (VS Code notebook API or custom webview) and how one implements the code execution part (Jupyter protocol or not).

I defer to you on that.  In my mind if there is a medium-term plan to have a more open jupyter terminal they are managing (not one you would write) it could change things.

pfitzseb:
It's important to note that this isn't going towards a re-implementation of Jupyter notebooks or whatever.
The goal is to display code and text cells (in plaintext documents -- what David uses here is a subset of Weave's jmd, I think) as well as results. There's also the added benefit of "serialization" and "deserialiazation" of results into a HTML file, but that's a seperate thing.

davidanthoff:
> Are you sure they would plan to replace it?

https://github.com/microsoft/vscode-python/issues/11898

jlperla:
> microsoft/vscode-python#11898

Thanks for confirming.  I am curious to see how they implement it for just `.py` files, which might be a good indicator of what is possible to do with just `.jl` files for people who don't want to move to a new file format but just want inline results.

Do you suspect they will have reusable jupyter infrastructure for other people, or is the python team just not in that mindset?

> It's important to note that this isn't going towards a re-implementation of Jupyter notebooks or whatever.

Good to hear.  File extension like `.jlnb` might sound scarier than it really is then.   

> The goal is to display code and text cells (in plaintext documents -- what David uses here is a subset of Weave's jmd, I think) as well as results

Then why not use `.jmd` and patch up Weave?  And lots of people already use it?  It is an enormous amount of work to implement the superset of features of Weave, Literate, etc.  and test all of the transcoding/etc.

davidanthoff:
> Then why not use .jmd and patch up Weave?

See my thoughts on that above. The problem is not Weave, the problem is that Weave has features that the notebook UI in VS Code doesn't support. So if we were to use jmd as a file extension, there would be jmd files out there that will not work in the VS Code UI, which is something I _really_ want to avoid. I think if we support a file format in the extension, we need to support it 100%.

jlperla:
>  I really want to avoid. I think if we support a file format in the extension, we need to support it 100%.

I just can't help that the python guys for vscode have figured something out here, unless they are completely dropping support for the interactive use of `.py` files.  That is why I brought up the interactive terminal.  It is for `.py` files, not `.ipynb` ones.

But if you need 100% support, I wonder whether if the current design is feasible.  Weave is pretty barebones already...  Why can't you just have text that doesn't enable GUI elements where appropriate?  And if you can't support adding features because it is an all-or-nothing thing, then is a new notebook format worth it or should you wait until the GUI is more capable?  I am not sure that otherwise fragmenting into a new fileformat with first class editor support - but otherwise very limited functionality - is a good idea for the ecosystem.

Anyways, no need for me to say anything more here, as you know the vscode stuff better than I do.  But consider the reason Rmd is so successful is because it is flexible and not minimalist.  The fact that it feels like an interactive text file rather than jupyter-style tightly controlled notebook environment with 100% gui coverage is a good thing to me, not a limitation.

davidanthoff:
I think there are really at least three entirely distinct conversations going on here.

UI: we realistically only have two options in VS Code. 1) notebook (feels a bit like JupyterLab, but actually a bit smoother), 2) text with limited inline results (as we are going to ship soonish). If we wanted to do something like the R Notebook UI experience, which feels like a text editor with output in-between lines, we would have to program that UI ourselves from scratch, which is a no-starter for us in terms of dev time. I believe the latter is what the Python extension did for their interactive Python window, and the fact that they seem to give up on that suggests that they also didn't feel comfortable maintaining that kind of custom UI. But in any case, we (the Julia side of things) _certainly_ don't have the person power to do anything like a custom webview notebook editor. So just from a technical point of view I don't see any scenario where we could get something like an R notebook UI going.

Code execution: doing this directly between the TypeScript extension and Julia is simple and way, way more robust than using some Jupyter middleware for that because we don't take any dependency on complicated setup stuff. Also, the code that does this is more or less done in this PR.

File format: this is clearly not decided at this point and we'll sort that part out later. There are many options on the table, and we'll just have to see how things shake out. But note also that this part is also not involved (at least for now while we don't save and load output cells). Right now that part is maybe 50 lines of code total or so.

jlperla:
100% agree.  Though the 3 issues all interact.   And what you are saying about the two options seems correct based on what is in vscode right now (though I was hoping it would change, maybe it won't).

Iif the move away from the interactive python window is an admission of defeat that `.py` files can ever be used interactively, that isn't a good sign.  Although alternatively it could mean they have figured out a way to have more of an Rmd style inline text for python.  We can ask them?

My position:  If using a notebook UI means that only a tiny subset of the features of Rmd/Weave/etc are available, maybe it is better off waiting to see if the UI makes progress to enable more Rmd style editing.  Otherwise, it could risk wasting timing and splintering the ecosystem's literate file formats by moving six months too early.  Or put another way...

>UI: we realistically only have two options in VS Code. 1) notebook (feels a bit like JupyterLab, but actually a bit smoother), 2) text with limited inline results (as we are going to ship soonish). I

Then is it work all the trouble at this point just for something a bit smoother?  People can use Jupyterlab for a notebook experience for the short term.   Those who want to have a more full experience with literate documents  can use Weave with some minor enhancements like Juno has.  It is a pretty solid piece of software, even if it is incomplete.

> File format: this is clearly not decided at this point and we'll sort that part out later.

100% agree.  The format isn't the key.  Whether it is open and extensible, or a tightly controlled subset of standard features is the key.

jlperla:
@davidanthoff  See all of the great news from the vscode crew https://github.com/microsoft/vscode-python/pull/12096  for example.

Whether this changes anything for a notebook interface or not is a separate issue, but it means that (1) using the jupyter interactive terminal is feasible for other languages; and (2) using jupyter as a backend protocol for executing whatever notebook features exist may make sense.

aviatesk:
@davidanthoff I've not dive into the details, but can't we register this notebook provider for `.ipynb` format ? I agree with not providing this notebook interface on `.jl` or `.jmd` formats, since they're supposed to have really different editing experience, while I don't think publishing the new file extension just for this feature won't be so good idea.
If we go with `.ipynb`, we can leave "converting into notebook task" to the existing package (like [nteract's one](https://github.com/nteract/nteract/tree/master/packages/commutable)), can have natural supports for (de-)serializing results, and we don't have any existing features that we "have to support" for this file format.
Well, currently vscode-python extension automatically tries to open `.ipynb` files in the notebook interface assuming it's a IPython notebook regardless of whatever language the `.ipynb` has, but I _guess_ if they ship that as public extension API, this should be done more smartly like looks up the kernel language and choose provider, or something like that.

Well, I would really like to work on this as my next thing to do on this extension.

davidanthoff:
@jlperla: I'm actually quite unhappy about the direction that the Microsoft Python extension team is taking with respects to Jupyter notebooks. They seem to go for a solution where the Python extension will handle ipynb files (even those that don't use Python, but a different language), and they _will_ support other kernels, but they will use a Python glue code to run code on those non-Python kernels. So if you have a Jupyter Notebook with Julia code in it, the Python extension will say "I can handle that", and it will, but you'll have to install Python to run that Julia Jupyter notebook. In my mind that is a really bad design, given that the Jupyter kernel spec was _explicitly_ designed to _not_ require that. I think what MS _should_ be doing is to re-animate the Jupyter VS Code extension, and then use the same techniques that nteract uses to directly to talk to individual language kernels. That is how the designers of the Jupyter kernel spec imagined clients would be implemented, and would mean that if someone installs VS Code and the Jupyter extension, they would need Julia installed to run a Julia Jupyter notebook, but not Python. But, alas, that is not what they are doing... We should probably move these kinds of broader discussions to a new issue if folks want to discuss that, so that we can keep the discussion in this PR focused on the code associated with it.

@aviatesk I believe the Python extension will continue to just "hijack" all ipynb files. I think if that is the case, then using ipynb as our format is not a good idea: if a user opens a repo with such a file in there, VS Code will recommend the Python extension to open it, and I think that is really not what we want. But I'll follow up with the VS Code folks again to see what the latest thinking on this topic is and whether there might be something that an extension can register only for say those ipynb files that have a specific kernel.

jlperla:
Thanks @davidanthoff   for the perspective.  I hear you on all fronts.

The only thing I will defend is that having an installation of python in order to use jupyter-based technologies seems a reasonable compromise.   In practice, jupyter/etc. are tied to python, for better or worse, and despite the intentions.  If you use IJulia, for example, you will end up using the conda julia installs, or you will install it yourself.  I would be shocked if there is anyone who both wants a GUI based notebook interface for julia, but is unwilling to install python and the standard vscode python extension.

But while having things python-free just doesn't seem worth the trouble to me, your problem with the hijacking of the ipynb files to the "python" extension does sound like a real issue.  But I am not sure that it makes sense to rename what is effectively a jupyter-based format just to avoid it.

> We should probably move these kinds of broader discussions to a new issue if folks want to discuss

The relevance here is only that it raises the question of whether this is the right time to do the notebook extension - and whether jupyterlab is good enough until a more complete text based weave/Rmd style setup can be achieved without splintering into more text-based notebook formats.   That was my main concern.

But I think we are all on largely the same page with respect to the tradeoffs for the existing technologies so I don't think that it is worth discussing those further here.   Anyways, you have heard more than enough from me on this already.

davidanthoff:
> The only thing I will defend is that having an installation of python in order to use jupyter-based technologies seems a reasonable compromise. In practice, jupyter/etc. are tied to python, for better or worse, and despite the intentions. If you use IJulia, for example, you will end up using the conda julia installs, or you will install it yourself. I would be shocked if there is anyone who both wants a GUI based notebook interface for julia, but is unwilling to install python and the standard vscode python extension.

Ok, so I really, emphatically disagree with this :) You can install Julia, IJulia.jl and [nteract](https://nteract.io/) and have a perfectly working Jupyter notebook editing experience with absolutely no Python (or conda) involved at all. The Jupyter team went out of its way to make that possible, and it is just a shame that MS is not following the nteract model of interacting with Jupyter kernels, which is literally the way that the Jupyter documentation says you should do it.

I think all of that would be less of a concern if installing Python and the Jupyter stuff was more robust, but in reality it just isn't. I tried to teach some courses a couple of times where we used Jupyter Notebooks, and I've never managed to get to a situation where a sufficient number of folks was even just able to get the basic notebook software to run. Yes, in theory, if they all used conda, maybe it would work (although my conda installs also degenerate into non-working chaos after a couple of months, mostly), but then you have someone who installed Python from the Windows Store in the class, and someone else who installed it from somewhere else, and it is just a mess. One of my _main_ goals with this PR is to create something that "just works", always and never has any setup issues. I just don't see how we can achieve that if we take any dependency on Python software being around.

> Anyways, you have heard more than enough from me on this already.

Na, never :)

jlperla:
@davidanthoff  I agree with 100% of what you have said in theory, and most of it in practice, but reach different conclusions.

>  You can install Julia, IJulia.jl and nteract and have a perfectly working Jupyter notebook editing experience with absolutely no Python (or conda) involved at all. 

Absolutely true.  You *can* use nteract.  But does anyone do it who is not sophisticated enough to install Python in first place?  I have never seen any newbies coming onto forums asking about nteract, so suspect the intersection of *can* and *do* includes only principled experts.

> Jupyter team went out of its way to make that possible, and it is just a shame that MS is not following the nteract model of interacting with Jupyter kernels, which is literally the way that the Jupyter documentation says you should do it.

100% true.  It sounds like what you really want is someone to write a typescript-only jupyter front-end implementation.  That would be great, but that isn't the world we live (yet) .  My conclusion: it doesn't make sense for a single programming language do that that directly (or indirectly by effectively cloning a subset of jupyter functionality in a proprietary single-language format).  Better to pressure MS to do it themselves, or to wait for them to have a typescript frontend that seamlessly uses python in the background without installation problems.

> I think all of that would be less of a concern if installing Python and the Jupyter stuff was more robust, but in reality it just isn't. I tried to teach some courses a couple of times where we used Jupyter Notebooks, and I've never managed to get to a situation where a sufficient number of folks was even just able to get the basic notebook software to run. Yes, in theory, if they all used conda, 

I agree with this.  You tell them to either install conda on their own, or you tell them to use the julia-installed conda.  The following instructions were made to have it fully dependable: https://julia.quantecon.org/getting_started_julia/getting_started.html#Installing-Jupyter  for example.

But there are two separate things here: one is whether having a python dependency within the julia code (through `Conda.jl`) "just works" and I think the answer is that it is pretty solid.  The second is whether depending on an external extension like the vscode python extension, where you don't have control over the python side, makes sense.  That is harder to justify, but I suspect that it is something the vscode people have put sufficient time into for the medium term.  Which is a good reason to wait for the medium-term.

> One of my main goals with this PR is to create something that "just works", always and never has any setup issues. I just don't see how we can achieve that if we take any dependency on Python software being around.

I understand the goal, and agree with it in principle, but what is the target market of the feature and what is the opportunity cost of jumping too early?

In effect, what you are suggesting is implementing a huge chunk of jupyter functionality in pure-julia just to get a nice notebook interface within vscode itself.  The file format doesn't matter, but the number of features does.  Either it doesn't do much and you tell people it isn't really intended as a jupyter/notebook replacement.... or you go down that path, you are on the hook for maintaining a massive amount of code, dealing with all sorts of the tedious bugs that occur with escaping of various text/latex (just look at the Rmarkdown backend layers, sphinx docutils, or python jupyter issues and quantiy of code) etc.  That is a rathole of maintenance misery.

And in the end, the experience is that you get a smoother version of the same jupyter interface within vscode rather than having someone use jupyterlab or nteract (as you say).  And that is the best possible result of this, and ignoring all of the massive effort required in the mundane work of training material, documentaiton, etc.  And even if successful and well documented, it sows all sorts of confusion of people asking whether they should be using Weave, Literate, Jupyter, or the vscode julia extension.  The answer becomes a list of: if you want feature XXX use YYY.

Is there a huge audience of jupyter users demanding this feature who are unwilling to run jupyterlab?  I think I see the opposite.  What julia people really seem to want is to move away from a notebook interface and more towards a Rmd style interface where things are more inline and the code is not tightly constrained in a format like ipynb.  That seems to be the way things are going (including in the jupyter community itself, as we discussed previously).  I am willing to bet the main vscode notebook interface in the longterm is slowly decoupled to make things work better piecemeal and inline.

But a big part of this is also a consideration of the opportunity cost of jumping too early due to the very limited number of julia developers capable of working on the vscode extension.  Of course, development time is not fungible, and if someone wants to play around with a cool new notebook interface in vscode rather than working on the language server bugs, that is their business.  But they might want to know whether their development time is spent building out a feature in hot demand vs. one some (e.g. maybe just me!) thinks a year too early and with an unclear audience due to its limited number of possible features.   There is no such doubt on the rest of the vscode features proposed, which is why I am saying:  "wait 6-12 months on the notebook interface to see where vscode/jupyter/notebook stuff goes"

But, as I said, maybe it is just me.  It wouldn't be the first time I have jumped to the wrong conclusions after agreeing with someone on the same evidence.

... OK.  My concerns now fully expressed, I really am jumping out of this unless you guys directly ask me something specific.

cloudhan:
> thinks a year too early and with an unclear audience due to its limited number of possible features.

Pick it early, you might get better support from officials. You evolve with the system and will be heard from the official and your specific needs might be addressed early. That is simply a tradeoff.

jlperla:
If you didn't notice, the latest vscode now supports Julia in the jupyter extension. It works pretty well, as far as I can tell, and installation was a joke. 

The only important bug I found was https://github.com/microsoft/vscode-python/issues/12760#issuecomment-654995857

gramster:
@davidanthoff, I am from the Python team at Microsoft. We are certainly aware of the issues you bring up and are having discussions about how best to address them, but our immediate goal for the next couple of months is the migration to the VS Code notebook APIs. I expect we'll have more to say soon, including soliciting input from the broader Jupyter community.

package.json
```diff
@@ -48,7 +49,8 @@
         "workspaceContains:deps/build.jl",
         "workspaceContains:test/runtests.jl",
         "workspaceContains:benchmark/benchmarks.jl",
-        "workspaceContains:docs/make.jl"
+        "workspaceContains:docs/make.jl",
+        "onNotebookEditor:julianotebook"
```
jrieken, line 53: fyi - this event has been renamed to `onNotebook` (no `Editor`)

pablosanjose:
I know GitHub is not for chitchat, but heck, I had to write this: I'm so happy to see this happening! Jupyter and nteract don't cut it, I think. The editor experience is not good. VS Code is the way to go. But we need a way to store the results of computations together with code (unlike Pluto). We need *precisely* this. Please keep at it!

jebej:
What is the status of this PR given this: https://devblogs.microsoft.com/python/python-in-visual-studio-code-november-2020-release/

It appears notebooks kinda work already for me, though I am not on the insider release of VS Code (which might explain the few missing features).

davidanthoff:
@donjayamanne, I got carried away :) I think the basics of what we discussed are actually working on this branch now, I was able to open a Jupyter notebook and run it with our kernel! Amazing.

I think the main questions right now are
- what is the right way to register our kernel? At the moment I do it for file pattern *, and that allows me to select it, but clearly that is no good.
- I need to understand how to add output to cells. Those code parts are commented out right now, and while I get that I need to do something with a workspace edit, I didn't fully understand how to handle that.

DonJayamanne:
> what is the right way to register our kernel? At the moment I do it for file pattern *, and that allows me to select it, but clearly that is no good.
> I need to understand how to add output to cells. Those code parts are commented out right now, and while I get that I need to do something with a workspace edit, I didn't fully understand how to handle that.

Will look into these.

davidanthoff:
I've managed to add stdout output already, I'll push that in a moment!

package.json
```diff
@@ -803,6 +805,17 @@
...
+            {
+                "viewType": "julianotebook",
+                "displayName": "Julia Notebook",
+                "selector": [
+                    {
+                        "filenamePattern": "*.jlnb"
+                    }
+                ]
+            }
+        ],
```
davidanthoff, line 818: @DonJayamanne Is there something like a `notebookKernelProvider` here in `package.json`? I'd like to remove the whole `jlnb` file format stuff here, and I think also the `julianotebook` stuff, but then I'm wondering how we would get activated?

src/notebook/notebookKernel.ts
```diff
@@ -0,0 +1,197 @@
...
+
+                this._msgConnection.onNotification(notifyTypeDisplay, ({ mimetype, current_request_id, data }) => {
+                    // TODO This doesn't work properly, nothing ever gets shown
+                    const executionRequest = this.executionRequests.get(current_request_id)
+
+                    if (executionRequest) {
+                        const cell = executionRequest.cell
+
+                        const edit = new vscode.WorkspaceEdit()
+                        edit.appendNotebookCellOutput(cell.notebook.uri, cell.index, [new vscode.NotebookCellOutput([new vscode.NotebookCellOutputItem(mimetype, data)])])
```
davidanthoff, line 124: @DonJayamanne I'm doing something wrong here, nothing gets ever shown. If I debug this, this code is called, so I must be making some mistake in how I use the cell editing API.

DonJayamanne: Odd, pretty me check. I just realized there were submodules. And that's why I could not get anything to work.

package.json
DonJayamanne: Basically there's an event that there's fired when a notebook of a specific type opens, I'll send that info tomorrow, it's an entry in package.json (static activation).

scripts/packages/VSCodeNotebookServer/src/stdio.jl
```diff
@@ -0,0 +1,277 @@
...
+                if buf.size >= max_bytes
+                    # send immediately
+                    send_stream(name)
+                end
+            end
+        end
+    catch e
+        # the IPython manager may send us a SIGINT if the user
+        # chooses to interrupt the kernel; don't crash on this
+        if isa(e, InterruptException)
```
DonJayamanne, line 108: Based on my knowledge of interrupts on windows, SIGINT doesn't work on windows.
I believe messages are sent to interrupt kernels.
Please could you confirm, and can we add a message in the Julia kernel for this if thats the case. 
FYI - We can do that later, and I could just create a separate issue to track such requirements

src/notebook/notebookKernel.ts
```diff
@@ -0,0 +1,197 @@
...
+                    asdf[0].dispose()
+                    this._terminal = undefined
+                }
+            }, this, asdf)
+
+            await connectedPromise.wait()
+        }
+    }
+
+    public async restart() {
```
DonJayamanne, line 194: Related VS Code issue https://github.com/microsoft/vscode/issues/118783

src/notebook/notebookKernel.ts
```diff
@@ -0,0 +1,206 @@
...
+import { getAbsEnvPath } from '../jlpkgenv'
+import { getJuliaExePath } from '../juliaexepath'
+import { generatePipeName } from '../utils'
+
+interface ExecutionRequest {
+    cell: vscode.NotebookCell
+    executionOrder: number
+}
+
+const notifyTypeDisplay = new NotificationType<{ mimetype: string, current_request_id: number, data: string }>('display')
```
DonJayamanne, line 16: Are the messages supported here a subset of all messages supported by the Julia Jupyter kernel.
I'm trying to figure out if there are some missing messages or attributes.
E.g. when we get display data, do we have metadata for the images? (in Python, they support additional metadata for image mimetypes).

Similarly it seems we're missing a message for interrupt (which i believe is necessary for Widows), i could be mistaken & this might not be required at all.

src/notebook/notebookKernel.ts
DonJayamanne: FYI - The API for execution will be changing significantly, we're expecting a lot of changes. I'm not going to focus too much on how we update the outputs and the like, as most will change.
Here's some of the on going work, https://github.com/microsoft/vscode/pull/116416
This API will hopefully be ready early next week, I'll continue with areas that doesn't relate to execution to avoid too much re-work or throw away.

scripts/packages/VSCodeNotebookServer/src/stdio.jl
davidanthoff: Yes, we need to add another message for that. We have that for the REPL kernel already, so I think that should be relatively straightforward.

src/notebook/notebookKernel.ts
davidanthoff: Yeah, so the interrupt message is missing, but that I think is the major one.

Once we move the Julia code over to reuse the REPL server code, we'll automatically get support for additional messages for example for the variables explorer.

For "rich" display types, right now we have one notification from the kernel to the extension that has a MIME type string, and a payload. I think that should be enough? Or do we need something beyond that?

scripts/packages/VSCodeNotebookServer/src/stdio.jl
DonJayamanne: I'll create an issue to track that requirement, to ensure we don't forget.

src/notebook/notebookKernel.ts
DonJayamanne: >  Or do we need something beyond that?

Not that i'm aware of.

src/notebook/notebookKernel.ts
DonJayamanne: I think the problem here is these edits are all async, and we're not chaining them.
We had similar (more messy) problems in Jupyter extension.
I'll submit a PR for this.

davidanthoff:
@DonJayamanne So I started to look at this again, thanks also so much for the PR that updated everything to the current API. I can right now run stuff, so that is great!

Should we next try to figure out how to handle different Julia versions? Would the idea be that we create on notebook controller per installed Julia version? So say we do
```ts
this.controller = vscode.notebooks.createNotebookController('julia-1.6.1', 'jupyter-notebook', 'Julia 1.6.1 Kernel')
```
and then the same for all currently installed Julia versions? At the end of the day we need that if we want all the installed Julia versions to show up in the UI as a kernel choice, right?

One complication, though, is that the way Julia works today in Jupyter notebooks is that the following metadata gets typically added:
```json
"metadata": {
  "kernelspec": {
   "display_name": "Julia 1.6.1",
   "language": "julia",
   "name": "julia-1.6"
  },
  "language_info": {
   "file_extension": ".jl",
   "mimetype": "application/julia",
   "name": "julia",
   "version": "1.6.1"
  }
 }
```
It is not entirely clear to me right now where and how the logic for kernel selection would live and work.

DonJayamanne:
> Should we next try to figure out how to handle different Julia versions? Would the idea be that we create on notebook controller per installed Julia version? So say we do

Sure thing, I can do that.
Question - will this confuse users. I know some Julia users were using the Eivnronment picker in Julia to select different versions and expected the kernel to respect that.

I think having the kernel picker to select the different versions is great, but if users are used to the environment picker, then we might want to consider hiding that or considering the possible confusion

davidanthoff:
> Question - will this confuse users. I know some Julia users were using the Eivnronment picker in Julia to select different versions and expected the kernel to respect that.

> I think having the kernel picker to select the different versions is great, but if users are used to the environment picker, then we might want to consider hiding that or considering the possible confusion

Do you mean this element in the status bar:

image

? That only picks a Julia environment, but no Julia version is recorded in a Julia environment, only a set of packages that are available. Julia version and Julia environment are really orthogonal.

davidanthoff:
Actually, I just looked at the original Jupyter implementation again, and I think we should only create one controller per distinct _minor_ version of Julia that is installed. So say that a user has the following Julia versions installed: 1.5.2, 1.6.0 and 1.6.1. I think in that case we should create two controllers, namely
```ts
vscode.notebooks.createNotebookController('julia-1.6., 'jupyter-notebook', 'Julia 1.6.1 Kernel')
vscode.notebooks.createNotebookController('julia-1.5', 'jupyter-notebook', 'Julia 1.5.2 Kernel')
```
So essentially for every minor version we are only surfacing the highest patch version as a controller, and so Julia 1.6.0 does _not_ get a controller and essentially can't be used.

I think that would essentially mimic how this is implemented in IJulia.jl and the original Jupyter kernelspec story at the moment. It would also allow the Jupyter extension to pick the proper kernel: the string we use as the controller id would correspond to the `name` field in the `kernelspec` in `metadata` in the notebook file. Does that make sense to you?

DonJayamanne:
> that would essentially mimic how this is implemented in IJulia.jl and the orig

Will work on that.

Did we want to also add a placeholder entry if user doesn't have Julia installed. We're planning on doing that for Python, because we found that users tend to start without the runtime installed.
The suggestion is, if Julia isn't installed, add a placeholder entry `Julia`, and when user attempts to run a cell, we display a message indicating `Jualia` runtime isn't available.
Else not having a `Julia` entry could confuse some begginners.

davidanthoff:
> Did we want to also add a placeholder entry if user doesn't have Julia installed.

Yes, that sounds like a good idea! I guess we might think about this scenario outside of the notebook experience as well, pretty much nothing from our extension works if there is no Julia installed...

I think in terms of message probably best to just point users to julialang.org, right?

> This change would probably be best rolled out once the Julia extension official supports the notebook in the stable version

Agreed!

I think what I'll work on next is to not open a terminal per kernel, but instead add UI in the Julia Workspace for running kernels. I think that could work well, or did you plan to have shared UI in the Jupyter extension somewhere that shows all running kernels with options to stop and restart?

DonJayamanne:
> somewhere that shows all running kernels with options to stop and restart?

Can we hold off on that for a while. I'd like to check with the team

davidanthoff:
@DonJayamanne I just merged https://github.com/julia-vscode/julia-vscode/pull/2228 into the notebook branch. That PR integrates the notebook kernels into our workspace view, and that would be a pretty natural way now to also surface a stop and restart button.

davidanthoff:
@DonJayamanne Just a quick FYI, this is how the UI now looks for exploring variables and controlling kernels:

image

So there is now a top level node for the Julia REPL and each notebook kernel in our Workspace view, and then one can drill down to see the variables in each of these, and further drill down into arrays, dicts and structures etc.

There is also a button to restart and stop there for each kernel, which I think for now is good enough.

I'm going to fix some pretty basic bugs around restarts/stop and cleanup next, lifecycle for the various things is a bit of a mess right now :)

Have you started with the multi Julia version support? If not, I could also take a stab, I'm well into the code base at the moment and have some time.

davidanthoff:
@DonJayamanne I'm going to start the multi Julia version implementation now, fits in very nicely where I am right now :)

davidanthoff:
@DonJayamanne Alright, this is really coming together :) One question: what would happen if we were to merge this into `master` right now and started shipping it? Essentially nothing, right? Because the shipping Jupyter extension on VS Code Release has not yet switched to the native notebook UI? So would the situation essentially be that for users that use VS Code Insiders, they would get all the new stuff, but everyone else would just not even notice that we are shipping this stuff? I'm kind of tempted to merge this rather sooner than later, before this PR gets too large und unwieldy...

DonJayamanne:
> ne question: what would happen if we were to merge this into master right now and started shipping it? Essentially nothing, right? Because the shipping Jupyter extension on VS Code Release has not yet switched to the native notebook UI? So would the situation essentially be that for users that use VS Code Insiders, they would get all the new stuff, but everyone else would just not even notice that we are shipping this stuff? I'm kind of tempted to merge this rather sooner than later, before this PR gets too large und unwieldy...

Sorry, some how i missed this message.
I'd say go for it, that would be great.

> o would the situation essentially be that for users that use VS Code Insiders, they would get all the new stuff

Yes

> but everyone else would just not even notice that we are shipping this stuff

Yes, though some in Stable VS Code also get this new Notebook experience. We're rolling it out to a few users in stable slowly (controlled roll out).

> I'm kind of tempted to merge this rather sooner than later, before this PR gets too large und unwieldy...

Agreed.

Again, as all of the API used is currently in stable VS Code there should be no problem in merging this today.

DonJayamanne (approved):

pfitzseb (approved):
LGTM. Probably best to get this in as soon as possible so we can fix some issues before JuliaCon :)

PR #995, opened 2020-02-14
debug decorations

ZacLN:
add notification handler for staticlint debug notifications


davidanthoff:
What is the status of this?

PR #1002, opened 2020-02-17
Code coverage view

ZacLN:
File decorations for accompanying `.mem` files. 
Todo:
[x] Extend to .cov files
[x] Create gradient colour scheme
[x] Add title/menu option
[x] Checks to match mem/cov files against age of .jl file 

Ready for review and definitely needs another set of eyes to judge how info is displayed (I like orange), and how it's accessed


ZacLN:
This is not needed now right?

davidanthoff:
Well, it would still be nice to have as a feature. I would say we put it on the backlog and try to get higher priority features before Juliacon in, but I don't think we should close it, right?

pfitzseb:
Ideally we'd just recommend people to use one of the various lcov compatible coverage viewers on the marketplace, right? 
Haven't tested any of those so far though.

davidanthoff:
I've tried a lot of them, and for one reason or another none of these other packages is a super smooth experience. So I do think it would be nice if we had something native in the extension here that just works directly with the julia .cov files.

BUT, I think this is really not urgent: we do have the ability to export lcov files from test runs already, and I think we have other higher priority things right now on our plate for Juliacon. But if someone wants to pick this up here in a while, why not? We should then check these other packages again and see whether they have improved and whether a homegrown thing still makes sense, though.

PR #1003, opened 2020-02-18
Add support for the Julia workspace view

### LINKED ISSUE #516: Add a Workspace / Variables / Data Explorer
Note: This is not really an issue but an enhancement request.

This is the only thing I really miss IDE-wise coming from R. With R, you get RStudio and R Tools for VS which both feature a great variable explorer which makes it quite easy to view and edit data. (Especially helpful if you need to know why your code fails and if your data has been read correctly.)

I know that Juno _does_ support some of these things (however you cannot view large DataFrames etc.) but this would really bring Julia forward and make it a feasible language for people working with data.  
 
Although speed is the main selling point of Julia (which is great) I think you need to attract people who do not really need performance improvement over languages like R or Python but would like to try Julia for other reasons (e.g. the much nicer syntax compared to R).

davidanthoff:
Fixes #516.

@pfitzseb told me how we can have a better hook into the REPL, so this PR brings this feature back.

And it is improved :) Better UI for vars, a button to open things like tables in the grid viewer or plots in the plot pane directly from the workspace.

Open issues:
- [x] Add support for hierarchical drill down into composite, array and dict variables EDIT: deferred
- [x] Decide whether we should for example show functions defined? EDIT: for now lets just show them
- [x] Should we have our own Julia explorer, instead of putting this into the file thingy?
- [x] Don't show `@run` and `@enter` in the var list
- [x] Make sure `internal_vscodedisplay` works with anything we throw at it (in particular vectors and matrices) EDIT: Solved in https://github.com/julia-vscode/julia-vscode/pull/1258.


zhopan77:
Glad to see this feature is already being developed! This is one of the top 3 in my wish list to get our colleague to adopt Julia + VSCode as standard development environment. 

FY the other two being intellisense for local module on LOAD_PATH and loaded with "using" or "import"; and GUI for profiler. All these 3 features are currently supported by Juno, but we prefer VSCode's overall performance.

davidanthoff:
@julia-vscode/core I'm tempted to merge an initial version of this without tree drill-down, and then add that in a later version? If we were to go with that, this is almost ready to go.

We could for example ship a v0.16 that includes the inline eval results, the module selector and this PR, and all relatively soon?

pfitzseb:
Yeah, let's do that. The main difficulty with the tree drill functionality is that it needs to be lazy to be useful (in case of circular refs and really big objects, both of which happen all the time). Juno in theory has the code for that, but in practice it's buggy and might mess with the GC at times, so we should be very careful about the implementation.

davidanthoff:
Ha, I had hoped you had a good way to handle the lazy thing :) I've been thinking how to do this, and always ended up at "that is actually quite tricky"... We should probably look at the debug adapter protocol implementation again, that seems fairly robust and we could probably mimic that. But yes, lets do that later.

scripts/terminalserver/terminalserver.jl
```diff
@@ -18,6 +18,24 @@ end
...
+function getVariables()
+    M = Main
+    variables = []
+    for n in names(M)
+        !isdefined(M, n) && continue
+        x = getfield(M, n)
+        x isa Module && continue
+        x==Main.vscodedisplay && continue
+        n==Symbol("@run") && continue
+        n==Symbol("@enter") && continue
```
aviatesk, line 30: why we don't wrap those three into `_vscodeserver` ? `export @run, @enter; using _vscodeserver` won't work ?

scripts/terminalserver/terminalserver.jl
```diff
@@ -18,6 +18,24 @@ end
...
+        t = typeof(x)
+        value_as_string = Base.invokelatest(repr, x)
```
aviatesk, line 32: guess we want to filter out gensymed variables here ? something like `startswith(value_as_string, '#') && continue`

scripts/terminalserver/terminalserver.jl
davidanthoff: Yep, good idea! That won't change the need to filter them out here, right?

Should we do this in a separate PR?

scripts/terminalserver/terminalserver.jl
davidanthoff: Yes, agreed, I made a change for that.

scripts/terminalserver/terminalserver.jl
```diff
@@ -159,8 +179,7 @@ end
 
 run(conn_endpoint)
 
-@async begin
-
+@async try
```
pfitzseb, line 182: What's the point of this? The REPL connection is effectively dead after the `catch` anyways, no?

scripts/terminalserver/terminalserver.jl
```diff
@@ -418,6 +447,52 @@ end
...
+
+        x = Base.parse_input_line(line,filename=REPL.repl_filename(repl,main_mode.hist))
+
+        if !(x isa Expr && x.head == :toplevel)
+            error("VS Code Julia REPL got an unexpected input.")
+        end
+
+        # Replace all top level assignments with a global top level assignment
+        # so that they happen, even though the code now runs inside a
+        # try ... finally block
```
pfitzseb, line 463: FWIW, Juno doesn't bother with all of this and instead does something like
```
  main_mode.on_done = REPL.respond(Base.active_repl, main_mode; pass_empty = false) do line
    if !isempty(line)
      quote
        $evalrepl($mod, $line)
      end
    end
  end
```
```
function evalrepl(mod, line)
  try
    errored = false
    JSONRPC.send_notification(conn_endpoint, "repl/starteval", nothing)
    with_logger(JunoProgressLogger()) do
      try
         line = Meta.parse(line)
      catch err
        errored = true
        display_parse_error(stderr, err)
      end
      errored && return nothing
      try
        repleval(mod, line)
      catch err
        display_error(stderr, err, stacktrace(catch_backtrace()))
      end
    end
    errored ? nothing : ans
  catch err
    # This is for internal errors only.
    display_error(stderr, err, stacktrace(catch_backtrace()))
  finally
    JSONRPC.send_notification(conn_endpoint, "repl/finisheval", nothing)
  end
end
```

I'm not sure which approach I like better though.

scripts/terminalserver/terminalserver.jl
```diff
@@ -456,10 +529,29 @@ function vscodedisplay(x)
...
-        _vscodeserver._display(_vscodeserver.InlineDisplay(), x)
+        _display(InlineDisplay(), x)
     end
 end
 
+end
+
+atreplinit() do repl
+    @async try
+        sleep(1)
```
pfitzseb, line 540: This shouldn't be necessary, right?

src/interactive/workspace.ts
```diff
@@ -1,45 +1,77 @@
...
         }
     }
 
-    getTreeItem(node: string): vscode.TreeItem {
-        let treeItem: vscode.TreeItem = new vscode.TreeItem(node)
+    getTreeItem(node: WorkspaceVariable): vscode.TreeItem {
+        let treeItem = new vscode.TreeItem(`${node.name}:`)
+        treeItem.description = node.value;
+        treeItem.tooltip = node.type;
+        treeItem.contextValue = 'globalvariable';
```
pfitzseb, line 40: Can we make this depend on whether it's actually possible for us to display them in VSCode? That way the button would only show up for tables and plots and stuff?

src/interactive/workspace.ts
```diff
@@ -1,45 +1,77 @@
...
-export class REPLTreeDataProvider implements vscode.TreeDataProvider<string> {
-    private _onDidChangeTreeData: vscode.EventEmitter<string | undefined> = new vscode.EventEmitter<string | undefined>();
-    readonly onDidChangeTreeData: vscode.Event<string | undefined> = this._onDidChangeTreeData.event;
+interface WorkspaceVariable {
+    name: string,
+    type: string,
+    value: string
+}
+
+let g_replVariables: WorkspaceVariable[] = undefined;
```
aviatesk, line 12: I think this just can be `let g_replVariables: WorkspaceVariable[] = []` and remove the condition check at L27, reuse `clearVariables` at L74-L75 etc

src/interactive/workspace.ts
```diff
@@ -1,45 +1,77 @@
 import * as vscode from 'vscode';
+import * as repl from './repl';
 
 let g_terminal: vscode.Terminal = null
-let g_replVariables: string = '';
```
aviatesk, line 4: let's remove `g_terminal` as well, it's no longer needed

src/interactive/workspace.ts
```diff
@@ -1,45 +1,77 @@
...
 
-// TODO Enable again
-// let g_REPLTreeDataProvider: REPLTreeDataProvider = null;
+let g_REPLTreeDataProvider: REPLTreeDataProvider = null;
+
+export async function updateReplVariables() {
+    g_replVariables = await repl.g_connection.sendRequest(repl.requestTypeGetVariables, undefined);
+
+    g_REPLTreeDataProvider.refresh();
+}
```
aviatesk, line 51: maybe it's better to defined `updateREPLVariables` and `showInVSCode` in repl.ts ? 
We may want to use them in another feature in the future, I guess (e.g. documentation pane).
And it would be more safe to introduce `g_connection` validation check here.

scripts/terminalserver/terminalserver.jl
aviatesk: If we wrap them in `_vscodeserver`, then these condition check won't be needed since you get those names only from `Main`

aviatesk:
So one thing I think we need to address before merging would be how to handle "big" objects.
Say, if we execute `julia> collect(1:1000000)` then vsc will kinda freeze on low spec PC.
I guess we might need to use `strlimit` or something like `IOContext(:limit => true)` [here](https://github.com/julia-vscode/julia-vscode/blob/b628f1b93229d0b748580f751169a4d825c303e2/scripts/terminalserver/terminalserver.jl#L34)

aviatesk:
> So one thing I think we need to address before merging would be how to handle "big" objects.

well, I think a fairly simple solution would be done in this PR. Implementing more complicated and fancier solution like lazy display as in Juno shouldn't hold this PR imho.

scripts/terminalserver/terminalserver.jl
```diff
@@ -18,6 +18,26 @@ end
...
+        !isdefined(M, n) && continue
+        x = getfield(M, n)
+        x isa Module && continue
+        x==Main.vscodedisplay && continue
+        n_as_string = string(n)
+        n_as_string=="@run" && continue
+        n_as_string=="@enter" && continue
+        startswith(n_as_string, "#") && continue
+        t = typeof(x)
+        value_as_string = Base.invokelatest(repr, x)
```
pfitzseb, line 34: This should probably be `show`, no? Also would be best to `stringlimit` this, as @aviatesk mentioned below.

scripts/terminalserver/terminalserver.jl
aviatesk: xref: #1228

pfitzseb: That raises a good point, actually -- we should probably have a way to switch modules (or follow the editor module).

scripts/terminalserver/terminalserver.jl
pfitzseb: Or maybe just wrap 
https://github.com/julia-vscode/julia-vscode/blob/513c5cbedfa60043c02078db6dd85652b17e988c/scripts/terminalserver/terminalserver.jl#L89
into a function and reuse that.

aviatesk: This is not related to this PR, but `strlimit` already handles `isvalid` character case, so we don't need outer `isvalid` filtering, right ?

scripts/terminalserver/terminalserver.jl
aviatesk: yeah, while I think module switching can be done in another follow up PR.

scripts/terminalserver/terminalserver.jl
pfitzseb: Yeah, I just mindlessly copy-pasted code there ;)

scripts/terminalserver/terminalserver.jl
```diff
@@ -231,6 +250,18 @@ run(conn_endpoint)
                     JSONRPC.send_success_response(conn_endpoint, msg, safe_render(res))
                 end
             end
+        elseif msg["method"] == "repl/getvariables"
+            vars = getVariables()
+
+            JSONRPC.send_success_response(conn_endpoint, msg, [Dict{String,String}("name"=>i.name, "type"=>i.type, "value"=>i.value) for i in vars])
+        elseif msg["method"] == "repl/showingrid"
+            var = Core.eval(Main, Meta.parse(msg["params"]))
```
pfitzseb, line 258: This should probably be written as
```
            var = try
                getfield(Main, Symbol(msg["params"]))
            catch err
                Base.display_error(err, catch_backtrace())
                return
            end

            try
                Base.invokelatest(internal_vscodedisplay, var)
            catch err
                Base.display_error(err, catch_backtrace())
            end
```
or something similar.

scripts/terminalserver/terminalserver.jl
davidanthoff: We should probably have a tree drill down for modules in the workspace, right? Definitely a new PR, though :)

I'll do the rearrangements that @aviatesk suggested in a new PR once this one here is merged.

scripts/terminalserver/terminalserver.jl
davidanthoff: There is at least an error message shown in the REPL, otherwise the error just silently disappears. What we should really do is put a crash report into the catch block, I'll do that in a new PR once this one is merged (it requires a few more things).

scripts/terminalserver/terminalserver.jl
davidanthoff: I'm also unsure :) Should we merge as is for now, and then we can revisit later?

src/interactive/workspace.ts
davidanthoff: Not really, the problem are tables: in the Queryverse world you can have a query where you can't tell whether something is a table or not until you try to iterate the rows for it, i.e. until you actually try to materialize the results of the query. So that pretty fundamentally prevents us from telling with certainty whether we should show the button or not.

We could, though, narrow things down a lot: the case I described above is rare, and in very many cases we should actually be able to tell whether we can show the value or not.

I've put this onto my list for another follow up PR, ok?

src/interactive/workspace.ts
davidanthoff: Probably yes, but can we leave them in here for now? I wanted to reorganize the code a bit more in general in a follow up PR.

aviatesk: yep, of course. I'm welcome to craft that.

scripts/terminalserver/terminalserver.jl
davidanthoff: I had used `repr` here because that is what we use for the debugger. But I guess the situation here is different: for the debugger we ideally want the ability to round-trip things (because users can modify the value of a var in the tree view that shows vars), whereas here that is not a concern. Although: we could of course in the future add an ability to change variable values in the workspace view? In any case, I've changed it to `show` for now, we can iterate in future PRs.

scripts/terminalserver/terminalserver.jl
pfitzseb: Sure.

tj10ai:
What is the status of the debugger for VScode for Julia? It is working great except my debug windows for variables, stack... are blank when I run debug on Julia code. Is this because of a bug or it is me? Here is what I have installed: 

VScode_Julia_Config_ _Debug_No_Info
Version: 1.48.1 (system setup)
Commit: 3dd905126b34dcd4de81fa624eb3a8cbe7485f13
Date: 2020-08-19T17:12:13.244Z
Electron: 7.3.2
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Windows_NT x64 10.0.18363

In the attachment/screenshot you can see that I have not made any changes to the config - is this part of my problem getting this to work properly?

davidanthoff:
@tj10ai can you open a new issue for distinct problems? This was a PR for the workspace view.

Issue #1028, opened 2020-03-03
Revise not working correctly


cojua8:
Hello,

I'm developing a package, but changes are not being made by Revise. My startup.jl file already has the suggested code in the wiki

I have to do a manual Revise.revise() every time I need to debug some code.

How can I (or you) make it automatic?

Thank you!

davidanthoff:
How are running your code? I think we might have a bug with respect to this if you run your code with Alt+Enter. Is that how you are executing code?

cojua8:
Just F5

davidanthoff:
Ah, ok. F5 uses the same mechanism that Alt+Enter uses, so I'm pretty positive that it is the same underlying issue. Should actually be fairly simple to fix, in theory.

PR #1031, opened 2020-03-03
Invoke revise in repl/runcode

davidanthoff:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1028.

We need to run Revise.jl before we execute code if it is loaded.

@timholy, does this look sane to you? This is essentially just the place that executes if someone selects some text in VS Code and then sends that for execution in the REPL. With this PR we call `Revise.revise()` before we execute the user code. One question: Was that API always around for Revise.jl, i.e. will this code work with any version of Revise.jl that has ever shipped?

### DIFF
--- a/scripts/terminalserver/terminalserver.jl
+++ b/scripts/terminalserver/terminalserver.jl
@@ -150,7 +150,7 @@ end
             source_code = payload_as_string[end_second_line_pos+1:end]
 
             hideprompt() do
-                if isdefined(Main, :Revise)
+                if isdefined(Main, :Revise) && isdefined(Main.Revise, :revise) && Main.Revise.revise isa Function
                     let mode = get(ENV, "JULIA_REVISE", "auto")
                         mode == "auto" && Main.Revise.revise()
                     end
--- a/scripts/terminalserver/terminalserver.jl
+++ b/scripts/terminalserver/terminalserver.jl
@@ -150,6 +150,11 @@ end
             source_code = payload_as_string[end_second_line_pos+1:end]
 
             hideprompt() do
+                if isdefined(Main, :Revise)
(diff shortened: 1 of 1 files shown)


timholy:
Hooray!

An alternative strategy would be to let revise push something onto a list of hooks that vscode could run in a manner agnostic to what each hook is. See what we do for Atom: https://github.com/timholy/Revise.jl/blob/01ce17d293cbb551ad70746a65e274c41df00be9/src/Revise.jl#L1211-L1223

And yes, I think `Revise.revise()` should work for any version I can remember.

scripts/terminalserver/terminalserver.jl
```diff
@@ -150,6 +150,11 @@ end
             source_code = payload_as_string[end_second_line_pos+1:end]
 
             hideprompt() do
+                if isdefined(Main, :Revise)
```
ZacLN: I don't have any understanding of how Revise works - is it possible JULIA_REVISE is set without it being turned on in the `julia.useRevise` vscode setting? If so would we want a more robust check that Main.Revise points to the package?

davidanthoff: Ah, good point, `Main.Revise` could be something other than the Revise.jl package. I've added a couple more tests. Still not a 100%, but maybe this is robust enough for now?

And yes, Revise.jl can for example be loaded in the startup script of the user, even if our setting is to not load it. That is maybe something we should also think about: maybe we should have our own startup script (like IJulia does) and not run the default startup script in our REPL... But we can decide that later, I'd say.

davidanthoff:
@timholy I guess a hook would also be an option, but for now I almost prefer this approach here: it doesn't add an API that we then are committed to. Plus, we want things to work with old versions of Revise, so we would have to keep the currently working option around no matter what.

PR #1067, opened 2020-03-24
Add support for running jobs on clusters

davidanthoff:


vchuravy:
I am curious. What is the plan here? How will it be different from the remote SSH plugin of VSCode?

davidanthoff:
Ha, I was hoping you would bite ;)

It is all a bit unclear right now, to be honest. My current thinking is this: VS Code gets the ability to connect to a cluster (SLURM only for starters). I think in version one the idea would be that you open a folder with some code locally, and connect via SSH to the cluster somewhere else. The UI then shows you a list of running SLURM jobs. And then there is a simple way to submit a new job to the cluster. In my mind it would just copy everything you have in your local workspace over to the cluster, and then start a new SLURM job with a `src/main.jl` or something like that.

But, I'm mostly playing around at this point.

I tried to use the SSH plugin with our cluster here at Berkeley, but the disable TCP port forwarding or something like that, so that it generally doesn't work, so right now I'm thinking that this would not utilize the SSH plugin.

vchuravy:
> I tried to use the SSH plugin with our cluster here at Berkeley, but the disable TCP port forwarding or something like that, so that it generally doesn't work, so right now I'm thinking that this would not utilize the SSH plugin.

Well that is just unkind. 

I think the big issues are:
1. State/File-system synchronization, the data from the project needs to be copied across which is meh and then you need to make sure that julia is loaded, and that the Project/Manifest is synchronized copied, as well as instantiated and precompiled
2. Interacting with SLURM, it would be super-cool to have a VSCode slurm integration, I vaguely recall that you could in theory run slurm locally and tell it to connect to the cluster and therfore run something like `squeue` on your laptop.

davidanthoff:
> State/File-system synchronization, the data from the project needs to be copied across which is meh and then you need to make sure that julia is loaded, and that the Project/Manifest is synchronized copied, as well as instantiated and precompiled

Yes, that is the plan. Right now I'm thinking that really shouldn't be too complicated, but we'll see. At least on the Berkeley cluster I have a pretty clear idea how that would all work. It probably won't be able to handle large datasets automatically, I think folks would have to come up with some other solution for that.

> Interacting with SLURM, it would be super-cool to have a VSCode slurm integration, I vaguely recall that you could in theory run slurm locally and tell it to connect to the cluster and therfore run something like squeue on your laptop.

I was planning to just run it on the cluster via the SSH connection. The minimal requirement would really only be that one can log into the cluster via SSH and then have the SLURM commands available there, which seems a pretty safe assumption?

vchuravy:
> I was planning to just run it on the cluster via the SSH connection. The minimal requirement would really only be that one can log into the cluster via SSH and then have the SLURM commands available there, which seems a pretty safe assumption?

Can you keep the SSH connection open and reuse it? There are several clusters who have "fun" login procedures.

>  Yes, that is the plan. Right now I'm thinking that really shouldn't be too complicated, but we'll see. At least on the Berkeley cluster I have a pretty clear idea how that would all work. It probably won't be able to handle large datasets automatically, I think folks would have to come up with some other solution for that.

Yeah it is hard to come up with something cleverer, in the past I have used `sshfs` or some clever file synchronisation @maleadt put together. Just copying the data is probably fine for projects where you can instantiate heavy dependencies separately, but I wouldn't want to scp CLIMA everytime I make a change.

davidanthoff:
> Can you keep the SSH connection open and reuse it? There are several clusters who have "fun" login procedures.

I can, but good to know that other clusters don't like that...

> Just copying the data is probably fine for projects where you can instantiate heavy dependencies separately, but I wouldn't want to scp CLIMA everytime I make a change.

So my idea was (for the start) that I just copy the whole content of the current workspace over, and leave it there. So essentially this would only work for a scenario where you don't have too much stuff in your main project folder, and then other code is pulled in via the package manager, and data via some other mechanism that you need to handle.

How is the CLIMA stuff structured? Is it all in packages, or some other model? And you guys are working on a SLURM cluster, right? Maybe we could try to get the initial version working on both the Berkeley and your cluster?

PR #1074, opened 2020-03-25
Update julia icon

### LINKED ISSUE #1045: Update extension icon so it's visible in dark mode
Unfortunately, as far as I can tell, it's not possible to provide separate light and dark-theme icons to VS Code Extensions... So, it might be nice if the icon were updated to use an intermediate color for the `julia` letters so that it would be more visible in both light and dark mode:

image

CameronBieganek:
Fixes #1045.

Update julia logo icon so that it works well for both light and dark modes. The PNG was created from the SVG following the guideline in [this Stack Overflow post](https://superuser.com/a/723031/1063997). In particular,

```
brew install librsvg
rsvg-convert -h 400 julia-logo.svg > julia-logo.png
```

The dimensions of the PNG are 640x400. I prefer to err on the side of too high of a resolution.

The changes from the official Julia logo were as follows:
- Change the fill color of the letters to white.
- Change the stroke-width of the letters from 0 to 3 (and choose black for the stroke color).
- Make the viewBox slightly longer to accommodate the additional letter width from having a non-zero stroke-width.


davidanthoff:
Fantastic, thanks!

Issue #507, opened 2018-06-11
could not read package directory if .juliarc.jl script contains print function.


YongHee-Kim:
If  `.juliarc.jl` file contains `print` like this
``` julia
println("Hello world!")
# do something...
```
Then if I try to open package directory at command palette, `Error:could not read package directory` pops up.
![image](https://user-images.githubusercontent.com/7087745/41216791-c394831e-6d90-11e8-8d2a-e503c7b96b40.png)

To see what happens at `...\julialang.language-julia-0.10.2\out\openpackagedirectory.js`
I've added debug print at L27 
``` javascript
27            var juliaVersionHomeDir = yield packagepath.getPkgPath();
28            vscode.window.showInformationMessage(juliaVersionHomeDir)
```

and it seems package directory is corrupted with print function from `.juliarc.jl`
![image](https://user-images.githubusercontent.com/7087745/41216987-72206790-6d91-11e8-8be1-410cd8c4db92.png)

davidanthoff:
That is a bug, thanks for reporting. The fix is pretty simple: we need to add a command line flag [here](https://github.com/JuliaEditorSupport/julia-vscode/blob/604479905331b1e9333674d999b842f8828ee614/src/packagepath.ts#L16) to not run the startup script.

Issue #672, opened 2019-01-30
intellisense for modules


sdewaele:
Is it possible to enable intellisense for content of modules that are in the current path, but not in a package? Suppose I have created a module `MyModule` that defines a function  `myfunction`. When I use this module in a script, it would be great if I could get autocompletion/help/code navigation for it. I have added it to LOAD_PATH, even in `startup.jl`, but it is not detected by the plugin.

davidanthoff:
I think it used to work for modules, so I would think that it should work again once we fix the julia 1.1 compat issues. But not sure, @ZacLN knows much more about that part of the extension than I :)

sdewaele:
FYI - I encountered this issue using Julia 1.0.2.

ZacLN:
Hi, if the module is not part of a package known by `Pkg` then you'll have to explicitly `include`the containing file, as you would when running a script. 

Using `startup.jl` files is another issue that does need to be addressed

sdewaele:
Thanks! Unfortunately IntelliSense is still spotty. I have taken the following steps in Julia 1.0.1.1: 1) `generate` a package in the package manager, see code below; 2) create a git repo in the root dir of this package. 3) Registering it in package mode with `dev StijnPackage`. 4) Made some small changes (`export greet`, add a comment.  The package manager is aware of the package, see output below. In the console, I can use the package.

After this, I occasionally get IntelliSense for the package, oftentimes none. When I do, I get an old version. For example, the documentation I added below for `greet` does not show up, even when I commit the change to the repo. Do I perhaps need to trigger the plugin somehow to become aware of the change? 

```
module StijnPackage

export greet

"greet user"
greet() = print("Hello World!")

end # module
```
In package mode:
```
(v1.0) pkg> status
    Status `C:\Users\marionstijn\.julia\environments\v1.0\Project.toml`
  [f2e7f880] MatrixExponential v0.1.0 [`..\..\..\source\MatrixExponential`]
  [295af30f] Revise v1.0.2
  [697ea9e2] StijnPackage v0.1.0 [`..\..\..\source\StijnPackage`]
  [fce5fe82] Turing v0.6.8
```

sdewaele:
Any thoughts on this? Perhaps Am I missing a step here?

sdewaele:
[...]

PR #1084, opened 2020-03-31
Do not append dev directory to PkgPath

caleb-allen:
There may be some context I'm missing here, but the modification in this PR fixes a broken home directory reference for me, specifically in this file:

https://github.com/julia-vscode/julia-vscode/blob/dcbeaf7976f3d0a3516331a0153f56c4270179b3/src/openpackagedirectory.ts#L22

Possibly related to #507

### DIFF
--- a/src/packagepath.ts
+++ b/src/packagepath.ts
@@ -20,7 +20,7 @@ export async function getPkgPath() {
         let jlexepath = await juliaexepath.getJuliaExePath();
         // TODO: there's got to be a better way to do this.
         var res = await exec(`"${jlexepath}" --startup-file=no --history-file=no -e "using Pkg;println(Pkg.depots()[1])"`);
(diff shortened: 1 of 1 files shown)


caleb-allen:
Related to #1084

caleb-allen:
Related to #672

davidanthoff:
I don't think that is related to either of the two issues you opened. It is not clear to me what this PR would fix, the current behavior seems correct: we want to show all the packages in `.julia/dev` when someone runs the open package command. With this PR we would show the content of `.julia`, which is not correct, right?

caleb-allen:
Yes, I think you're right. It seems this was the correct behavior, and my understanding of the correct way to handle packages was incomplete.

Thanks! Apologies for the unnecessary issue.

davidanthoff:
No worries, opening a PR is always a good thing :)

PR #1098, opened 2020-04-04
Fix a wrong config name

davidanthoff:
@tk3369 found this one, at some point we must have renamed the config setting.


tk3369:
Great! Thanks!

PR #1101, opened 2020-04-06
allow code block execution in JMD files

### LINKED ISSUE #1100: Treat .jmd markdown julia files as `julia` language
Is it possible to allow in `Weave.jl` files *.jmd to have the same code execution logics as in *.jl files.
So that language-julia.executeJuliaCellInREPL  also works well in *.jmd files and executes either `##` fencing, or ```julia fencing, whatever comes first.

phlavenk:
Fixes #1100

### DIFF
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -684,16 +684,23 @@ async function selectJuliaBlock() {
 
 async function executeJuliaCellInRepl() {
     telemetry.traceEvent('command-executejuliacellinrepl');
-
     let ed = vscode.window.activeTextEditor;
     let doc = ed.document;
     let rx = new RegExp("^##");
+    let rx_jmd_start = new RegExp("^```julia");
+    let rx_jmd_stop = new RegExp("^```");
+    let stop_type_jmd = false
+
     let curr = ed.selection.active.line;
     var start = curr;
     while (start >= 0) {
         if (rx.test(doc.lineAt(start).text)) {
             break;
-        } else {
+        }
+        else if (rx_jmd_start.test(doc.lineAt(start).text)) {                
+            break;
+        }
+        else {
             start -= 1;
         }
     }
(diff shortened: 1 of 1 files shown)


davidanthoff:
As I explained over at #1100 I think we should hold-off for now and see whether we will replace the jmd editing experience with the new notebook UI. But lets leave this PR open, can always happen that those notebook plans don't work out and then we could revisit this PR :) Thanks for opening it!

phlavenk:
Hi David,
thank you for your comment. Do you have an idea how far they are with the implementation of the notebook-style in core vscode? if we are talking about few months, that's OK. If they need more than 5 months, I think its worth to have a temporal workaround for that time period. 
And again let me express a big thank for what you and Zac are doing with vscode support - it's a pleasure to use it.

davidanthoff:
It already works in the insiders build of VS Code, so my best guess is a lot sooner than five months.

phlavenk:
Thank you for the update, I'm looking for the new notebook interface.

storopoli:
Hey Since it's been a while and according to #980 the VSCode Notebook implementation force us to use Python, we could give this PR a go? Would be lovely to execute `*.jmd` cells with `alt+enter`!

davidanthoff:
The new Jupyter extension for VS Code drops the dependency on Python entirely, I believe.

storopoli:
@davidanthoff not quite we still have to use `IJulia` that install a miniconda inside `.julia` folder.
See [this post from Microsoft in November 2020](https://devblogs.microsoft.com/python/introducing-the-jupyter-extension-for-vs-code/) (specially bullet point 2):

> Try it out!
>* To work with Python Jupyter Notebooks in VS Code, simply install or update the Python extension. There’s no need to install the Jupyter extension separately. For those who are using the VS Code Insiders build, you may notice that the new preview notebooks experience that was first introduced in July has now been turned on by default.
>* **To use the Jupyter extension for notebooks other than Python, install the VS Code Insiders build, the Jupyter extension, the Jupyter kernel of your language (e.g. kernel for the Julia language), and start playing around!**
>* To contribute to the Jupyter extension, check out our contribution guide. This AI Tools sample extension demonstrates how to use the custom authentication API that’s exposed through the Jupyter extension.

davidanthoff:
I believe IJulia.jl only installs miniconda if you try to start a notebook server via it. If you just install the package IJulia, it will register the Julia kernel with Jupyter, but it will _not_ install anything Python. And I think that should be enough to make everything work with the Jupyter extension for VS Code. Having said that, I do hope that we can get that to a point where one doesn't even have to install IJulia.jl to make things work inside VS Code.

I should also say that I'm not opposed to this PR here, I just don't have the bandwidth right now to review it. But if someone else from the core team reviews, we should be all good.

src/repl.ts
```diff
@@ -684,36 +684,65 @@ async function selectJuliaBlock() {
 
 async function executeJuliaCellInRepl() {
     telemetry.traceEvent('command-executejuliacellinrepl');
-
     let ed = vscode.window.activeTextEditor;
     let doc = ed.document;
     let rx = new RegExp("^##");
+    let rx_jmd_start = new RegExp("^```julia");
```
storopoli, line 690: what about `"^```\{julia"` ? It is also a chunk option in `Literate.jl`. I cannot help since my knowledge in JavaScript is `NULL`...

pfitzseb:
This has been implemented now.

storopoli:
Thank you!

Issue #1090, opened 2020-04-01
How to set JULIA_EDITOR in a Julia REPL in a docker container?


hsm207:
I have the following line In `devcontainer.json`:

`"runArgs": [ "-e", "JULIA_EDITOR=code-insiders"]`

While inside the container, if I execute `Julia: Start REPL`  check the value of the `JULIA_EDITOR` environment variable:
```
julia> ENV["JULIA_EDITOR"]
"\"/home/user/.vscode-server-insiders/bin/6cf3dd97465b2f4f666b66758991d720e003b0f4/node\""
```

On the other hand, if I instead create a new integrated terminal and run Julia from inside it, the value of `ENV["JULIA_EDITOR"]"` is equal to the value set in `runArgs`.

Is there a config somewhere that hard codes the value of `JULIA_EDITOR` inside a Julia REPL?

I am using version 0.15.

davidanthoff:
We currently set these env vars before we start the Julia REPL [here](https://github.com/julia-vscode/julia-vscode/blob/master/src/repl.ts#L173) and [here](https://github.com/julia-vscode/julia-vscode/blob/master/src/repl.ts#L199). One thing that is almost certainly wrong there is that those two lines of code are different :)

The other questions is what the ideal setting for the remote case is... Presumably the current setting does not work, right?

hsm207:
Yes, the current setting does not work for me.

What was the reason to define this environment variable in the first place?

I only came across this bug because I tried to use the @edit macro. Looking the Julia code [here](https://github.com/JuliaLang/julia/blob/b8e9a9ecc62ab49003bd4f1834771bdeb8cb1aa2/stdlib/InteractiveUtils/src/editless.jl#L167-L179), it already has fallbacks when none of the `JULIA_EDITOR`, `VISUAL` or `EDITOR` environment variables are defined.

davidanthoff:
We want to make sure that if you use `@edit` inside the REPL, files open up in VS Code.

hsm207:
if that's the case, why do we need to call `"${process.execPath}"`? Don't we just need to set the environment variables to be `code` or `code-insiders` regardless of platform?

davidanthoff:
That would probably not work if someone didn't add VS Code to the `PATH`, right?

hsm207:
[...]

PR #1103, opened 2020-04-10
Fix JULIA_EDITOR to work on remote environments

hsm207:
See issue #1090 for details.

### DIFF
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -160,7 +160,7 @@ function is_remote_env(): boolean {
 }
 
 function get_editor(): string {
-    if (is_remote_env()) {
+    if (is_remote_env() || process.platform == 'darwin') {
         let cmd = vscode.env.appName.includes("Insiders") ? "code-insiders" : "code"
         return `"${path.join(vscode.env.appRoot, "bin", cmd)}"`
     }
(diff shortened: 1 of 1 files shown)


src/repl.ts
```diff
@@ -196,7 +210,7 @@ async function startREPL(preserveFocus: boolean) {
                     shellPath: exepath,
                     shellArgs: jlarg1.concat(jlarg2),
                     env: {
-                        JULIA_EDITOR: process.platform == 'darwin' ? `"${path.join(vscode.env.appRoot, 'bin', 'code')}"` : `"${process.execPath}"`,
+                        JULIA_EDITOR: process.platform == 'darwin' ? `"${path.join(vscode.env.appRoot, 'bin', 'code')}"` : get_editor(),
```
davidanthoff, line 213: Actually, while you are editing this part: could you move this check for MacOS into the `get_editor` function as well? Then it would apply for both cases where we set `JULIA_EDITOR`, which it should, i.e. you could fix an old bug with this PR as well :)

hsm207: done.

davidanthoff:
Excellent, thank you!

Issue #1096, opened 2020-04-03
Option to preserve existing JULIA_NUM_THREADS setting


oschulz:
It would nice to have a value for `julia.NumThreads` (empty string or so?) that results in keeping a `$JULIA_NUM_THREADS` already set in the environment (e.g. before starting VS Code).

We usually have machine-dependent settings for `$JULIA_NUM_THREADS` (depending on number of cores and hyperthreading), so having a way to tell the Julia extension to keep her fingers off it would  be cool.

hsm207:
Let's close this issue. It's been fixed in PR #1104 .

PR #1104, opened 2020-04-11
Prioritize JULIA_NUM_THREADS if it exists

hsm207:
See #1096 for details.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -23,6 +23,18 @@
             "sourceMaps": true,
             "outFiles": [ "${workspaceRoot}/out/test/**/*.js" ],
             "preLaunchTask": "npm: compile"
+        },
+        {
+			"name": "Run Extension Tests",
+			"type": "extensionHost",
+			"request": "launch",
+			"runtimeExecutable": "${execPath}",
+			"args": [
+				"--extensionDevelopmentPath=${workspaceFolder}",
+				"--extensionTestsPath=${workspaceFolder}/out/test/suite/index"
+            ],
+			"outFiles": ["${workspaceFolder}/out/test/**/*.js"],
+			"preLaunchTask": "npm: watch"
         }
     ]
 }
--- a/package.json
+++ b/package.json
@@ -455,8 +455,10 @@
                     "scope": "application"
                 },
                 "julia.NumThreads": {
-                    "type": "integer",
-                    "default": 1,
+                    "type": ["integer",
+                             "null"
+                    ],
+                    "default": null,
                     "description": "Number of threads to use for Julia processes."
                 },
                 "julia.format.indent": {
--- a/package.json
+++ b/package.json
@@ -455,8 +455,9 @@
                     "scope": "application"
                 },
                 "julia.NumThreads": {
-                    "type": ["integer",
-                             "null"
+                    "type": [
+                        "integer",
+                        "null"
                     ],
                     "default": null,
                     "description": "Number of threads to use for Julia processes."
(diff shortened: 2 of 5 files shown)


oschulz:
> if an env var JULIA_NUM_THREADS exists, we use that  [...] otherwise, we pass 1 as the value

Hm, if neither `julia.NumThreads` is set, nor `JULIA_NUM_THREADS`, then I'd suggest we don't do anything, letting Julia use it's default behavior (which may or may not change to automatically use all available thread in the future)? I'd day that's the natural, "least-surprise"-principle thing to do.

davidanthoff:
Yeah, that would make even more sense, I agree.

Is setting `JULIA_NUM_THREADS` to an empty string equivalent to not setting it at all? That would allow us to keep the code somewhat cleanly structured...

hsm207:
@davidanthoff 

> First, we change the extension setting julia.NumThreads to have an empty default value.
I edited `package.json` to implement this. Is this the way to do it?

> Is setting JULIA_NUM_THREADS to an empty string equivalent to not setting it at all? That would allow us to keep the code somewhat cleanly structured...

It looks like setting `JULIA_NUM_THREADS` to anything other than positive integers is equivalent to not setting it at all.

Executing the following from the command line will give the same results i.e. 1:

```
julia -e "Threads.nthreads() |> println"
JULIA_NUM_THREADS=1 julia -e "Threads.nthreads() |> println"
JULIA_NUM_THREADS='' julia -e "Threads.nthreads() |> println"
JULIA_NUM_THREADS="" julia -e "Threads.nthreads() |> println"
JULIA_NUM_THREADS="''" julia -e "Threads.nthreads() |> println"
JULIA_NUM_THREADS='""' julia -e "Threads.nthreads() |> println"
JULIA_NUM_THREADS=hello julia -e "Threads.nthreads() |> println"
```

> Yeah, that would make even more sense, I agree.

I don't follow what the fallback hierarchy should be given  @oschulz's comment. Could you please write it out?

davidanthoff:
> I edited package.json to implement this. Is this the way to do it?

Yes, that looks good to me!

> I don't follow what the fallback hierarchy should be given @oschulz's comment. Could you please write it out?

Sure, I think the fallback story should be this now:
- if julia.NumThreads has a value, we use that, no matter what.
- otherwise, if an env var JULIA_NUM_THREADS exists, we use that.
- otherwise, we pass an empty string as the value

So with that we essentially have a situation where the VS Code config element has the highest priority, followed by the ENV var. If neither is set, we just don't pass anything (by setting `JULIA_NUM_THREADS` to an empty string) and let Julia use whatever the default is.

hsm207:
@davidanthoff 

I've made the changes.

I tested it manually and it works.

I couldn't get the unit test to work because I couldn't figure out how to mock `vscode.workspace.getConfiguration("julia")`. Do you know how?

davidanthoff:
Fantastic!

No idea how we could get the unit tests to run... To be honest, this would be the first piece of code in the TypeScript part of the extension that would have tests, i.e. this is something we need to generally figure out how to do. I didn't want to hold this PR back for this.

oschulz:
Thanks again for this! @davidanthoff , will there be a new release with this feature (or maybe there is, already)?

davidanthoff:
We moved to a rolling release schedule now, so yes, this should be out really soon.

The way things work is that whenever we have something new on `master`, I release a build to the insider channel. If that build survives 24 hours without major new regressions from telemetry, I push that build out to everyone. Right now we had 2-3 builds that had regressions, so I haven't published anything to release in a couple of days, but fingers crossed we'll have a stable build soon (tomorrow or so), and then it will go out to everyone.

oschulz:
Thanks!

PR #1125, opened 2020-04-18
Update grammar

davidanthoff:
@twadleigh if there is a chance that you could review this it would be awesome!

Here is the (somewhat stupid) backstory. I had forgotten that we have https://github.com/julia-vscode/julia-vscode/blob/master/syntaxes/update_syntax.jl... So sometimes in the fall I did my manual updating, at which point the extra replacements from that script were no longer included in the shipping version of the grammar.

Then, today, I started adding automatic download and conversion from cson2json to my "new" [update script](https://github.com/julia-vscode/julia-vscode/blob/master/src/scripts/updateDeps.ts). When all was done, I rediscovered that we already have an update/conversion script. Grrr, that was really silly on my end. In any case, I'm now using a mix: I'm downloading and doing the conversion in the TypeScript part, and then call the Julia script you created to do the regex replacements. I would probably make sense to move those regex replacements into native TypeScript as well, then things would be a little smoother. But hey, for now it works.

BUT, we now haven't used a grammar with these extra replacements for quite a while and I'm a bit nervous about shipping it. @twadleigh if you could just take a quick look whether it is still making the kind of changes you intend it to make? Maybe you even have some example files where you can try this grammar out?

To me these grammar files are a mystery, so I wouldn't even know where to start...

### DIFF
--- a/syntaxes/julia.json
+++ b/syntaxes/julia.json
@@ -62,6 +62,10 @@
           "name": "meta.array.julia",
           "patterns": [
             {
+              "match": "\\bbegin\\b",
+              "name": "constant.numeric.julia"
+            },
+            {
               "match": "\\bend\\b",
               "name": "constant.numeric.julia"
             },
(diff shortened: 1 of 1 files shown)


davidanthoff:
@ZacLN should we just merge this and see how it goes?

PR #1159, opened 2020-05-01
Moves the symbol store out of the extension path

davidanthoff:
Requires https://github.com/julia-vscode/LanguageServer.jl/pull/654. 

I now just create a folder `symbolstorev1` in the global state folder that we get from VS Code. If we change the format of the cache, we can just update that path to `v2`, and then we need to find some way to delete the old folder, but that should all be fairly simple. We just need to make sure we make that adjustment if anything about the store format changes.


davidanthoff:
@ZacLN and ready for review. Could you maybe try this branch? Just to make sure...

ZacLN:
Will do

ZacLN:
Do we know where `g_context.globalStoragePath` is?

davidanthoff:
On Windows it is at `~AppData\Roaming\Code\User\globalStorage\julialang.language-julia`, which is a "good" location. No idea where it is on other platforms :) I added some output, probably in general a good idea if one can easily see the path.

ZacLN:
OK got it: `/.config/Code/User/globalStorage/julialang.language-julia/symbolstorev1/`.

Could we merge https://github.com/julia-vscode/SymbolServer.jl/pull/147 before this goes in (or you make a new release) - I can't face re-caching everything again...

davidanthoff:
Yep, I merged the other PR first, so now `master` should have this and the other PR.

Issue #1170, opened 2020-05-07
Add missing configuration options


davidanthoff:
@ZacLN I think we probably need new configuration settings here in `package.json`, right? I just updated all the submodules and other deps here in the repo, but I'm not entirely sure what we might need in terms of new config options. Could you fix that?

davidanthoff:
@ZacLN we now need new PRs for various packages to reflect the config options in #1172 before we can ship a new version, right?

ZacLN:
I don't think so, which did you have in mind?

davidanthoff:
`julia.lint.missingrefs`, I thought because the values of the enum are new we would have to update something in the packages to match those values?

ZacLN:
No, fortunately the handling of the option in [StaticLint](https://github.com/julia-vscode/StaticLint.jl/blob/master/src/linting/checks.jl#L492) doesn't care about the name of the intermediate (i.e. nogetfield) mode.

davidanthoff:
Ah, cool :) I'll update submodules and tag now.

PR #1172, opened 2020-05-08
update julia config options

ZacLN:
#1170

### DIFF
--- a/package.json
+++ b/package.json
@@ -360,10 +360,10 @@
                     "default": "all",
                     "enum": [
                         "none",
-                        "getfield missing",
+                        "symbols",
                         "all"
                     ],
-                    "description": "Highlight unknown symbols. The `getfield missing` option will not mark unknown fieldnames."
+                    "description": "Highlight unknown symbols. The `symbols` option will not mark unknown fields."
                 },
 
                 "julia.lint.call": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -356,45 +356,67 @@
...
                 },
                 "julia.lint.missingrefs": {
-                    "type": "boolean",
-                    "default": true,
-                    "description": "Report possibly missing references."
+                    "type": "string",
+                    "default": "all",
+                    "enum": [
+                        "none",
+                        "nogetfield",
```
davidanthoff, line 363: Could we use something like `getfield missing`, or just generally a string that is less an identifier? I think this will show up in the UI, right? Or maybe there is a way to specify a display string for each of the values that can be chosen that would then appear in the UI?

package.json
```diff
@@ -356,45 +356,67 @@
...
-                    "default": true,
-                    "description": "Report possibly missing references."
+                    "type": "string",
+                    "default": "all",
+                    "enum": [
+                        "none",
+                        "getfield missing",
+                        "all"
+                    ],
+                    "description": "Highlight unknown symbols. The `getfield missing` option will not mark unknown fieldnames."
```
davidanthoff, line 366: Is this right? I would have thought that the option `getfield missing` would _mark_ the cases where I'm accessing a field that doesn't exist?

Not for this release, but should we maybe just split this into two bool options, i.e. have just have one `bool` option that enables the `f.a` detection stuff?

PR #1177, opened 2020-05-10
Fix a bug in REPLTree

davidanthoff:
Without this the CI compile doesn't finish. Not sure why this is showing up now, probably I updated some TypeScript defs to the latest version. In any case, should be harmless, this isn't used right now in any case :)


davidanthoff:
Merged without review because a) trivial and b) fixes the build :)

PR #1187, opened 2020-05-11
Handle empty file case in debugger

davidanthoff:
Fixes a bug from crash reporting. This essentially finishes the debugger if one tries to debug an empty file.


davidanthoff:
@ZacLN could you approve this?

PR #1208, opened 2020-05-15
Remove dead code from REPL

davidanthoff:
This removes a whole lot of code that isn't currently used. Some of it might come back when we re-add features, but for now I just want to clean up a bit before I start to rewrite the communication part.


davidanthoff:
As soon as this is merged I'll open a PR on top of this that has a rewrite of the communication between the REPL and the extension based on JSON RPC. That code if done and finished, but we first need to merge this here.

PR #1209, opened 2020-05-15
Properly fix a bug

davidanthoff:
Turns out https://github.com/julia-vscode/julia-vscode/pull/1186 is a bug everywhere...

This is not super elegant but works everywhere. I plan to replace the communication code with JSON RPC very soon, so this is a short term bug fix.

We should merge ASAP because minimally master debugging is completely broken right now, not sure whether I had shipped the fix widely already.


davidanthoff:
I'm going to merge without review, I think this needs to get onto master ASAP :) If someone wants to review after the fact it would be great!

Issue #1203, opened 2020-05-14
Improve RPC protocol for terminalserver


pfitzseb:
Would be nice if this was a bit less ad-hoc and allow for both messaging and two-way communication.

The callback mechanism Juno uses should work fine here as well.

davidanthoff:
Yes, the current implementation is a mess :)

Do you have a link to the Juno code for this? I'd like to take a look.

I actually think that if we were to combine the variuos bits and pieces that are spread out over the LS, debug and REPL code base, and just picked the best aspect of each, we would also end up with something reasonable :)

I think my gut preference would be that we just adopt JSON-RCP for this, as that is already used on the LS side and we could probably use the same implementation for both. The existing Julia implementation of that on the LS side I think is solid, i.e. it supports request-response type patterns, and notifications, and has things like correlating responses to the right request etc. implemented.

I think right now the implementation also uses two named pipes, which is also silly, we should just use one.

I'll take a quick look later today whether we can just reuse some of the existing code.

pfitzseb:
> I think my gut preference would be that we just adopt JSON-RCP for this

Yeah, agreed. The only question is how to load JSON into the Julia process, but I guess vendoring JSON@0.20 (which doesn't have any deps) should be fine.

> Do you have a link to the Juno code for this? I'd like to take a look.

Not really, no. At least the JS side is a bit hard to understand. The [Julia side](https://github.com/JunoLab/Atom.jl/blob/master/src/comm.jl) is nice though, imho.

davidanthoff:
> The only question is how to load JSON into the Julia process, but I guess vendoring JSON@0.20 (which doesn't have any deps) should be fine.

Agreed.

> At least the JS side is a bit hard to understand.

We could try to just use https://www.npmjs.com/package/vscode-jsonrpc on the TypeScript side?

> The Julia side is nice though, imho.

Yes, I think the callback style API makes a lot of sense. I implemented something like that (but without an API or the elegance of the code you linked to) for the debugger [here](https://github.com/julia-vscode/julia-vscode/blob/4d06225f009944ade1a730cbea931236bf4a0f28/scripts/debugger/debugger.jl#L53).

[...]

PR #1210, opened 2020-05-15
Use JSON-RPC for REPL-extension comm

davidanthoff:
Improves #1203.

Major changes:
- Uses JSON RPC
- We use the same implementation in the REPL that we use in LS on the Julia side (I moved the code into a package)
- On the TypeScript side we use the vscode-jsonrpc package
- We use only one named pipe instead of multiple in the previous version
- All the weird message format hacks are gone :)
- We vendor JSON.jl into the REPL process, which will be super useful for lots of other things as well
- I'm starting to move all the Julia packages that we ship into the folder `scripts/packages`, right now they are all over the place.

Once we have this merged, I'll also use this for the debugger communication.

I don't yet have the kind of callback implementation that Juno has. The main reason is that I realized that won't work in the simple form for LS, but I want the JSONRPC package to work for that scenario as well. I'll open another issue to discuss the specific problem.

### DIFF
--- a/scripts/terminalserver/terminalserver.jl
+++ b/scripts/terminalserver/terminalserver.jl
@@ -87,8 +87,6 @@ run(conn_endpoint)
                 try
                     VSCodeDebugger.startdebug(debug_pipename)
                 catch err
-                    Base.display_error(err, catch_backtrace())
-                    sleep(10)
                     VSCodeDebugger.global_err_handler(err, catch_backtrace(), ARGS[4])
                 end
             end
--- a/scripts/terminalserver/terminalserver.jl
+++ b/scripts/terminalserver/terminalserver.jl
@@ -1,59 +1,55 @@
 module _vscodeserver
 
-include("repl.jl")
-include("../debugger/debugger.jl")
-
 using REPL, Sockets, Base64, Pkg, UUIDs
 import Base: display, redisplay
 import Dates
 
+include("../languageserver/packages/JSON/src/JSON.jl")
+
+module JSONRPC
+    import ..JSON
+    import ..UUIDs
+
+    include("../packages/JSONRPC/src/core.jl")
+end
+
+include("repl.jl")
+include("../debugger/debugger.jl")
+
 struct InlineDisplay <: AbstractDisplay end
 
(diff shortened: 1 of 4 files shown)


davidanthoff:
@ZacLN: just fixed another small problem here, but now it should be ready for review.

scripts/terminalserver/terminalserver.jl
```diff
@@ -1,59 +1,55 @@
...
 function ends_with_semicolon(x)
     return REPL.ends_with_semicolon(split(x,'\n',keepempty = false)[end])
 end
 
+repl_conn = connect(repl_pipename)
+
+conn_endpoint = JSONRPC.JSONRPCEndpoint(repl_conn, repl_conn)
+
+function sendDisplayMsg(kind, data)
+    JSONRPC.send_notification(conn_endpoint, "display", Dict{String,String}("kind"=>kind, "data"=>data))
```
ZacLN, line 34: Is a more specific notification name needed?

src/repl.ts
```diff
@@ -559,62 +570,40 @@ function processMsg(cmd, payload) {
...
-                let msg_len = parseInt(msg_len_as_string);
-                if (accumulatingBuffer.length >= cmd.length + msg_len_as_string.length + 2 + msg_len) {
-                    let payload = s.substring(index_of_sep_2 + 1);
-                    if (accumulatingBuffer.length > cmd.length + msg_len_as_string.length + 2 + msg_len) {
-                        accumulatingBuffer = Buffer.from(accumulatingBuffer.slice(cmd.length + msg_len_as_string.length + 2 + msg_len + 1));
-                    }
-                    else {
-                        accumulatingBuffer = Buffer.alloc(0);
-                    }
+const notifyTypeDisplay = new rpc.NotificationType<{kind: string, data: any}, void>('display');
```
ZacLN, line 578: As above?

davidanthoff:
I think `display` is fine for now, it really is the equivalent of `Base.display`. But we can also revisit later and change.

> is there a reason we don't pass NUM_THREADS here?

That looks like a bug/oversight to me, I think we should pass it there as well.

PR #1211, opened 2020-05-15
highlight current prospective code eval block

ZacLN:


davidanthoff:
That is pretty cool!

We should probably combine this LS request with the one for the active module, and then have that one request return both the extent and the module, right? Otherwise we'll have two requests for each cursor move.

How do we handle code cells? I think I'd be ok to deemphasis code cells and hope that most folks will use notebooks once they ship?

I would make the decoration more subtle, I think: no bold for the font, and for light themes maybe a very light gray background color, and not sure what to use for dark themes?

This feature _does_ really expose the long startup time of the LS in a much more "in your face" way than previously.... Not a reason to not merge, but solving the startup time on the LS side I think becomes much more urgent with this.

I wonder whether we should combine this with a codelens on top of the block that has a command "Run code" in it?

ZacLN:
Yep, we should definitely have a single `julia/getEvalInfoAt` command. 

I'm equally uninterested in the code cells as a workflow, but if there's demand I don't see why we couldn't have an option to highlight them instead.

I've not really explored the formatting much. Is it possible there is a mechanism to use some preset 'highlight' colour format set by themes themselves, similar to how duplicates of the current selection are highlighted throughout a doc? That'd probably be the best way forward.

Setting aside our start up time issues - I quite like that this gives a clear indication that everything is setup and ready to go.

davidanthoff:
> I quite like that this gives a clear indication that everything is setup and ready to go.

Haha, yes, that is a very useful side-effect :) We should probably add a status bar item in the short term that says "Starting Julia language server..." and that disappears once the LS is ready to process messages?

pfitzseb:
We could potentially re-use Juno's block detection code to do this before the LS is ready, but that's probably not worth the additional complexity (because it's only needed for the first ~20s or whatever).

Re the UI/UX side: Juno subtly highlights the current code block when you execute code and has a command to select the current code block. We could probably add those without hiding them behind an option because they should be universally useful and not very intrusive.

ZacLN:
Refactored to use [theme colours](https://code.visualstudio.com/api/references/theme-color) ("editor.inactiveSelectionBackground"). 

I'm not so keen on having the codeLens button (it seems a bit obtrusive to have that on every top-level block).

> Re the UI/UX side: Juno subtly highlights the current code block when you execute code and has a command to select the current code block. We could probably add those without hiding them behind an option because they should be universally useful and not very intrusive.

The eval'ed indicator sounds nice. Whats the purpose of the command to select the current block?

pfitzseb:
> Whats the purpose of the command to select the current block?

Dunno, but someone asked for it. So not very important I suppose ;)

package.json
```diff
@@ -467,6 +467,11 @@
                     "default": null,
                     "description": "Number of threads to use for Julia processes."
                 },
+                "julia.highlightCurrentBlock": {
+                    "type": "boolean",
+                    "default": true,
```
aviatesk, line 472: when I implemented "highlight current code block" feature in Juno, I found it a bit noisy to me imho (so I didn't push that anywhere remote). so I guess it might be better if this defaults to `false` but not sure. ppl may love this.

ZacLN: Yeah, I think I'd turn it off pretty quickly myself. I'm slightly wary of implementing new features that are disabled from the off as I think they're then likely to be largely ignored. It would be good to get a bit of a consensus view on this (and that approach more broadly)

davidanthoff:
I'm undecided re my preferences on this ;) On one hand it is cool, but I'm also a bit worried to add UI elements that deviate from what users get in the standard VS Code editor experience for other languages/files... 

So we could put it behind a default off configuration, but I also think @ZacLN's point that then no one will use it is valid...

So, I don't really know. A conservative approach might be that if we are not really 100% sold by a feature, we hold off from adding it? But I'm happy to be overruled entirely on this issue.

pfitzseb:
In almost all cases the current code block is easy enough to recognize from indentation etc, so I don't it's very useful to show it all the time.

I do quite like the short flash of eval'ed text that Juno has though, just as an indication that something's happening.

jkrumbiegel:
Just to add to this, my motivation for the highlight is that sometimes I think I’m executing a certain code block (by indentation e.g.) but then much more code is executed because there was something wrong with the code and the block detection included a lot more stuff than I wanted. Sometimes this is just annoying but at other times it has messed up the whole session for me. Basically it’s weird that the user has to trust that what they judge is the block is actually the block. Because that can easily go wrong

pfitzseb:
How about something subtle like
image
(although I really don't like how there's no margin between the line and the text)

davidanthoff:
I guess the question is whether we do something subtle by default on, or something more drastic (like a background color, but that could still be more a shading thing) that is off by default?

Issue #1221, opened 2020-05-19
Code execution commands


davidanthoff:
This is an attempt to come up with a more or less comprehensive list of commands we should have for code execution. I'm leaving out key bindings for now, I think that is a complicated topic we can tackle later, for now it would just be nice if we had a complete set of commands.

I think we have three different dimensions in this problem, and then we should offer one command for all possible combinations, taking out combinations that don't make sense (e.g. run whole file and move cursor).

1. WHAT to run: a) selected text, b) code block, c) code cell, d) everything above cursor position, e) whole file, e) auto mode (if text is selected => run that, if cells is defined => run cell, else => run code block)
2. CURSOR movement: a) don't move cursor, b) move cursor to next cell/block
3. DEBUG mode: a) no debugging, b) enter debugger, c) run debugger

Or something like that. Maybe we could first try to complete this list, i.e. add more items that come to mind. Then we could write down all permutations and check which ones make sense and whether we missed some aspect, and then we could start to implement this?

pfitzseb:
In Juno we currently have (also added in the current keybindings):
- Run Block (either runs the currently selected text or the block around the cursor) <kbd>Ctrl-Enter</kbd>
- Run Block and Move <kbd>Shift-Enter</kbd>
- Run Cell <kbd>Alt-Enter</kbd>
- Run Cell and Move <kbd>Alt-Shift-Enter</kbd>
- Run File <kbd>Ctrl-Shift-Enter</kbd>
- Debug Block
- Step Through Block
- Debug Cell
- Step Through Cell
- Debug File
- Step Through File

That's pretty close to what you listed above, but
- has only one ("auto mode") command for running the selected text or the current block
- doesn't have "run everything above cursor position"

IMHO having one command for selection/block/cell execution doesn't make too much sense because block- and cell-level execution have fairly distinct use cases.

One other somewhat related set of commands is for moving through blocks/cells:
- Next Cell <kbd>Alt-Down</kbd>
- Previous Cell <kbd>Alt-Up</kbd>
- Next Block (not in Juno)
- Previous Block (not in Juno)

pfitzseb:
`Ctrl-Shift-J` is a possible keybinding for showing the "Julia Explorer" pane (which currently only has the workspace).

pfitzseb:
Since we wanted to re-use the notebook keybindings here:
[...]

PR #1222, opened 2020-05-19
WIP: Inline Results

pfitzseb:
Peek 2020-05-19 11-19


phlavenk:
Would it be possible for  longer running calculations to have a "in_progress" animation before result is available and draws. by the way, I think the progress visualization is even more valuable than showing inline the value itself. I know Juno works heavily in the inline-results visual mode and a lot of clutter, unstable screen redraws and slowdown comes from the instant inline result output.
I'm definitely for having this available, and switchable using settings config.

pfitzseb:
Yeah, sure, that already works:
Peek 2020-05-20 15-36

pfitzseb:
image

aviatesk:
fantastic, will try to give a review on this

aviatesk:
Maybe code reorganization might be better to be done in a separate PR, I guess ?
Currently diffs in repl.ts doesn't seem to be obvious to me.

---

I've tested around this, and found there is an issue of evaluating code before we establish the connection with REPL will make all the `execute-hoge` commands un-functional somehow. I've not dig into that yet though.

pfitzseb:
> Maybe code reorganization might be better to be done in a separate PR, I guess ?
Currently diffs in repl.ts doesn't seem to be obvious to me.

Yeah, one thing led to another... :) I'll try to separate this into multiple PRs.

> I've tested around this, and found there is an issue of evaluating code before we establish the connection with REPL will make all the execute-hoge commands un-functional somehow. I've not dig into that yet though.

That's to be expected. We need the LS running to figure out the current code block (and module, once that's in).

pfitzseb:
image

pfitzseb:
Now with basic in-module evaluation:
image

src/interactive/results.ts
```diff
@@ -0,0 +1,170 @@
...
+
+        const color = new vscode.ThemeColor(content.isError ? 'editorError.foreground' : 'editor.foreground')
+
+        let decoration = {
+            after: {
+                contentIconPath: undefined,
+                contentText: undefined,
+                backgroundColor: new vscode.ThemeColor('editorWidget.background'),
+                margin: '0 0 0 10px',
+                color: color
```
aviatesk, line 46: It seems like the popular extension code lens's inline view has the precedence over:
image

I tried `textDecoration: 'z-index: some_high_number'` here, but without any succeed.

src/interactive/repl.ts
```diff
@@ -0,0 +1,414 @@
...
+    await startREPL(true);
+
+    let r = results.addResult(editor, range, {
+        content: ' ⟳ ', 
+        isIcon: false, 
+        hoverContent: '',
+        isError: false
+    })
+
+    let result: any = await g_connection.sendRequest(
```
aviatesk, line 336: > That's to be expected. We need the LS running to figure out the current code block (and module, once that's in).

hm, the problem I faced often with this branch was like, even after the LS started to run, sometimes the request here gets kinda stuck and after that every inline result won't work (only shows `⟳`.)
This happens if when use some inline execution command just after REPL process started, so I guess there maybe some problem around establishing a connection with REPL.

pfitzseb:
This is now on top of https://github.com/julia-vscode/julia-vscode/pull/1231 to make review easier.

src/interactive/results.ts
pfitzseb: I don't think we can really do anything about this. https://github.com/Microsoft/vscode/issues/18666 means there can only ever be one decoration with `after` text for a given range. 

We could switch to an empty range at the end of the line and use `before` instead, but that makes the internals here much more complicated because we now need to keep track of two ranges.

src/interactive/repl.ts
pfitzseb: Hm, I haven't seen this issue before. Might be fixed now?

pfitzseb:
Ok, so to summarize what this does in it's current form:

- All code evaluation now happens in a non-REPL thread.
- I've added inline results, where the first 100 chars of the first line are displayed inline and the rest in a hover pop-up. We can also add buttons to display the result in the REPL/workspace pane. Also added keybindings and commands for clearing the current inline result (<kbd>Escape</kbd>) and all inline results in an editor (<kbd>Ctrl+I Ctrl+C</kbd>).
- The backend supports different display settings (show inline, show in REPL, print executed code in REPL). Ref https://github.com/julia-vscode/julia-vscode/issues/1228.
- Basic in-module evaluation is implemented, but there's no UI to show the current module (or change away from the automatically detected one).
- Reworked the set of commands/keybindings for code execution (ref https://github.com/julia-vscode/julia-vscode/issues/1221):
  - Run Block (either runs the currently selected text or the block around the cursor) <kbd>Ctrl-Enter</kbd>
  - Run Block and Move <kbd>Shift-Enter</kbd>
  - Run Cell <kbd>Alt-Enter</kbd>
  - Run Cell and Move <kbd>Alt-Shift-Enter</kbd>
  - Run File <kbd>Ctrl-Shift-Enter</kbd>

davidanthoff:
This is awesome!! I haven't looked at the code itself (but will once this is ready for review).

> All code evaluation now happens in a non-REPL thread

That is no different than previously, right?

pfitzseb:
> That is no different than previously, right?

The current release still has the paste-to-REPL code execution mode, no?

davidanthoff:
> The current release still has the paste-to-REPL code execution mode

Ah, yes, but I assume we just leave that alone for now, right? I'm all in favor of getting rid of that command once we have solved the startup latency of the LS, but until then I think we need it.

src/interactive/results.ts
aviatesk: makes sense.

>  use `before` instead,

do you mean use `before` here ?:
https://github.com/julia-vscode/julia-vscode/blob/b57c94d4ff5fca5b9dd0f389575cd37471c881cb/src/interactive/results.ts#L41
(it looked really un-intuitively though...)

pfitzseb: Yes.

scripts/terminalserver/terminalserver.jl
```diff
@@ -34,6 +34,83 @@ function sendDisplayMsg(kind, data)
...
+            if resolved isa Module
+                out = resolved
+            else
+                return out
+            end
+        end
+    end
+
+    return out
+end
```
aviatesk, line 114: looks valid

scripts/terminalserver/terminalserver.jl
```diff
@@ -34,6 +34,83 @@ function sendDisplayMsg(kind, data)
...
```
aviatesk, line 113: looks valid

src/interactive/repl.ts
aviatesk: Yeah stuck on the initialization seems to have been solve somehow ... 🤔 Let's resolve this for now.

aviatesk:
@pfitzseb how about rebasing against master ?

package.json
```diff
@@ -188,13 +192,29 @@
...
+                "command": "language-julia.removeAllInlineResults",
+                "title": "Julia: Remove All Inline Results"
+            },
+            {
+                "command": "language-julia.removeAllInlineResultsInEditor",
+                "title": "Julia: Remove Inline Results In Editor"
+            },
+            {
+                "command": "language-julia.removeCurrentInlineResult",
+                "title": "Julia: Remove Current Inline Result"
```
aviatesk, line 217: nitpick, how about renaming these commands so that we use "clear" instead of "remove" ?
ink exports these functionalities under the convention so it may sound more natural for Juno users.

pfitzseb:
Rebased against https://github.com/julia-vscode/julia-vscode/pull/1232 now.

pfitzseb:
Ok, I think this should be ready for review.

Current issues that I'd like to address in follow-up PRs -- this is big enough already:
- Julia code should be tested somewhere like @aviatesk mentions above. Might make sense to pull it into a separate repo at some point maybe?
- Code execution doesn't play nice with REPL. 
  - `readline` doesn't work.
  - Interrupts probably don't work (and definitely not consistently).
  - Plots etc (everything that should hit a different display) aren't displayed unless the `Result in REPL` option is enabled.

Parts of this that should probably be discussed before merging:
- Default settings.
- Keybindings.
- Is it ok to remove the copy-paste-into-REPL code path (which I've done here). @davidanthoff argues against that above, but imho it doesn't warrant the additional complexity.

pfitzseb:
Oh, this also (unfortunately) contains https://github.com/julia-vscode/julia-vscode/pull/1232 now.

pfitzseb:
> Another (small) problem I've noticed is that it seems that deadlocks can happen when we execute code consecutively in a very instant interval. My hunch is that the "pre-communication" with the language client can get confused in that case, but I've not dug into.

I can't repro this:
Peek 2020-05-26 13-27

aviatesk:
> I can't repro this:

oh god, this has been caused by https://github.com/julia-vscode/LanguageServer.jl/issues/716; I've been trying this feature on a script which unfortunatelly happened to contain `readline` in the middle of the file... 
I think we can do `readline` woes later, so nvm

aviatesk:
> Julia code should be tested somewhere like @aviatesk mentions above. Might make sense to pull it into a separate repo at some point maybe?

For the meantime, I guess we can package up terminalserver code and wrap those connection stuff into a function, which will be called according to flags in `ARGS` or something like that.
Well, I don't care if we won't add tests on this PR.

aviatesk:
fwiw, rebasing against master again and `git submodule update` solved [this issue](https://github.com/julia-vscode/LanguageServer.jl/issues/713 ).

package.json
```diff
@@ -526,6 +566,24 @@
...
+                    "default": false,
+                    "description": "Show inline result bubbles.",
+                    "scope": "window"
+                },
+                "julia.execution.resultInREPL": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Show inline execution results in REPL.",
+                    "scope": "window"
+                },
```
aviatesk, line 581: we may want to "unify" this, maybe ? I mean, if `inlineResutls` are `false`, we show results in REPL, otherwise show inline bubbles (as in Juno).
I happened to turn off both configurations and I a bit confused by that any result didn't pop up anywhere (well, silly me)

pfitzseb: Dunno, people might want both?

aviatesk: okay, how about #1250 ?

pfitzseb: I like it.

davidanthoff:
@pfitzseb is this done from your end and ready for review? That is how I'm interpreting the status right now, just want to make sure :)

pfitzseb:
[Yep](https://github.com/julia-vscode/julia-vscode/pull/1222#issuecomment-633961179).

src/interactive/repl.ts
```diff
@@ -272,7 +251,7 @@ async function selectJuliaBlock() {
     }
 }
 
-async function executeJuliaCellInRepl() {
+async function executeJuliaCellInRepl(shouldMove: boolean = false) {
```
aviatesk, line 254: so currently `shouldMove` parameter doesn't seem to be respected anyway, right ?
I think currently this function will always behave as `run-cell-and-move`.

pfitzseb: Good point. Fixed.

aviatesk: confirmed ;)

aviatesk:
> this is big enough already:

aha, yeah, I will stop adding more small patches into this PR. Let's review as is and think whether we can merge this or need to fix something in bigger aspect.

aviatesk:
> Is it ok to remove the copy-paste-into-REPL code path (which I've done here).

Agree.

aviatesk:
I will merge `master` into this branch and resolve conflicts by #1251 and #1246

src/interactive/repl.ts
```diff
@@ -381,13 +398,19 @@ export function activate(context: vscode.ExtensionContext, settings: settings.IS
...
     context.subscriptions.push(vscode.commands.registerCommand('language-julia.startREPL', startREPLCommand));
 
-    context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaCodeInREPL', executeSelection));
+    context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaCodeInREPL', evaluateBlockOrSelection));
+    context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaCodeInREPLAndMove', () => evaluateBlockOrSelection(true)));
+
+    context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaCellInREPL', executeJuliaCellInRepl));
+    context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaCellInREPLAndMove', () => executeJuliaCellInRepl(true)));
+
+    // context.subscriptions.push(vscode.commands.registerCommand('language-julia.executeJuliaCodeInREPL', executeSelection));
```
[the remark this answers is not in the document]
aviatesk: That's because [`evaluateBlockOrSelection`](
https://github.com/julia-vscode/julia-vscode/blob/3ae906bffe633e39daf2b65e1081836196578cc0/src/interactive/repl.ts#L297-L329) will automatically handles both cases with or without selections.
It's better to not require them to use two different commands (and so probably different keybindings), no ?

package.json
```diff
@@ -250,21 +270,41 @@
             ]
         },
         "keybindings": [
+            {
+                "command": "language-julia.executeJuliaFileInREPL",
+                "key": "ctrl+shift+Enter",
```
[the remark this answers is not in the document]
pfitzseb: Sure. The same is true for `Ctrl-Enter` and `Shift-Enter` though. This of course ties into https://github.com/julia-vscode/julia-vscode/issues/1221 and we need to figure out a good keybinding story.

package.json
```diff
@@ -250,21 +270,41 @@
...
+                "when": "editorTextFocus && editorLangId == julia"
+            },
+            {
+                "command": "language-julia.clearAllInlineResultsInEditor",
+                "key": "ctrl+I ctrl+C",
+                "when": "editorTextFocus && editorLangId == julia"
+            },
+            {
+                "command": "language-julia.executeJuliaCellInREPL",
                 "key": "alt+Enter",
```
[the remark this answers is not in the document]
pfitzseb: `Ctrl-Enter` evaluates the current block (or selection, if one exists).

scripts/terminalserver/terminalserver.jl
```diff
@@ -50,37 +148,72 @@ run(conn_endpoint)
             code_line = params["line"]
             code_column = params["column"]
             source_code = params["code"]
+            mod = params["module"]
+
+            resolved_mod = try
+                module_from_string(mod)
+            catch err
+                @error err
```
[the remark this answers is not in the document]
pfitzseb: Yeah, I think this is just a remnant from debugging. Should probably just silently fall back to `Main` here.

scripts/terminalserver/terminalserver.jl
```diff
@@ -50,37 +148,72 @@ run(conn_endpoint)
...
+                            Base.display_error(stderr, res.err, res.bt)
+                        elseif res !== nothing && !ends_with_semicolon(source_code)
                             Base.invokelatest(display, res)
                         end
+                    else
+                        try
+                            Base.invokelatest(display, InlineDisplay(), res)
+                        catch err
+                            if !(err isa MethodError)
+                                @error("Display Error", ex=(err, catch_backtrace()))
```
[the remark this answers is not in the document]
pfitzseb: This is a proper bug, yes, but (usually) not with our code. We need to show this so people can fix their `show` methods. Would be better not to use `@error` though, agreed.

src/interactive/repl.ts
```diff
@@ -304,57 +283,95 @@ async function executeJuliaCellInRepl() {
...
+
+    if (resultType !== "REPL") {
+        const hoverString = '```\n' + result.all.toString() + '\n```'
+
+        r.setContent({
+            content: ' ' + result.inline.toString() + ' ',
+            isIcon: false,
+            hoverContent: hoverString,
+            isError: result.iserr
+        })
```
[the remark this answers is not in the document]
pfitzseb: You won't get any output (at least that's the intended behaviour).

davidanthoff:
Ah, and I missed some comments from above! Here are some reactions on those questions:

> Reworked the set of commands/keybindings for code execution

I think we should not do any changes to keybindings in this PR, but do that later. I think eventually we should make sure that we have consistent bindings between the notebook experience and what is happening in normal Julia files. But we don't know the notebook keybindings yet. I think what we should _not_ do is change keybindings now, and then change them again once we know the notebook keybindings, better to just do everything in one go in a couple of weeks.

> Progress bar integration would be cool to have.

The workspace PR has a notification for "eval starts" and "eval ends", that could be the first step towards that.

> Is it ok to remove the copy-paste-into-REPL code path

I feel pretty strongly that we need to keep this until we solved the startup latency of the LS. If we remove this now, then we end up in a situation where we don't have any keyboard shortcut that just works right away to execute Julia code if one starts up the extension, but one has to wait something like 10-15 seconds before execution works (until the LS is operational). I think that is too much of a regression for a way, way too common code path.

scripts/terminalserver/terminalserver.jl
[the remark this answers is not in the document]
davidanthoff: I think showing something would be actually quite good, I just think it shouldn't look like an error in our code :)

scripts/terminalserver/terminalserver.jl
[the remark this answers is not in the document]
davidanthoff: Ah, yes, you are right, so we should probably just do a `Base.display_error` here? Why are we special casing `MethodError`?

scripts/terminalserver/terminalserver.jl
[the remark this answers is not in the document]
pfitzseb: I'll just remove the printing for now.

Generally, is there some way to trigger error reporting from here? The user shouldn't ever need to see this error, but we may want to know about whether (and in what circumstances) this code errors.

src/interactive/modules.ts
```diff
@@ -0,0 +1,136 @@
+import * as vscode from 'vscode';
+import * as vslc from 'vscode-languageclient';
+import * as rpc from 'vscode-jsonrpc';
+import { onInit, onExit } from './repl'
+import { onSetLanguageClient } from '../extension';
+
+let statusBarItem: vscode.StatusBarItem = null
+let g_connection: rpc.MessageConnection = null
+let g_languageClient: vslc.LanguageClient = null
```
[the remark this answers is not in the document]
pfitzseb: I feel like explicitly `null`ing this is better because it's more explicit (and I think we're setting these to `null` e.g. when the connection is closed).

scripts/terminalserver/terminalserver.jl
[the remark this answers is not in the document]
pfitzseb: Because we're explicitly calling `display` on our `InlineDisplay`, which will throw a `MethodError` if it can't handle it's argument.

pfitzseb:
Ok, I think I have addressed all points raised above.

src/interactive/modules.ts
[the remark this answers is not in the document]
davidanthoff: Ok, lets leave for now, we should just come up with a consistent story across the code base at some point :)

PR #1232, opened 2020-05-22
Module Selector

pfitzseb:
Builds on #1231.

### DIFF
--- a/src/interactive/modules.ts
+++ b/src/interactive/modules.ts
@@ -3,7 +3,7 @@ import * as vslc from 'vscode-languageclient';
 import * as rpc from 'vscode-jsonrpc';
 
 let statusBarItem: vscode.StatusBarItem = null
-let g_connection: rpc.MessageConnection = undefined;
+let g_connection: rpc.MessageConnection = null
 let g_languageClient: vslc.LanguageClient = null
 
 interface TextDocumentPositionParams {
@@ -60,7 +60,7 @@ export function deactivate() {
 }
 
 async function updateStatusBarItem(editor: vscode.TextEditor) {
-    if (editor.document.languageId === 'julia') {
+    if (editor && editor.document && editor.document.languageId === 'julia') {
         statusBarItem.show()
         
         await updateModuleForEditor(editor)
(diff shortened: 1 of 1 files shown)


pfitzseb:
Ok, I think this is almost feature complete and adds
- A status bar item that displays the current module and whether it is actually loaded (not loaded when in parens):
image
- A module selector:
image

No automatic detection will happen for the current session and file if you set a module with the module selector (until you select `Choose Automatically).

This isn't at all hooked up to terminals yet. The current module is also not used anywhere, but https://github.com/julia-vscode/julia-vscode/pull/1222 will make use of it.

scripts/terminalserver/terminalserver.jl
```diff
@@ -35,6 +35,56 @@ function sendDisplayMsg(kind, data)
...
+    end
+
+    for m in Symbol.(ms)
+        if isdefined(out, m)
+            resolved = getfield(out, m)
+
+            if resolved isa Module
+                out = resolved
+            else
+                return out
```
aviatesk, line 61: maybe you want 
```suggestion
                return out !== Main
```
?

scripts/terminalserver/terminalserver.jl
```diff
@@ -35,6 +35,56 @@ function sendDisplayMsg(kind, data)
...
+            end
+        end
+    end
+
+    return out !== Main
+end
```
aviatesk, line 67: How about refactor this function like
```julia
is_module_loaded(mod) = mod == "Main" || module_from_string(mod) !== Main
```
?

aviatesk: nvm, I missed that this branch is on #1231

scripts/terminalserver/terminalserver.jl
pfitzseb: Yeah, good catch.

davidanthoff:
Can we delete this branch?

pfitzseb:
Unless someone wants to review this separately, sure.

davidanthoff:
It actually doesn't show any diffs in this PR anymore, for some reason :) I'll delete the branch and we can review as part of the bigger PR.

PR #1241, opened 2020-05-25
Add a status bar item for LS startup

davidanthoff:
I'm not actually sure this is a great idea, would like to hear some feedback.

This adds a status bar item that says "Starting Julia Language Server..." that will disappear as soon as the LS is read to accept requests, for example requests to identify the current block for say the Alt+Enter command.

It does add busy UI, but on the other hand it might be useful until we fix the LS startup performance problems?

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -25,11 +25,16 @@ import { JuliaDebugSession } from './juliaDebug';
 let g_settings: settings.ISettings = null;
 let g_languageClient: LanguageClient = null;
 let g_context: vscode.ExtensionContext = null;
+let g_lsStartup: vscode.StatusBarItem = null;
 
 export async function activate(context: vscode.ExtensionContext) {
     await telemetry.init(context);
 
     telemetry.traceEvent('activate');
+
+    g_lsStartup = vscode.window.createStatusBarItem();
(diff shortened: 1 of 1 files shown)


pfitzseb:
So we have the environment selector status bar item and may [soon have a module selector](https://github.com/julia-vscode/julia-vscode/pull/1232) item. 
What do you think about re-using those until their respective backend (LS/REPL) starts up?

davidanthoff:
The problem is that neither is just tied to the LS, the env selector status bar also controls the REPL startup env...

ZacLN:
I thought the module selector was dependent on the LS or is that no longer the case?

pfitzseb:
It is, yeah. But it's only relevant once the REPL has started.

Maybe a better solution would be to incorporate the startup progress into the future progress indicator or something like that?

davidanthoff:
In my mind the medium term goal here actually must be that we get the startup time down so much that we don't need this at all :) I.e. statically compiling the LS.

Maybe we just go with this PR here until then? It is not ideal, but on the other hand it is super simple, and maybe as a stop-gate ok?

pfitzseb:
Yeah, let's go with this as the short term solution.

davidanthoff:
Some needs to approve it, then I'll merge.

PR #1246, opened 2020-05-27
basic connection emission/listening

aviatesk:
build on top of #1222

### DIFF
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -16,7 +16,6 @@ import * as results from './results'
 import * as plots from './plots'
 import * as workspace from './workspace'
 import * as modules from './modules'
-import { DocumentColorRequest } from 'vscode-languageclient';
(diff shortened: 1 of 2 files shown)


aviatesk:
@pfitzseb what do you think on this ? This isn't necessary for your PR to be completed, while I think something like this will be useful as we implement more dynamic features.

src/interactive/repl.ts
```diff
@@ -16,6 +16,7 @@ import * as results from './results'
 import * as plots from './plots'
 import * as workspace from './workspace'
 import * as modules from './modules'
+import { DocumentColorRequest } from 'vscode-languageclient';
```
pfitzseb, line 19: This import is unnecessary, right?

aviatesk: ugh, vscode was so smart ...

PR #1249, opened 2020-05-28
refactor: listen for onSetLanguageClient, onDidChangeConfig event

aviatesk:
- make each sub file listen for those events
- now we don't need to modify extensions.ts when we add another feature
- ~~... and if we start to dispose things, those subscriptions will also just be cleared correctly~~ nvm, vscode will automatically cleans up subscription pushed into `context.subscriptions`, so these will be cleaned up correctly as is

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -45,7 +45,7 @@ export async function activate(context: vscode.ExtensionContext) {
     g_settings = settings.loadSettings();
 
     // Config change
-    context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(configChanged));
+    context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(changeConfig));
 
     // Language settings
     vscode.languages.setLanguageConfiguration('julia', {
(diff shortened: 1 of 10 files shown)


src/extension.ts
```diff
@@ -92,47 +92,27 @@ export async function activate(context: vscode.ExtensionContext) {
 }
 
 // this method is called when your extension is deactivated
-export function deactivate() {
-}
+export function deactivate() {}
 
+const _onSetLanguageClient = new vscode.EventEmitter<vslc.LanguageClient>()
+export const onSetLanguageClient = _onSetLanguageClient.event
```
davidanthoff, line 98: So far we have followed the convention to name any global variables `g_`. Probably best to follow that for now, just so that we don't have a mix of conventions. I think this would apply to the variables in the two lines here.

src/extension.ts
```diff
@@ -92,47 +92,27 @@ export async function activate(context: vscode.ExtensionContext) {
...
-    openpackagedirectory.onDidChangeConfiguration(newSettings);
-    jlpkgenv.onDidChangeConfiguration(newSettings);
-
-    let need_to_restart_server = false;
-
-    if (g_settings.juliaExePath != newSettings.juliaExePath) {
-        need_to_restart_server = true;
-    }
+const _onDidChangeConfig = new vscode.EventEmitter<settings.ISettings>()
+export const onDidChangeConfig = _onDidChangeConfig.event
```
davidanthoff, line 105: Same here, ideally start the names with `g_`.

src/interactive/repl.ts
```diff
@@ -14,6 +14,8 @@ import { Subject } from 'await-notify';
 
 import * as plots from './plots'
 import * as workspace from './workspace'
+import { DocumentColorRequest } from 'vscode-languageclient';
```
davidanthoff, line 17: This line is not needed, right?

src/smallcommands.ts
```diff
@@ -2,6 +2,8 @@ import * as vscode from 'vscode';
 import * as settings from './settings'
 import * as vslc from 'vscode-languageclient';
 import * as telemetry from './telemetry';
+import { onSetLanguageClient, onDidChangeConfig } from './extension';
+import { ContextTagKeys } from 'applicationinsights/out/Declarations/Contracts';
```
davidanthoff, line 6: This import is not needed, right?

src/extension.ts
aviatesk: okay, I just realize what `g_` stands for !
well, I'm okay to change `_onSetLanguageClient` to `g_onSetLanguageClient`, but imho changing `onSetLanguageClient` to `g_` prefixed sounds weird.
Most of event subscription APIs are defined under the jQuery's `on` naming convention, so I guess we're better to follow that and it deserves the exception of our convention ?

davidanthoff: Yes, agreed, that makes sense!

PR #1251, opened 2020-05-28
Formatting

davidanthoff:
This a) formats all TypeScript files and non-package Julia files and b) switches on as many auto-formatting settings I found, so that hopefully going forward we all end up formatting things routinely.

Main hope here is that we get less noisy PRs going forward, and have good formatting.

@aviatesk, did you use additional settings that we maybe should also use here?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,5 +8,11 @@
     },
     // we want to use the TS server from our node_modules folder to control its version
     "typescript.tsdk": "./node_modules/typescript/lib",
-    "git.detectSubmodulesLimit": 20 
+    "git.detectSubmodulesLimit": 20,
+    "editor.formatOnPaste": true,
+    "editor.formatOnSave": true,
+    "editor.formatOnType": true,
+    "files.trimTrailingWhitespace": true,
+    "files.trimFinalNewlines": true,
+    "files.insertFinalNewline": true,
 }
--- a/scripts/debugger/debugger.jl
+++ b/scripts/debugger/debugger.jl
@@ -51,7 +51,7 @@ function startdebug(pipename)
     conn = Sockets.connect(pipename)
     @debug "Connected to debug adapter."
     try
-        state = DebuggerState()     
+        state = DebuggerState()
 
         request_handlers = Dict{String,Function}()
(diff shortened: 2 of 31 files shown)


aviatesk:
nice,

> @aviatesk, did you use additional settings that we maybe should also use here?

I found I turned on
```json
"files.trimTrailingWhitespace": true,
"files.trimFinalNewlines": true,
"files.insertFinalNewline": true,
```
so maybe adding the last two as well will make me very happy.

As for TS code, [prettier](https://prettier.io/docs/en/index.html) produces simple and nice code, and it has [pre-commit hook format](https://prettier.io/docs/en/precommit.html) support, which will automatically format committed files rather instantly while synchronizing configurations; I'd used that for my other project (like [this](https://github.com/nteract/hydrogen)) and it worked pretty good.
So how about using that as well ? If you like to see how it feels, I'm glad to make an additional PR against this branch.

davidanthoff:
> adding the last two as well will make me very happy

Yes, added them!

> prettier produces simple and nice code

I have to admit I'm a bit weary of these opinionated formatters :) Also a bit worried about this commit pre hook stuff. Maybe we hold off from that for now and see how things go with the settings we have currently?

@julia-vscode/core I want to merge this quickly, because every time we now push something to `master` we end up with horrible merge conflicts with this PR, i.e. this kind of PR is painful to maintain if we don't merge it quickly. I don't think we need to do a careful review, all the changes here were made by tools that hopefully don't make changes that would affect semantics. I think I'm going to hold off merging anything else into `master` until we have merged this PR here, ok?

davidanthoff:
> Related, do I need to trigger the formatting bit for the JL packages- the changes seem to have disappeared

I think they are all there? They might look empty if you follow the latest notification to them (because they show the diff of the diff when the formatting PR was updated the last time), but if you look at the PR in itself they all look good to me?

aviatesk:
> I have to admit I'm a bit weary of these opinionated formatters :) Also a bit worried about this commit pre hook stuff. Maybe we hold off from that for now and see how things go with the settings we have currently?

okay, fair enough. And I really agree with your idea of merging this as soon as possible 👍

PR #1252, opened 2020-05-29
Add github issue notebooks

davidanthoff:
Not sure how useful this is, but maybe worth giving a try.

These notebooks only work in the VS Code Insiders version with the Github Issues Notebook extension installed.

They generally query all repos in this org here and I have prepopulated them with what I think are useful queries.


davidanthoff:
I'm just going to merge this, it seems really entirely risk free and would help me keep track of things.

PR #1253, opened 2020-05-29
Add Julia 1.5 support

davidanthoff:
For now I just want to trigger a CI build that includes Julia 1.5 to see whether that works or not.


davidanthoff:
I think we should merge now, just so that we are ready when 1.5 ships.

PR #1254, opened 2020-05-29
configurable cell separator

aviatesk:


pfitzseb:
See https://github.com/julia-vscode/julia-vscode/pull/777 for previous discussion on whether this should be configurable.

aviatesk:
agh, thanks for the pointer, I think I have to make a habit of searching "previous" things seriously...

pfitzseb:
FWIW, I would kinda like the default regex to change to `^##(?!#)`.

aviatesk:
yeah, I prefer that. Maybe post as an issue separate from #1222 ? We already have many things to be discussed or reviewed in that PR.

PR #1259, opened 2020-05-30
improve cell delimiter regex

### LINKED ISSUE #1256: discussion: change cell delimiter regex ?
Currently julia vscode extension uses `^##` as a cell delimiter regex, while from v0.12.5 we had modified Juno's default delimiter so that it won't recognize YAS style section headers as cells, and we ended up with `^##(?!#)` by default.
(if interested, see https://github.com/JunoLab/atom-julia-client/pull/721 or https://discourse.julialang.org/t/ann-juno-0-12-5/39744/2?u=aviatesk).
Maybe we also want to use this here as well ?

aviatesk:
- closes #1256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 0.16
+## 0.15.39
 * Improve cell delimiter regex so that it won't recognize YAS-style section headers as cell separator anymore (#1256, #1259)
 
 ## 0.15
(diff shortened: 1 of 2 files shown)


pfitzseb:
Should we mention this somewhere (e.g. CHANGELOG.md) since it's techincally breaking?

aviatesk:
ah, yes, I just remember we discussed that yesterday. Will do.

davidanthoff:
I wonder, though, whether we should just merge this now and ship with the next patch release? We can still mention it when we make a big announcement for 0.16, but this seems small and harmless enough that we can just slip it in?

aviatesk:
yeah, let's ship this in patch release.

PR #1262, opened 2020-05-31
introduce eslint

aviatesk:
In this PR I would like to introduce eslint for our code base.

I basically only want to introduce "minimum" rules; linting will make our code cleaner and less error prone while it shouldn't be annoying even with the future frequent code changes and not hinder our rapid development.
So all the rules introduced will automatically be fixed when an editor gets saved, except one rule `eqeqeq`; the `eqeqeq` rule will enforce us to use `===` instead of `==`, that is I believe really better habit when writing JS code (e.g. [this link](https://www.ecma-international.org/ecma-262/5.1/#sec-11.9.3) explains JS's weird comparison rule, e.g. by default `undefined == null` ...), and imho worth to be enforced.

---

In the second commit, I also introduced unused variable rule (which will enable us easily detect [this kind of problem](https://github.com/julia-vscode/julia-vscode/pull/1249#discussion_r431964771)) and cleaned up existing unused variable using `--noUnusedLocals` compiler(`tsc`) option. It revealed some code which should clearly be improved, but also banned seemingly intentional unused variables, like
```ts
let g_context: vscode.ExtensionContext = null;
let g_settings: settings.ISettings = null
```
at the top of file (I guess you wrote them maybe because they may be used in the future, or something).
So imho this rule is better to be introduced as a just "warning" rule rather than compiler option, but currently eslint can't handle TS's unused variables well yet, so `--noUnusedLocals` is our current only option. Under this rule, we can't compile the code if there is unused variables remaining and this can be really annoying I guess, so I'm totally okay to remove this option, even more, okay to revert the entire 2nd commit if you like to keep those intentional unused code.


davidanthoff:
Yes, I'm very much in favor of using eslint!

Is this PR the result of following the instructions in https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration? We should probably follow the config pattern described in there, and for example use the eslint extension that is mentioned there.

Finally, I'd say we should probably wait with merging this until we have gotten some of the larger pending PRs in, like the inline eval and workspace edit? Just to avoid merge conflicts?

davidanthoff:
PS: Oh, and I'm going to assign this to the next patch release, we don't need to wait for a minor release to merge something like this.

pfitzseb:
The only other thing I feel somewhat strongly about are semicolons -- I'd much prefer us never using them and therefore adding `"semi": "never"` to the eslint rules.

aviatesk:
> Is this PR the result of following the instructions in https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration? We should probably follow the config pattern described in there, and for example use the eslint extension that is mentioned there.

I basically followed the configurations in [vscode-eslint](https://github.com/microsoft/vscode-eslint), assuming who develops eslint integration makes best use of it :)
Well, as far as I read your link, the one thing this PR lacks is recommendation of extension.

If we agree with introducing this, I will revert the code base changes once and re-clean up against current master.

aviatesk:
> The only other thing I feel somewhat strongly about are semicolons -- I'd much prefer us never using them and therefore adding "semi": "never" to the eslint rules.

ah I remember that. I'm totally in favor of that; what do you think @davidanthoff ? 
I forgot the link that describes why we don't need to care them though. Maybe you know @pfitzseb ?

davidanthoff:
I'm on board with all of this!

@aviatesk do you want to clean up this PR so that we can merge it? I think we should merge this one next, before we merge any of the other pending PRs.

aviatesk:
yes of course. Let me do rebasing for this PR. I will first revert the changes.

aviatesk:
Done: couple of notes of changes:
- most changes are done automatically
- manual process:
  * fix `eqeqeq` error, i.e. `==` -> `===`
  * rm unused variables, running `npm run compile-dry-run`

aviatesk:
Azure failed because of this serious typo: `complile` -> `compile`.

aviatesk:
Summary of developer notes:
- most linting errors will be automatically fixed on save, e.g.: organizing imports, remove semicolons
- some "serious" errors needs to be fixed manually
  * `eqeqeq` rule will force us to use `===` and `!==` instead of `==` and `!=`
  * `noUnsedLocal` tsc rule will force us to remove unused variables: well, it actually compiles, and we even can webpack the code with them, but it's highly recommended to remove them before publishing code and CI will fail otherwise

We may want to keep this somewhere ? I guess we will want to concentrate our documentation into [this repository](https://github.com/julia-vscode/docs), right ?

aviatesk:
Okay I've done everything I wanted.

package.json
```diff
@@ -726,8 +726,10 @@
         "vscode:prepublish": "webpack --mode production",
         "webpack": "webpack --mode development",
         "webpack-dev": "webpack --mode development --watch",
-        "test-compile": "tsc -p ./",
```
davidanthoff, line 729: Is there a reason this is removed? I don't even know what it is for, but still would like to understand why we removed it :)

aviatesk: guessing from its name, it is supposed to test whether our TS code compiles or not, but it's exactly same as the another existing `compile` command.
Newly added `compile-dry-run` effectively supersedes that.

aviatesk:
> I did not review the individual ts files, under the assumption that those are all automatic eslint made changes that are fine :)

well, some of the changes aren't automatically done (see [this comment](https://github.com/julia-vscode/julia-vscode/pull/1262#issuecomment-639441572)) :P 
@davidanthoff could you tell me where we want to keep our developer documentation ?

davidanthoff:
I think there is a section for dev docs in docs repo. Or we could just continue to use the wiki for that? I’m indifferent :)

aviatesk:
Okay, I just wanted to check if there is something decided. I'm feeling like I prefer maintaining repository, so I will update its dev section maybe.

PR #1266, opened 2020-06-01
Use julia executable on the PATH environment variable for updateDeps npm task

cloudgiant:
The "hard-coded" julia executable path was wrong on my machine, which caused the `updateDeps` npm task to fail. I updated the code so that the julia executable path in the PATH environment variable is used instead to fix this.

### DIFF
--- a/src/scripts/updateDeps.ts
+++ b/src/scripts/updateDeps.ts
@@ -47,7 +47,8 @@ async function download_and_convert_grammar(juliaPath: string) {
 }
 
 async function main() {
-    let juliaPath = path.join(homedir(), "AppData", "Local", "Julia-1.3.1", "bin", "julia.exe");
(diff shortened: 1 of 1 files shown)


cloudgiant:
Makes sense. In that case, I don't see any reason to change this. I was just going through the different npm tasks to check if they all work while trying to figure out how to install the dependencies for a local run properly since just an `npm install` like on the vscode extension tutorials wasn't enough. It looks like I should have used `update2latest` instead :)

davidanthoff:
I think `npm i` and `git submodule init` and `git submodule update` should be enough. We should maybe have a npm task that just does that, but not the `git pull`?

cloudgiant:
Created a new PR for the new npm task :https://github.com/julia-vscode/julia-vscode/pull/1267

PR #1273, opened 2020-06-02
Remove a duplicate def of TextDocumentPositionParams

davidanthoff:
I don't think we need to define this ourselves, right?


pfitzseb:
Oh, I didn't realize this is exported from `vscode-languageclient`.

PR #1275, opened 2020-06-02
Update changelog

davidanthoff:
I would suggest we only differentiate by minor version, given how often we ship patch versions now, I think it is too cumbersome to list in which patch version exactly something shipped (also, would anyone care?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
 # Changelog
 
-## 0.15.39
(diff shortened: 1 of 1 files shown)


aviatesk:
> also, would anyone care?

no, let's just list minors :)

pfitzseb:
Although it would be nice to keep track of all changes somewhere -- maybe we could just collect them for each minor version?

aviatesk:
yeah, or use create milestones for each minor version and update CHANGELOG all at once when releasing the version.

I suggest just having a habit of always updating CHANGELOG for each PR (at least in this repo) would be best ?

PR #1276, opened 2020-06-03
More crashreporting

### LINKED ISSUE #1247: Add crash reporting to TypeScript parts
Right now we send crash reports for problems in the LS and in the debugger, but not at all for crashes in the TypeScript part of the extension itself.

Unfortunately there is no way right now to register a handler for all unhandled exceptions, see https://github.com/microsoft/vscode/issues/45264. So the only way to actually implement this right now is to wrap our code at _every_ entry point in a `try catch` block and send a crash report in the `catch` part and then rethrow the exception. This is super annoying, and we might easily miss some entry points, but on the other hand, even if we miss some, it would be an improvement in that we get more crash reports.

One thing that would make this a lot easier is if we could do the equivalent of the following Julia code in TypeScript:
```julia
function createCrashWrapped(f)
    return (i...) -> try
        f(i...)
    catch err
        # Put logic to send crash report here
        rethrow(err)
    end
end
```
We could then at least do things like:
```typescript
context.subscriptions.push(vscode.commands.registerCommand('language-julia.startREPL', createCrashWrapped(startREPLCommand)));
```
There would still be a lot of manual cases, though...

davidanthoff:
Fixes #1247.

On some level this is ridiculously bad and cumbersome, but as far as I can tell this kind of structure is the only way that we can add crash reporting to the TypeScript part of the extension... Essentially we need to make sure that every entry point into our code that can be called from VS Code is wrapped in a try catch... We could hope and wait for https://github.com/microsoft/vscode/issues/45264, but that hasn't seen much activity, and I'd really like to get crash reports for the TypeScript part of the extension rather sooner than later.

For now I have only done this for extension.ts. @julia-vscode/core any thoughts on this approach? Maybe some better ideas? If not, I would finish this by applying the same type of changes everywhere.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -21,6 +21,7 @@ import * as openpackagedirectory from './openpackagedirectory';
 import * as juliaexepath from './juliaexepath';
 import * as jlpkgenv from './jlpkgenv';
 import { JuliaDebugSession } from './juliaDebug';
+import { handleNewCrashReportFromError, wrapFuncWithCrashReporting, wrapAsyncFuncWithCrashReporting } from './telemetry';
 
 let g_settings: settings.ISettings = null;
 let g_languageClient: LanguageClient = null;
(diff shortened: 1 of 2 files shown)


davidanthoff:
@julia-vscode/core Bump, any thoughts on this?

I would love to add crash reporting for the whole extension well in advance of Juliacon, so that (hopefully) it allows us to fix more bugs before Juliacon.

Issue #1091, opened 2020-04-02
Debugger ignores Revised code or code edits 


phlavenk:
I'm excited by the usability of the debugger inside recetn vscode julia plugin.
One thing I use debugger for is to check where my bug in code is and i'd like tore-verify, after correcting, that the fix indeed did what i wanted.
In 0.15.18, the code changes get not propagated into the new debugging sessions:
```
function t1(a, b)
    if a>b
        return (a, b)
    else
        return  a #this is bug
    end
end

function t2(a,b)
    a, b = t1(a,b)
    return a, b
end
t2(4, 2) #works
t2(2, 4) #error
```
now, I run `@enter t2(2, 4)`, and step until i get to the bug - i'm expecting t1 returns a tuple. I'll correct t1 to:
```
function t1(a, b)
    if a>b
        return (a, b)
    else
        return  (b, a) #corrected bug
    end
end
``` 
and save the file and re-evaluate t1.  But, a next `@enter t2(2,4)` still errors during step-through, even when, on repl, `t2(2,4)` executes correctly.
I've seen this behavior in more complex situation, where debugger was ignoring `Revised` changes in a deved package, although the fixed code executed correctly.

davidanthoff:
Can you explain a bit more in detail how I can reproduce this bug? So far I haven't had any luck.

In particular: I assume you have one file that has the first code block in your example? How to you evaluate that file into the REPL? Ctrl+Enter, or Alt+Enter, or something else?

How do you re-evaluate t1?

I think the more details you provide, the better. Thanks!

phlavenk:
I'll try to give more detailed workflow:

Edit `vscode-mwe.jl` in vscode:
```julia
function t1(a, b)
    if a>b
        return (a, b)
    else
        return  a #this is bug
    end
end

function t2(a,b)
    a, b = t1(a,b)
    return a, b
end
```
save it.
Start REPL session:
```
julia> using Revise

julia> include("vscode_mwe.jl")
t2 (generic function with 1 method)
julia> Revise.track("vscode_mwe.jl")

julia> t2(4, 2)
(4, 2)

julia> t2(2, 4)
ERROR: BoundsError: attempt to access Int64
  at index [2]
Stacktrace:
 [1] indexed_iterate(::Int64, ::Int64, ::Nothing) at .\tuple.jl:90
[...]

PR #1283, opened 2020-06-03
Ship Revise

### LINKED ISSUE #1391: Revise error
After upgrading the julia-vscode to 0.16.7 I get errors from Revise (that is not part of the active project, bu the one injected by julia-vscode  itself). The error is identical to this one: timholy/Revise.jl#508 
They discuss the possible mismatch of `CodeTracking` versions. In my case:
```julia
(PhaseUnmixing) pkg> st -m Revise
ERROR: The following package names could not be resolved:
 * Revise (not found in project or manifest)

(PhaseUnmixing) pkg> st -m CodeTracking
Project PhaseUnmixing v1.0.5
Status `C:\Users\petr.hlavenka\.julia\dev\PhaseUnmixing\Manifest.toml`
  [da1fd8a2] CodeTracking v1.0.0
  [b77e0a4c] InteractiveUtils
  [cf7118a7] UUIDs
```
I'm not sure what is the problem here because I'm not familiar with the architecture of how Revise functionality is injected into the running julia session.

davidanthoff:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1091. Fixes #1391.

@timholy: this is roughly how I imagine we would ship Revise with the extension itself.

I think there are a few minor things we would need to work out, but overall I think this is pretty close.

This does require some code moving in Revise.jl, LoweredCodeUtils.jl and JuliaInterpreter.jl, see https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/401, https://github.com/JuliaDebug/LoweredCodeUtils.jl/pull/33 and https://github.com/timholy/Revise.jl/pull/501.

@julia-vscode/core: if we get this ready say by the end of the week, we might even include this in v0.16?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,7 +14,7 @@
     "typescript.tsdk": "./node_modules/typescript/lib",
     "git.detectSubmodulesLimit": 20,
     "editor.formatOnPaste": true,
-    "editor.formatOnSave": true,
+    "editor.formatOnSave": false,
     "editor.formatOnType": true,
     "files.trimTrailingWhitespace": true,
     "files.trimFinalNewlines": true,
--- a/package.json
+++ b/package.json
@@ -659,7 +659,7 @@
                 },
                 "julia.useRevise": {
                     "type": "boolean",
-                    "default": true,
+                    "default": false,
                     "description": "Load Revise.jl on startup of the REPL."
                 },
                 "julia.usePlotPane": {
--- a/scripts/packages/VSCodeServer/Project.toml
+++ b/scripts/packages/VSCodeServer/Project.toml
@@ -15,6 +15,9 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
+FileWatching = "7b1f6079-737a-58dc-b8bc-7a2ca5c1b5ee"
+Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
+LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
 
 [compat]
 julia = "1"
--- a/scripts/packages/VSCodeServer/src/VSCodeServer.jl
+++ b/scripts/packages/VSCodeServer/src/VSCodeServer.jl
@@ -23,6 +23,7 @@ end
 
 include("../../JSON/src/JSON.jl")
 include("../../CodeTracking/src/CodeTracking.jl")
+include("../../OrderedCollections/src/OrderedCollections.jl")
 
 module JSONRPC
     import ..JSON
@@ -37,6 +38,16 @@ module JuliaInterpreter
     include("../../JuliaInterpreter/src/packagedef.jl")
 end
 
+module LoweredCodeUtils
+    using ..JuliaInterpreter
+    using ..JuliaInterpreter: SSAValue, SlotNumber, Frame
+    using ..JuliaInterpreter: @lookup, moduleof, pc_expr, step_expr!, is_global_ref, whichtt,
+                        next_until!, finish_and_return!, nstatements, codelocation,
+                        is_return, lookup_return, is_GotoIfNot, is_ReturnNode
+
+    include("../../LoweredCodeUtils/src/packagedef.jl")
+end
+
 module DebugAdapter
     import ..JuliaInterpreter
     import ..JSON
@@ -46,6 +57,25 @@ module DebugAdapter
     include("../../DebugAdapter/src/packagedef.jl")
 end
 
+module Revise
+    using ..OrderedCollections
+    using ..CodeTracking
+    using ..JuliaInterpreter
+    using ..LoweredCodeUtils
+
+    using ..CodeTracking: PkgFiles, basedir, srcfiles, line_is_decl, basepath
+    using ..JuliaInterpreter: whichtt, is_doc_expr, step_expr!, finish_and_return!, get_return,
+                        @lookup, moduleof, scopeof, pc_expr, is_quotenode_egal,
+                        linetable, codelocs, LineTypes, is_GotoIfNot, isassign, isidentical
+    using ..LoweredCodeUtils: next_or_nothing!, trackedheads, structheads, callee_matches
+
+    include("../../Revise/src/packagedef.jl")
+end
+
+using .Revise:includet
+
+export includet
+
 module ChromeProfileFormat
     import ..JSON
     import Profile
@@ -54,6 +84,7 @@ module ChromeProfileFormat
 end
 
 const conn_endpoint = Ref{Union{Nothing,JSONRPC.JSONRPCEndpoint}}(nothing)
+const g_use_revise = Ref{Bool}(false)
 
 include("../../../error_handler.jl")
 include("repl_protocol.jl")
--- a/scripts/packages/VSCodeServer/src/eval.jl
+++ b/scripts/packages/VSCodeServer/src/eval.jl
@@ -114,11 +114,10 @@ function repl_runcode_request(conn, params::ReplRunCodeRequestParams)
 
         rendered_result = nothing
         f = () -> hideprompt() do
-            if isdefined(Main, :Revise) && isdefined(Main.Revise, :revise) && Main.Revise.revise isa Function
-                let mode = get(ENV, "JULIA_REVISE", "auto")
-                    mode == "auto" && Main.Revise.revise()
+            if g_use_revise[]
+                Revise.revise()
                 end
-            end
+
             if show_code
                 add_code_to_repl_history(source_code)
 
--- a/scripts/packages/VSCodeServer/src/repl.jl
+++ b/scripts/packages/VSCodeServer/src/repl.jl
@@ -87,6 +87,11 @@ end
 function evalrepl(m, line, repl, main_mode)
     return try
         JSONRPC.send_notification(conn_endpoint[], "repl/starteval", nothing)
+
+        if g_use_revise[]
+            Revise.revise()
+        end
+
         r = run_with_backend() do
             fix_displays()
             f = () -> repleval(m, line, REPL.repl_filename(repl, main_mode.hist))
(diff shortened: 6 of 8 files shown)


timholy:
So, the big challenge to merging this is https://github.com/timholy/Revise.jl/pull/497. I don't want code-movement to make a difficult rebase for major change.But realistically Revise 3.0 might be a month away (I'd like there to be a discourse-advertised beta period and there's still a bit I want to add to that PR first). I thought you were already in feature-freeze for 0.16? Ref https://github.com/julia-vscode/SymbolServer.jl/pull/160#issuecomment-637696370.

KristofferC:
Maybe I missed it, but I haven't seen it explicit. Why does the code layout matter for VSCode? And it feels a bit brittle as we update things that we don't follow the "blessed" way of structuring code such that it works with VSCode.

timholy:
https://github.com/julia-vscode/julia-vscode/issues/1091#issuecomment-636972226 describes the issues; basically they need to have a private copy.

I still wonder whether we might be better off having some kind of communication channel with the current runtime, rather than structure the packages to allow "forking." I guess one challenge, though, is "which runtime"? What if I have 5 Julia sessions open?

pfitzseb:
The extension has one REPL session that runs all user code. There's no communication channel between the LS and that REPL, but both of them can (and do) communicate with the editor.

IMHO we shouldn't rush these changes into 0.16 -- I'd like to take some time to think through the solution-space :)

davidanthoff:
> So, the big challenge to merging this is timholy/Revise.jl#497.

We actually don't need a tagged version of Revise.jl here, because we integrate things as a git submodule we can ship any branch we want. So one option would be that we just push the branch from the PR that I opened into the origin repo and leave it there, with no intention of ever merging it. For now the VS Code extension could ship that branch, and once you are done with Revise.jl 3.0 we just start shipping that branch (assuming that you make the same file reorganization on `master` for 3.0 as well).

> I thought you were already in feature-freeze for 0.16?

So we use a slightly weird process: I wanted to make sure we have a version shipped that is stable and not more buggy than what we do normally _before_ we started merging features for v0.16 into `master`. So we really had a feature freeze for the last version of the v0.15 line, because I wanted to make sure we shipped one last stable version from that series. But now that we are merging things for v0.16 we can do so, until we decide to ship v0.16 to the release channel, at which point new features will have to wait until we open `master` up for new feature merges for v0.17.

> Maybe I missed it, but I haven't seen it explicit. Why does the code layout matter for VSCode? And it feels a bit brittle as we update things that we don't follow the "blessed" way of structuring code such that it works with VSCode.

YES, this is certainly a hack. The (alternative) hack that I'm using right now for the debugger [here](https://github.com/julia-vscode/julia-vscode/blob/master/scripts/debugger/debugger.jl#L10) is probably even worse. I would certainly love some "official" way to load packages privately, but even if that was added to base Julia at some point, we would probably have to stay with our hacks for quite a while because we always want to support the last LTS release, which limits our ability to adopt new features in Julia base.

Having said that, I think it is a hack well worth it: I think in general the deployment story of the VS Code extension has been a big success. In all the years that we have shipped, we have never had a single user complaint about incompatible packages or anything like that, which is one of my favorite features of the extension :)

> IMHO we shouldn't rush these changes into 0.16 -- I'd like to take some time to think through the solution-space :)

I think we should probably actually treat this PR as a bug fix, which means we can merge and ship it at any time, i.e. also in some subsequent patch release, so no rush.

This PR really doesn't add new features: we already have the config setting for Revise, and this PR would essentially just make that work more reliable, and also fix the integration issues with the debugger. I am also a bit worried right now how the REPL hooking we are doing for the workspace stuff interacts with the hooking that Revise.jl does, and if we were to ship Revise.jl out of the box we could just entirely control that and make sure these don't interact in some weird way.

One thing I am getting more worried about, though (just to play devils advocate here) is that we are starting to load more and more stuff into our REPL process, and when we load packages in this private way, I think we don't make any use of precompile files. So we probably need to watch startup times quite carefully and at some point draw a line in terms of what we load out of the box.... For this Revise.jl story, we could probably add something that only loads the code if the config option is actually selected.

timholy:
So, let me say that I don't know enough to comment usefully. If you're confident this won't cause problems, won't get in the way of an even better solution sometime in the future, and doesn't require uglification of Revise, I'd have no objections to you shipping this in whatever form you think is appropriate. I can leave your branch(es) open in JI, LCU, and Revise for however long you wish.

For a more long-term solution I'm with @pfitzseb. It just seems likely that there are a number of subtle issues that deserve to be carefully evaluated for a period. Heck, as you say, startup-time issues alone drive a lot of the user experience but can be difficult to diagnose and fix. I'm pleased to say that Revise 3.0 seems to start up in a fraction of the time of Revise 2.0, but currently I don't deeply understand the reason for the improvement. It may all be invalidation-related, but I am not sure.

davidanthoff:
I thought more about this, and I think we should give the approach in this PR here a try. If it doesn't work, we can still explorer other approaches, and if we come up with something more elegant later, we can also try.

But right now we have at least two errors (https://github.com/julia-vscode/julia-vscode/issues/1360 and https://github.com/julia-vscode/julia-vscode/issues/1091) that would be fixed by the approach here, with no real alternative idea for a fix.

davidanthoff:
Alright, I think this is more or less ready for review, I think we just need to resolve the branch issue at https://github.com/timholy/Revise.jl/pull/501.

The general strategy is this now:
- We _always_ start the Julia REPL process with `JULIA_REVISE=manual`. That should hopefully disable any Revise.jl loading that is happening in a user startup.jl, correct @timholy?
- We _always_ load our own private copy of Revise.jl into `VSCodeServer`. If the user has configured the setting to use revise in VS Code, we briefly change the env var `JULIA_REVISE` to `auto` while we load `VSCodeServer` and our own copy of `Revise`. In that case we also call `VSCodeServer.Revise.async_steal_repl_backend()`. More on that below.
- As soon as we are done loading `VSCodeServer`, we set `JULIA_REVISE=manual` again, so that if the user types `using Revise` in the REPL, it will hopefully also not do anything.
- The general philosophy here is that we try to disable any user loaded Revise.jl as much as we can, and instead load our own private Revise.jl if the VS Code extension config setting for that is enabled.
- I've changed the default for the setting `ulia.useRevise` to false, that seems safer for now.

I think we should try to ship this in a patch release before v0.17 and see how it goes. If it causes problems, we can still take it out pre-Juliacon. But right now I don't have any other strategy to fix the integration bug with the debugger, and overall I'm actually quite happy with this PR here now, my hope is that it is actually quite robust.

timholy:
> if the user types using Revise in the REPL, it will hopefully also not do anything.

I'm not sure I fully understand. But probably that's because of my ignorance of vscode's organization. IIUC correctly there are two Julia sessions, one backing vscode (including its debugger) and the user's. I guess the question is whether vscode should always run Revise or run Revise only if the user is running it? Lots of people start it from their `startup.jl` file.

davidanthoff:
@timholy for the purpose of this PR, there is only one Julia process that is relevant: the Julia REPL that is started inside VS Code. With this PR, when you start the Julia REPL, it loads the following module structure at startup:
```
Main
    VSCodeServer
        JSON (p)
        CodeTracking (p)
        OrderedCollections (p)
        JSONRPC (p)
        JuliaInterpreter (p)
        LoweredCodeUtils (p)
        DebugAdapter (p)
        Revise (p)
```
All the packages that I marked with `(p)` are loaded in our "private" way of loading them: while we use the code in them, they are not loaded using the normal Julia package loading mechanism, but instead use the pattern [here](https://github.com/julia-vscode/julia-vscode/blob/47f07020b3882a7d944196752a704aa16b3fe066/scripts/packages/VSCodeServer/src/VSCodeServer.jl#L21). All the packages marked with `(p)` use each other, so for example `Main.VSCodeServer.Revise` uses `Main.VSCodeServer.CodeTracking`, and `Main.VSCodeServer.JuliaInterpreter` also uses `Main.VSCodeServer.CodeTracking`.

If the user at the REPL (or in their startup.jl) now types `using Revise`, then it gets loaded _again_, and we end up with the following module structure at runtime:
```
Main
    VSCodeServer
        JSON (p)
        CodeTracking (p)
        OrderedCollections (p)
        JSONRPC (p)
        JuliaInterpreter (p)
        LoweredCodeUtils (p)
        DebugAdapter (p)
        Revise (p)
    CodeTracking
    OrderedCollections
    JuliaInterpreter
    LoweredCodeUtils
    Revise
```
So now we have `CodeTracking` loaded twice. And in fact we might have loaded different versions at the same time: there is no reason that the CodeTracking we loaded privately has to be the same version that is loaded by the user via their `using Revise` command.

What I'm trying to do in this PR here is to make the `Main.Revise` not do _anything_ if we are in this VS Code REPL situation. Right now I'm doing that by setting the `JULIA_REVISE=manual` env variable at the startup of the Julia REPL process.

davidanthoff:
@julia-vscode/core we should have talked about this today, I forgot to bring it up!

Should we try this? I think we could even ship this in a patch release like now, and then if we run into big problems still remove before 7/1?

ZacLN:
Yep, on that basis I agree.

davidanthoff:
I need to create a new branch in Revise.jl off a newer version than what we have right now, there were some bugs fixed that are really crucial for this working.

@timholy, could you maybe give me push access to Revise.jl? I would only ever push to the `vscode-integration` and leave everything else alone. Also happy, though, to stick with the current workflow where I prepare a branch in my fork and you then push that to `origin`.

davidanthoff:
I'm running into too many problems right now. I don't think any of them are such that we can't solve them, but right now I think we should push this to after Juliacon. I've moved the PR to the backlog for now.

davidanthoff:
This is starting to work fairly well, I think. If some folks from the @julia-vscode/core team want to try it and look it over, I think it is in that kind of state now.

It does not yet fix https://github.com/julia-vscode/julia-vscode/issues/1091, and I don't really understand why not... Might need a debug session with @timholy to figure that one out :)

The other major question that is still open is how this will interact with any Revise.jl that the user might load via other methods, for example in a startup.jl or just manually.

Not for this PR, but I'm starting to wonder whether we should actually have an even closer integration. In particular, I have this scenario in mind: someone has Revise enabled in VS Code, then runs some function definition via Alt+Enter. Should we at that point start to track that function definition, even though the file it is in of course was never `includet`d?

timholy:
I'd be happy to try to figure out why it's not working with the debugger.

davidanthoff:
Closing as this is now happening in other PRs.

Issue #1279, opened 2020-06-03
[FR] Inline results hover improvements


marius311:
(From my comment on Slack regarding 0.16.0, split into a few issues)

* I'm guessing that hijacking the hover to display expdandable output is a necessary (and clever) workaround of VSCode's limitations with regards to showing output, but it does mean I have to enable hover, which I otherwise had off because I find it distracting. Is there any way to enable hover only for the inline results, but not for other stuff in my Julia files?

* When hovering to expand output, it seems like the the current line gets highlighted, which is slightly disorienting (if anything should get highlighted, I'd say it should be the inline result itself which you're expanding)

* Would be nice if the hover aligned to the inline result rather the beginning of the line, and perhaps if it was only ever below the output, although maybe VSCode handles that not you guys.

pfitzseb:
> I'm guessing that hijacking the hover to display expdandable output is a necessary (and clever) workaround of VSCode's limitations with regards to showing output, but it does mean I have to enable hover, which I otherwise had off because I find it distracting. Is there any way to enable hover only for the inline results, but not for other stuff in my Julia files?

No, I don't think there is.

> When hovering to expand output, it seems like the the current line gets highlighted, which is slightly disorienting (if anything should get highlighted, I'd say it should be the inline result itself which you're expanding)

Fixed with https://github.com/julia-vscode/julia-vscode/pull/1287.

> Would be nice if the hover aligned to the inline result rather the beginning of the line, and perhaps if it was only ever below the output, although maybe VSCode handles that not you guys.

Mostly fixed with https://github.com/julia-vscode/julia-vscode/pull/1287. VSCode collects all highlights on the current line (or maybe cursor position), so we can't guarantee positioning.

Issue #1282, opened 2020-06-03
[FR] Indicator of just-evaluated code


marius311:
(From my comment on Slack regarding 0.16.0, split into a few issues)

* I really like the thing Juno does where it flashes the evaluated code, so you get a visual confirmation that you evaluated just the part you thought you were evaluating. Is that possible here?

pfitzseb:
https://github.com/julia-vscode/julia-vscode/pull/1211 might be an alternative.

PR #1287, opened 2020-06-03
improve inline results

pfitzseb:
image
image

Fixes https://github.com/julia-vscode/julia-vscode/issues/1279 (the highlighting happens because of git lens in the first screenshot above).
Also makes our inline results compatible with git lens (as requested by @aviatesk).

### DIFF
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -336,6 +336,17 @@ async function evaluateBlockOrSelection(shouldMove: boolean = false) {
             editor.revealRange(new vscode.Range(nextBlock, nextBlock))
         }
 
+
+        const tempDecoration = vscode.window.createTextEditorDecorationType({
+            backgroundColor: new vscode.ThemeColor("editor.hoverHighlightBackground"),
+            isWholeLine: true
+        })
+        editor.setDecorations(tempDecoration, [range])
+
+        setTimeout(() => {
+            editor.setDecorations(tempDecoration, [])
+        }, 200)
+
         await evaluate(editor, range, text, module)
     }
 }
--- a/src/interactive/results.ts
+++ b/src/interactive/results.ts
@@ -11,7 +11,6 @@ export class Result {
     document: vscode.TextDocument
     text: string
     range: vscode.Range
-    decorationRange: vscode.Range
     content: ResultContent
     decoration: vscode.TextEditorDecorationType
     destroyed: boolean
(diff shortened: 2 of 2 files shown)


src/interactive/results.ts
```diff
@@ -60,18 +63,21 @@ export class Result {
...
                     hoverMessage: this.content.hoverContent,
-                    range: this.range
+                    range: this.decorationRange
                 }])
             }
         }
+    }
 
+    updateDecorationRange() {
+        this.decorationRange = new vscode.Range(this.range.end.translate(0, 9999), this.range.end.translate(0, 9999))
```
aviatesk, line 73: using getter will make code clearer imho:
```ts
get decorationRange(): vscode.Range 
    return new vscode.Range(this.range.end.translate(0, 9999), this.range.end.translate(0, 9999))
}
```

pfitzseb:
The last commit adds an indicator for the currently executed code (fixes #1282):
Peek 2020-06-03 12-57

src/interactive/results.ts
pfitzseb: Good call. Will make that change.

src/interactive/repl.ts
```diff
@@ -336,6 +336,17 @@ async function evaluateBlockOrSelection(shouldMove: boolean = false) {
...
+        const tempDecoration = vscode.window.createTextEditorDecorationType({
+            backgroundColor: new vscode.ThemeColor("editor.hoverHighlightBackground"),
+            isWholeLine: true
+        })
+        editor.setDecorations(tempDecoration, [range])
+
+        setTimeout(() => {
+            editor.setDecorations(tempDecoration, [])
+        }, 200)
+
```
aviatesk, line 349: before we have a progress metre, how about just clearing decoration after `await evaluate(editor, range, text, module)` rather than flashing ?
It may look kinda verbose if code runs long time, but it also indicates code is being executed.

Please discard this idea if you don't like this, I even not sure if I like that.

aviatesk: nvm, we've not established the way to interruption mechanism yet, so simply there won't be no way to clear the temporary decoration if code never terminates.
Introducing interruption should be done in separate PR. Let's keep this as flashing for now.

pfitzseb: Peek 2020-06-03 13-14

Not sure if I like it or not...

aviatesk: yeah, me neither, let's ignore this lol

aviatesk:
Feels valid, I think you can merge this as is @pfitzseb

PR #1288, opened 2020-06-03
follow up #1003

aviatesk:


davidanthoff:
Could we do this as a follow up PR against `master` _after_ the workspace PR is merged? I'd like to slow down the churn on the workspace PR, so that we can merge it rather sooner than later.

aviatesk:
sure, I guess this PR is ready for review/merge now ?

PR #1290, opened 2020-06-03
improve communication

aviatesk:
- add two different handlers, `rpc_handler` returns response, while `msg_handler` not
- simplify core communication handling point using those handlers

things I would like to do:
- [ ] maybe fix `rpc/msg` naming ? they may sound same
- [ ] split terminalserver.jl into multiple subfiles, say, communication.jl, display.jl, evaluation.jl, and misc.jl

---

If we want to merge this, other big PR with new communication methods like #1003 may be better to be merged in advance.

### DIFF
--- a/scripts/terminalserver/terminalserver.jl
+++ b/scripts/terminalserver/terminalserver.jl
@@ -157,98 +157,122 @@ function get_modules(toplevel = nothing, mods = Set(Module[]))
     mods
 end
 
-run(conn_endpoint)
+"""
+    handle_rpc(f::Function, meth::AbstractString)
 
-@async begin
+Sets `f` as handler for `meth`, whose return value will be sent to the frontend over JSON endpoint.
+"""
+handle_rpc(f::Function, meth::AbstractString) = rpc_handlers[meth] = f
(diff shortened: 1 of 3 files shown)


davidanthoff:
In general I'm in favor of something like this, _but_ we have something like this already in JSONRPC.jl, I just haven't come around to integrate it into the REPL code here, I wanted to wait with that until we have merged the inline eval and workspace PRs.

You can see a (rough) version of how this might look when integrated in the LS code [here](https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/languageserverinstance.jl#L268). I still want to add a few tweaks in JSONRPC.jl before we switch the REPL code over, though.

aviatesk:
ah, okay ! So I will wait until you craft the change. Maybe your PR will be enough and we will end up closing this.

aviatesk:
let's close this, code base changed a lot

davidanthoff:
I'll migrate the REPL code to the message dispatcher in JSONRPC once we have v0.16 out, for now I think we should hold off internal refactorings, just to let everything settle a bit.

PR #1296, opened 2020-06-04
improve render/strlimit etc

pfitzseb:
This makes two improvements to the Julia side of our rendering code (for inline results and the workspace):
- Removes ANSI control chars since we can't handle them.
- Improve performance for expensive (or even non-terminating) `show` methods by returning early once `limit` bytes have been written to the IO.

### DIFF
--- a/scripts/terminalserver/misc.jl
+++ b/scripts/terminalserver/misc.jl
@@ -1,0 +1,57 @@
+# https://github.com/JuliaDebug/Debugger.jl/blob/4cf99c662ab89da0fe7380c1e81461e2428e8b00/src/limitio.jl
+
+mutable struct LimitIO{IO_t <: IO} <: IO
+    io::IO_t
+    maxbytes::Int
+    n::Int
+end
+LimitIO(io::IO, maxbytes) = LimitIO(io, maxbytes, 0)
+
+struct LimitIOException <: Exception end
+
+function Base.write(io::LimitIO, v::UInt8)
+    io.n > io.maxbytes && throw(LimitIOException())
+    io.n += write(io.io, v)
+end
+
+function sprintlimited(args...; func = show, limit::Int = 30, ellipsis::AbstractString = "…", color = false)
+    io = IOBuffer()
+    ioctx = IOContext(LimitIO(io, limit - length(ellipsis)), :limit => true, :color => color, :displaysize => (30, 64))
+
+    try
+        Base.invokelatest(func, ioctx, args...)
+    catch err
+        if err isa LimitIOException
+            print(io, ellipsis)
+        else
+            rethrow(err)
+        end
+    end
+
+    str = filter(isvalid, String(take!(io)))
+
+    return color ? str : remove_ansi_control_chars(str)
+end
+
+function strlimit(str; limit::Int = 30, ellipsis::AbstractString = "…")
+    will_append = length(str) > limit
+
+    io = IOBuffer()
+    i = 1
+    for c in str
+        will_append && i > limit - length(ellipsis) && break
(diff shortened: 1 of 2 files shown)


scripts/terminalserver/misc.jl
```diff
@@ -0,0 +1,57 @@
...
+    n::Int
+end
+LimitIO(io::IO, maxbytes) = LimitIO(io, maxbytes, 0)
+
+struct LimitIOException <: Exception end
+
+function Base.write(io::LimitIO, v::UInt8)
+    io.n > io.maxbytes && throw(LimitIOException())
+    io.n += write(io.io, v)
+end
```
aviatesk, line 15: It seems impossible to me, but do you think can we respect "length" of strings instead of its bytes ?
If not, I guess `length(ellipsis)` inside of `sprintlimited` is better to be `ncodeunits(ellipsis)` ?
And maybe it makes sense to make `strlimit` behave consistently (i.e. respect bytes rather than length, maybe, or mention the difference if we keep them different)

scripts/terminalserver/misc.jl
```diff
@@ -0,0 +1,57 @@
...
+
+function sprintlimited(args...; func = show, limit::Int = 30, ellipsis::AbstractString = "…", color = false)
+    io = IOBuffer()
+    ioctx = IOContext(LimitIO(io, limit - length(ellipsis)), :limit => true, :color => color, :displaysize => (30, 64))
```
aviatesk, line 19: Could you explain why do we want to have `:color => true` case from the beginning ?

scripts/terminalserver/terminalserver.jl
```diff
@@ -90,10 +67,10 @@ the following fields:
 - `iserr`: Boolean. The frontend may style the UI differently depending on this value.
 """
 function render(x)
-    str = show_with_strlimit(x)
+    str = sprintlimited(MIME"text/plain"(), x, limit = 10_000)
 
     return Dict(
-        "inline" => strlimit(first(split(str, "\n")), 100),
+        "inline" => strlimit(first(split(str, "\n")), limit = 100),
```
aviatesk, line 73: how about extracting `100` and `10_000` into global constants ?

scripts/terminalserver/misc.jl
aviatesk: fwiw, here is our `strlimit` doc:
```julia
"""
    strlimit(str::AbstractString, limit::Integer = 30, ellipsis::AbstractString = "…")

Chops off `str` so that its _length_ doesn't exceed `limit`. The excessive part
  will be replaced by `ellipsis`.

!!! note
    The length of returned string will _never_ exceed `limit`.
"""
```

aviatesk:
nitpick: overloading `unsafe_write` seems to be more "official" approach for this ?
ref: https://docs.julialang.org/en/v1/base/io-network/#Base.unsafe_write

scripts/terminalserver/misc.jl
pfitzseb: The docstring still applies since `length(x) <= ncodeunits(x)` for all `x`. So yes, we might cut off too much and even produce invalid UTF8, but that's a price I'm willing to pay ;)

scripts/terminalserver/misc.jl
pfitzseb: Dunno, but it seems possible that we might be able to render color at some point.

scripts/terminalserver/misc.jl
aviatesk: aha, yeah true.
Don't know we need to be strict on the difference, so I'm going to approve this.

PR #1297, opened 2020-06-04
workspace module

aviatesk:
and bunch of improvements:
- show variables correctly (we need `all` option for modules other than `Main`)
- simplify TS code
- update workspace on module selection, REPL init

remaining TODOs:
- [x] add icon for workspace view to toggle workspace module selector
- [ ] fix `show` method -- it currently fails on `show`ing `Enum` and thus we can't get workspace for `Base`, for instance

### DIFF
--- a/package.json
+++ b/package.json
@@ -220,6 +220,10 @@
             {
                 "command": "language-julia.chooseModule",
                 "title": "Julia: Select Current Module"
+            },
+            {
+                "command": "language-julia.chooseWorkspaceModule",
+                "title": "Julia: Select Workspace Module"
             }
         ],
         "menus": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -618,16 +622,16 @@
         "viewsContainers": {
             "activitybar": [
                 {
-                    "id": "julia-explorer",
-                    "title": "Julia Explorer",
+                    "id": "julia-workspace",
+                    "title": "Julia Workspace",
                     "icon": "images/julia-dots.svg"
```
aviatesk, line 627: I tried to unify workspace/REPL variable/explorer terminology, but let me know if you don't like this.

pfitzseb:
Kinda, but that doesn't work for packages (because they are not necessarily reachable from `Main`).

davidanthoff:
If you load a package in the REPL, you can access it from Main, right?

aviatesk:
close in favor of #1302 . Small enhancements in this PR will be added afterwards.

PR #1302, opened 2020-06-05
Lazy Tree View in Workspace

pfitzseb:
This isn't the prettiest yet, but should be a minimal working implementation. I'll iterate over this during the next few hours.


pfitzseb:
Ok, this is the current state:

image

Should be in a pretty good state by now. There's still a bit of an inconsistency with the light text next to the symbol (or field) name: Sometimes it displays the type and sometimes it displays the value. Usually it's pretty clear though.

aviatesk:
nice, how do you feel on drilling into a module ?

pfitzseb:
image

Now has support for arbitrarily *big* arrays (and dicts).

pfitzseb:
> nice, how do you feel on drilling into a module ?

It's ok for now, imho.

scripts/terminalserver/trees.jl
```diff
@@ -0,0 +1,156 @@
...
+        for n in ns
+            isdefined(x, n) || continue
+            Base.isdeprecated(x, n) && continue
+            startswith(string(n), '#') && continue
+            push!(out, SubTree(string(n), getfield(x, n)))
+        end
+
+        out
+    end))
+end
```
aviatesk, line 95: maybe we want to filter out `x` module itself here ?

scripts/terminalserver/terminalserver.jl
```diff
@@ -16,26 +16,37 @@ module JSONRPC
...
     M = Main
     variables = []
+    clear_lazy()
+
     for n in names(M)
         !isdefined(M, n) && continue
         x = getfield(M, n)
-        x isa Module && continue
-        x==Main.vscodedisplay && continue
+        # x isa Module && continue
```
aviatesk, line 34: same here, filter `Main` here to remove duplication.

scripts/terminalserver/trees.jl
```diff
@@ -0,0 +1,156 @@
...
+
+function treerender(x::AbstractDict)
+    treerender(LazyTree(string(nameof(typeof(x)), "{$(eltype(x).parameters[1]), $(eltype(x).parameters[2])}"), function ()
+        if length(keys(x)) > 25
+            partition_by_keys(x)
+        else
+            [SubTree(repr(k), x[k]) for k in keys(x)]
+        end
+    end))
+end
```
aviatesk, line 80:
```julia
function treerender(x::D) where {D <: AbstractDict{K,V}} where {K,V}
    treerender(LazyTree(string(nameof(D), "{$K, $V}"), function ()
        if length(keys(x)) > 25
            partition_by_keys(x)
        else
            [SubTree(repr(k), v) for (k, v) in x]
        end
    end))
end
```
, maybe ?

scripts/terminalserver/trees.jl
pfitzseb: Good idea. Done.

scripts/terminalserver/trees.jl
```diff
@@ -0,0 +1,160 @@
...
+        if haskey(TREES, id)
+            x = [treerender(x) for x in pop!(TREES, id).children()]
+            return x
+        else
+            return ["[out of date result]"]
+        end
+    catch err
+        @error exception=(err, catch_backtrace())
+        return ["nope"]
+    end
```
aviatesk, line 149: Can this case ever happen ? Well, I think we should keep this anyway though.

aviatesk:
> This also seems to work well with the GC, afaict. We're deleting all references to lazy views on each evaluation.

This sounds most robust, and I guess it doesn't take much time since we only need to clear toplevel tree keys (which are usually not so many).
I looked through the code, and to say honestly it looks very cool to me. Great work @pfitzseb !

scripts/terminalserver/terminalserver.jl
pfitzseb: Done.

scripts/terminalserver/trees.jl
pfitzseb: Done.

scripts/terminalserver/trees.jl
pfitzseb: I don't think so, but let's see if someone complains about it ;)

pfitzseb:
Ok, I think I've addressed all of those comments.

scripts/terminalserver/trees.jl
aviatesk: good 👍

pfitzseb:
No, those sneaked (snuck?) in by accident. Removed them.

Let's also see what @davidanthoff says about this.

pfitzseb:
This now also has a heuristic for whether to show the "Open in VSCode" button or not.

pfitzseb:
Now also shows the size of arrays/dicts and hides the expander arrows when there are no child elements.

scripts/terminalserver/terminalserver.jl
```diff
@@ -358,32 +377,46 @@ Base.Multimedia.istextmime(::MIME{Symbol("application/vnd.dataresource+json")})
...
+        if showable(mime, x)
+            return true
+        end
+    end
+
+    if showable("application/vnd.dataresource+json", x)
+        return true
+    end
+
+    if _isiterabletable(x) === missing || _isiterabletable(x) === true || x isa AbstractVector || x isa AbstractMatrix
```
davidanthoff, line 405:
```suggestion
    istable = Base.invokelatest(_isiterabletable, x)

    if istable === missing || istable === true || x isa AbstractVector || x isa AbstractMatrix
```
I just got a world age error without this.

src/interactive/workspace.ts
```diff
@@ -17,28 +41,46 @@ export class REPLTreeDataProvider implements vscode.TreeDataProvider<WorkspaceVa
...
+                    }
+                    resolve(out)
+                })
+            })
         }
         else {
-            return g_replVariables
+            return Promise.resolve(g_replVariables)
         }
     }
```
davidanthoff, line 68: I think we can just write this in terms of `async`/`await`:
```typescript
    async getChildren(node?: WorkspaceVariable) {
        if (node) {
            const children = await g_connection.sendRequest(requestTypeGetLazy, node.id)

            const out: WorkspaceVariable[] = []

            for (const c of children) {
                out.push({
                    name: c.head,
                    type: '',
                    value: c.value,
                    id: c.id,
                    lazy: c.lazy,
                    haschildren: c.haschildren,
                    canshow: c.canshow
                })
            }

            return out
        }
        else {
            return g_replVariables
        }
    }
```

pfitzseb: I was so sure I tried this before and got a compiler error. But yes, this looks much better ¯\\_(ツ)\_/¯

pfitzseb:
> One question: when I tried this and executed using VegaDatasets, it didn't show VegaDatasets in the workspace. Should it, though?

It does now :)

pfitzseb:
Now with icons:
image

PR #1308, opened 2020-06-06
add documentation browser support

aviatesk:
- [x] retrieve docstrings in LS side
- [x] handle `@ref`s (very hacky though)
- [x] browsing back and forward
- [x] persistence
- [x] disposal handling
- [x] auto style switching
- ~[x] no connection handling~
- ~[ ] security stuff~ EDIT: no longer consideration
- [x] vendor scripts ~I couldn't make this work. The idea is to vendor cdn scripts inside documenter.js by ourselves~ EDIT: done thanks to @davidanthoff 
- [ ] handle links, again
- [ ] fix "Failed to load downloaded font ..." err, maybe ?

| dark | light |
|---|---|
| Screen Shot 2020-06-07 at 1 36 47 |  Screen Shot 2020-06-07 at 1 35 57 |


davidanthoff:
I haven't reviewed in detail, but I think in general we should try to not use the REPL process for this. At least right now I think we should be able to get all the info we need from the LS. One major benefit of that is that the documentation will work even if the user hasn't loaded any package into the REPL.

src/interactive/repl.ts
```diff
@@ -25,6 +26,18 @@ let g_terminal: vscode.Terminal = null
 
 export let g_connection: rpc.MessageConnection = undefined
 
+export function withConnection(what: string, callback: (connection: rpc.MessageConnection) => void) {
+    if (isConnectionActive()) {
+        callback(g_connection)
+    } else {
+        vscode.window.showInformationMessage(`${what} requires an active REPL.`)
+    }
+}
```
davidanthoff, line 35: Just one random comment here: I think general we probably need to use a different pattern for these kinds of things that takes into account that `g_connection` can easily get closed right after our call to `isConnectionActive` by the other end of the communication channel. As far as I can tell the only way to make these kinds of situations robust is to use `try catch` blocks.

We had a similar situation in the LS: for example we checked whether a file exists on disc, and then opened it. Turns out that with the huge user numbers we had we ran into situations where we checked for the existence of the file, then the file was deleted, and then we tried to open it. So we had to change all of those pieces of code to a pattern where we didn't even bother to check whether the file exists or not, but instead just catch errors if the `open` call fails because that covers both the case that the file didn't exist in the first place _and_ the case that the file gets deleted in the middle of our own code path.

aviatesk:
> I think in general we should try to not use the REPL process for this. At least right now I think we should be able to get all the info we need from the LS. One major benefit of that is that the documentation will work even if the user hasn't loaded any package into the REPL.

I totally I agree with that it's better to use LS for this, while, why I use the REPL process here is that LS doesn't seem to provide the information I want right now.
E.g.:
- if we hover on `sin`, it will only show docstrings in `LinearAlgebra`'s one and won't show `Base`'s
- we can't get docstrings on structs, variables
So in other word the REPL process seems to be more "stable" in the current state.
It's obviously better to improve LS of course, but as a first iteration, isn't acceptable to use the REPL ?

aviatesk:
I think this is ready for review -- the biggest discussion would be whether to use LS or not for this PR.

davidanthoff:
Yeah, I think all the issues you mention should be fixed in LS, @ZacLN can probably best say how involved that is.

I think the major downside of relying on the REPL here is that then a) you don't get anything if the REPL hasn't started and b) you don't get anything for packages that haven't been loaded in the REPL. To me those are really major downsides, in particular because we had this philosophy here from the beginning that the experience you get in the editor should not depend on the state of your REPL or anything like that.

pfitzseb:
Agreed. Let's fix try to fix the LS issues. If that's too hard we can think about a hybrid approach as well.

I'll try to review this tomorrow.

aviatesk:
> To me those are really major downsides, in particular because we had this philosophy here from the beginning that the experience you get in the editor should not depend on the state of your REPL or anything like that.

fair enough. Let's try to ship this feature with LS then. Not sure if I can write what I want in LS side though :p

src/interactive/repl.ts
aviatesk: maybe addressed by eb5ab9b ?

aviatesk:
Ok, now this PR uses LS for doc retrieve and needs https://github.com/julia-vscode/LanguageServer.jl/pull/744

src/interactive/repl.ts
davidanthoff: I think that is too broad, as the `catch` clause will now catch any exception in the `callback`, even exceptions that are not due to the connection being closed. I don't think this pattern here with where one passes one fairly large callback will work, I think instead we need a narrow try catch block around every call that sends a request or notification, with narrow error handling for that particular location.

davidanthoff:
I also converted this to a draft, as it is not ready to be merged, right?

I general, I would say this would be v0.17 material, right?

aviatesk:
> I also converted this to a draft, as it is not ready to be merged, right?

imho the frontend implementation (which is done in this PR) is ready for review (and so I turned this out of draft) and maybe ready for merge.

As I said, LS side implementation should be improved, but I don't think we need to hold this feature to get in until we get satisfied with LS's. That can be gradually improved, no ?
For the time being we even can do "hybrid" approach, that will use documentation obtained from REPL if REPL is active and a target module is loaded, as @pfitzseb suggested above.

pfitzseb:
> I general, I would say this would be v0.17 material, right?

src/extension.ts
```diff
@@ -83,20 +83,43 @@ export function deactivate() { }
...
-            g_languageClient.stop()
-            setLanguageClient()
-        }
+        withLanguageClient(
+            languageClient => {
+                languageClient.stop()
+                setLanguageClient()
+            },
+            err => { }
+        )
```
davidanthoff, line 122: Uh, I think we should stay away from this callback style, isn't the whole idea of promises, `async` and `await` to avoid that? Also, that seems an unrelated design change, if we were to introduce it, can we do it in a different PR?

src/interactive/documentation.ts
```diff
@@ -0,0 +1,228 @@
+import * as path from 'path'
+import * as vscode from 'vscode'
+import { withLanguageClient } from '../extension'
+import { getParamsAtPosition, setContext } from '../utils'
+
+const viewType = 'JuliaDocumentationBrowser'
+const panelActiveContextKey = 'juliaDocumentationPaneActive'
+let extensionPath: string = undefined
```
davidanthoff, line 8:
```suggestion
let extensionPath: string | undefined = undefined
```
Just in preparation for strict mode :)

src/interactive/repl.ts
```diff
@@ -25,6 +26,25 @@ let g_terminal: vscode.Terminal = null
...
+    if (!isConnectionActive()) {
+        return callbackOnHandledErr(new Error('REPL is not active'))
+    }
+    try {
+        return callback(g_connection)
+    } catch (err) {
+        // TODO: identify which error should be handled
+        throw err
+    }
+}
```
davidanthoff, line 46: Same here, I'm not a huge fan of these callback style APIs. But either way, if we do introduce that into the codebase, we should do it in a separate PR.

package.json
```diff
@@ -596,6 +640,11 @@
                     "default": false,
                     "description": "Print executed code in REPL.",
                     "scope": "window"
+                },
+                "julia.documentation.darkMode": {
+                    "type": "boolean",
+                    "default": true,
+                    "markdownDescription": "Use [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl)'s dark theme stylesheet for documentation pane, otherwise its light theme stylesheet will be used."
```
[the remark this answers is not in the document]
aviatesk: afaict no, look at [this comment](https://github.com/julia-vscode/julia-vscode/pull/1308#pullrequestreview-440156230):

> light/dark mode needs to be switched manually; this is because Documenter.jl's stylesheet is supposed to be applied to html itself rather than body, so we can't use this vscode webview style API

But I feel @pfitzseb may have an idea to solve this.

src/extension.ts
aviatesk: Yeah we should modify the existing part in another PR. I will revert back this code.

src/interactive/repl.ts
aviatesk: Actually `withREPL` is now not used anywhere in this PR, so I will remove this one.
Well, as for the callback API (i..e. `withLanguageClient` in this PR), I would like to use such a thing when we don't need to subscribe to frequent events, which may need frequent access to the connection.
The main benefit is that we don't need any global (`g_languageClient` and such), and we can focus on error handling at one place.

aviatesk:
> * We have so far put all assets and libs that are used in webviews in the extension into the `libs` folder. At least some of the stuff that is currently in the `assets` folder feels like stuff that should go into the `libs` folder.

done

> * Could we incorporate the content of `update_assets.jl` into `src/scripts/updateDeps.ts`? That is the central "update all things in the repo" command that I run regularly, so I think we should try to put everything in there.

Of course we can do, but I couldn't run `updateDeps.ts` successfully locally (even with fixing `juliaPath`) somehow.

> * Looking at `documenter.js`, I get the impression that all of this will only work when online because it tries to download lots of JavaScript stuff from the web, right? We tried very hard so far to include and ship all the JavaScript libraries we use in the extension, so that everything works offline. Can we try to do the same thing here? There is no fundamental reason to require to be online, right?

yeah I remember I tried exactly same thing weeks ago:
> - ~[ ] vendor scripts~ I couldn't make this work. The idea is to vendor cdn scripts inside documenter.js by ourselves

In short, I couldn't make it work; documenter.js uses require.js and I couldn't make it work locally. I can take a look again but rather I feel like I want to finish the first iteration soon.

> * if we need to ship our own `documenter.js` that works offline, we can get rid of the whole sandbox doc build step, right?

> * it seems that somewhere along the line a commit that updates the LanguageServer submodule was added. Generally, try not to do that, it makes merging a lot more complicated :) The workflow is that we update submodules on `master`, and then merge `master` back into feature branches. I think the workflow now is that you need to merge `master` into this branch here, then manually make sure that the LS submodule points to the same git commit that we use on `master`, and then commit that as the merge conflict resolution. But I'm not entirely sure. If you want I can also try to resolve this.

I had experienced the hell when I merged the master :p Removed checkin and will do in the future forever.

> * Oh, and generally needs a merge conflict resolve.

Done.

ZacLN:
Hi - catching up here.
> E.g.:
if we hover on sin, it will only show docstrings in LinearAlgebra's one and won't show Base's
we can't get docstrings on structs, variables

I can't replicate either of these examples - is this because of recent changes or because I'm not looking for the right thing?

aviatesk:
> I can't replicate either of these examples - is this because of recent changes or because I'm not looking for the right thing?

thanks for your checks. I confirmed the `sin` example has been solved -- now LS seems to recognize all the `sin` methods and so we have signature help etc. correctly.

As for `struct` and variable example, this is the MRE:

```julia
"""
    TestDoc

show me this documentation
"""
struct TestDoc
   field
end

TestDoc # <- hover over this
```
An hover actually comes up, but it just shows the definition:
```julia
struct TestDoc
    field
end
```

In REPL, we have 
```julia-repl
julia> """
           TestDoc
       
       show me this documentation
       """
       struct TestDoc
           field
       end
TestDoc

help?> TestDoc
search: TestDoc

  TestDoc


  show me this documentation
```

Same goes for variables.

src/interactive/documentation.ts
```diff
@@ -0,0 +1,227 @@
...
+function setDocumentation(inner: string) {
+    if (!inner) { return }
+    forwardStack = [] // initialize forward page stack for manual search
+    showDocumentationPane()
+    const html = createWebviewHTML(inner)
+    _setHTML(html)
+}
+
+function createWebviewHTML(inner: string) {
+    const darkMode: boolean = vscode.workspace.getConfiguration('julia.documentation').darkMode
```
pfitzseb, line 115: Can't we use `vscode.window.activeColorTheme.kind == ColorThemeKind.Dark` here? I'm not a big fan of this being a setting.

src/interactive/documentation.ts
```diff
@@ -0,0 +1,227 @@
...
+                }
+            }
+        }
+        vscode.setState({ inner: \`${inner}\` })
+    </script>
+
+</head>
+
+<body>
+    <div class="docs-main" style="padding: 1em">
```
pfitzseb, line 169: Wanna add a search bar here? Doesn't have to be functional yet, but would be cool to have :)

davidanthoff:
I'll fiddle a little bit with requirejs, maybe I can figure out a way that is entirely offline.

package.json
```diff
@@ -43,6 +43,7 @@
         "onCommand:language-julia.debug.getActiveJuliaEnvironment",
         "onLanguage:julia",
         "onLanguage:juliamarkdown",
+        "onWebviewPanel:JuliaDocumentationBrowser",
```
davidanthoff, line 46: I think I would skip this whole part, it seems reasonable to me that the doc pane doesn't stay open across VS Code shutdowns.

src/interactive/documentation.ts
```diff
@@ -0,0 +1,227 @@
...
+
+class DocumentationPaneSerializer implements vscode.WebviewPanelSerializer {
+    async deserializeWebviewPanel(deserializedPanel: vscode.WebviewPanel, state: any) {
+        panel = deserializedPanel
+        setPanelSubscription(panel)
+        const { inner } = state
+        const html = createWebviewHTML(inner)
+        _setHTML(html)
+    }
+}
```
davidanthoff, line 60: Do we need this? It seems to me we should have poster case of a stateless webview and so we should be able to skip all of this state management stuff?

src/interactive/documentation.ts
```diff
@@ -0,0 +1,227 @@
...
+function setPanelSubscription(panel: vscode.WebviewPanel) {
+    panel.onDidChangeViewState(({ webviewPanel }) => {
+        setPanelContext(webviewPanel.active)
+    })
+    panel.onDidDispose(() => {
+        setPanelContext(false)
+        if (messageSubscription !== undefined) {
+            messageSubscription.dispose()
+        }
+        panel = null
```
davidanthoff, line 71: This will set the function argument, but you probably intent to set the global? Probably best to follow the `g_` convention for global variables for now in this codebase until we move to classes to prevent that kind of bug.

src/interactive/documentation.ts
```diff
@@ -0,0 +1,227 @@
...
+                            method: 'search',
+                            params: {
+                                word: el.text,
+                                mod
+                            }
+                        })
+                    }
+                }
+            }
+        }
```
davidanthoff, line 162: Can't we do all of this in the extension, rather than in the webview?

src/interactive/documentation.ts
```diff
@@ -0,0 +1,227 @@
...
+        vscode.setState({ inner: \`${inner}\` })
```
davidanthoff, line 163: I think we should try to avoid this whole state management stuff, can't we do everything stateless?

src/interactive/documentation.ts
```diff
@@ -0,0 +1,245 @@
...
+        message => {
+            if (message.method === 'search') {
+                // withREPL(
+                //     async connection => {
+                //         const { word, mod } = message.params
+                //         const inner = await connection.sendRequest(requestTypeGetDoc, { word, mod, })
+                //         setDocumentation(inner)
+                //     },
+                //     err => { return '' }
+                // )
```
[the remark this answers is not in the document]
davidanthoff: I think the pattern here should be that we set this handler once when we create the webview, not every time we update the content, though?

src/interactive/documentation.ts
aviatesk: How/where did you find the API ??? I tried many times to find such an API and have been unable to find that ...
That should work, and yes I also really don't like to make this being a setting, of course !

aviatesk: It works, thanks for the suggestion.

aviatesk:
> Is pulling in Documenter really worth it here? We're only using it for the styling and latex rendering right now, no? Does it seem likely we'll use other parts of it in the future?

yes, we only need highlightjs and katex stuff for now. Why I just pulled Documenter.jl's assets was that I imagined that we can turn the documentation pane into the in-editor version of [our online documentation](https://docs.julialang.org/en/v1/) (by using `joinpath(Sys.BINDIR, Base.DATAROOTDIR, "doc", "julia", "html", "en", ...)` and then we want those naviation bar and stuff would be helpful.
But for now I'm okay with just go with our own minimal assets. I believe that idea would be great personally, but it would take some time and work to be done, I think.

src/interactive/documentation.ts
aviatesk: Unfortunately no, what I was using here is [the message passing API between vscode and webview](https://code.visualstudio.com/api/extension-guides/webview#scripts-and-message-passing), and this code needs to live within webview
(note that we currently don't use this code, since we disable the link handling for now)

aviatesk:
> I think the pattern here should be that we set this handler once when we create the webview, not every time we update the content, though?

When I used this code, it belongs to each webview (`message.params` here was really specific to an webview), so no, this was needed to be set each time we constructed webview. 
I _guess_ with LS, we need to subscribe to envrionment change to update this message subscription (since the documentation search would be done per-env basis).

src/interactive/documentation.ts
aviatesk: This code is needed for [webview pane serialization](https://code.visualstudio.com/api/extension-guides/webview#getstate-and-setstate), which definitely needs state

src/interactive/documentation.ts
aviatesk: ah, yes, good catch :)

package.json
aviatesk: hm, I'm honestly against this, since some users would like to see a doc pane in a specific location, which might be different from the default location, and then she wants to keep it in a same location across sessions. I think many Juno users appreciate how Juno preserve the layout in the last session.
Why do you want to skip this ?

src/interactive/documentation.ts
aviatesk: addressed, I didn't include `g_` convention just to avoid a conflict with https://github.com/julia-vscode/julia-vscode/pull/1438.
Maybe later I will rename it into `g_panel`

aviatesk:
So for now, KaTeX rendering doesn't work for multiline maths; the MRE would be `exp`:

> @davidanthoff 
> I'll follow up on the problem I mentioned, somehow the math for exp didn't render correctly for me.

davidanthoff:
> And I think this is the problem of LS side

Yes, that was my impression as well. I think that is something we still need to figure out before we can merge this, right?

davidanthoff:
I narrowed the multi-line problem down a bit more. The [line](https://github.com/julia-vscode/LanguageServer.jl/blob/295a48901e509bc096f0b2c890665af34957fe00/src/requests/features.jl#L393) that converts things into the Julia markdown types returns something that is wrong, i.e. there are no nice Latex nodes in that. I _think_ (but not sure) that is because what `get_hover` returns is somehow not properly formatted. The string that is returned by `get_hover` seems to have `$$` around equation paragraphs, and I don't think that is what we want there: I think we want equation paragraphs to be still in tripple tick and then `{math}` format (can't quote that here). That is, we want what we pass to `Markdown.parse` to be Julia markdown, but presumably `get_hover` returns "VS Code Markdown" because we are sending that directly to VS Code?

So multiple options out here:
- Could we have a flag for `get_hover` that either returns VS Code style markdown or Julia markdown? @ZacLN?
- Should we maybe in general always return markdown from the LS and do the conversion to HTML in the extension? I think that would actually be good for other reasons as well: the LS would no longer need to know about the VS Code specific command URIs, for example. We could probably use markdown-it in the extension to do the conversion? But I'm a bit worried whether we can pull this off in half a day...

Other ideas?

davidanthoff:
Here is another potential benefit of doing the MD=>HTML conversion in the extension: I think we can use plugins for both highlighting and math  in the conversion step, and don't have to rely on the Javascript running in the webview anymore.

I'm going to explore this option a bit more now.

package.json
```diff
@@ -303,19 +331,34 @@
...
             },
             {
                 "command": "language-julia.executeJuliaCodeInREPL",
-                "key": "ctrl+Enter",
+                "key": "ctrl+enter",
                 "when": "editorTextFocus && editorLangId == juliamarkdown"
             },
+            {
+                "command": "language-julia.show-documentation",
+                "key": "ctrl+j ctrl+d",
```
pfitzseb, line 349: Should be `Alt+J Alt+D` if we go with https://github.com/julia-vscode/julia-vscode/pull/1454.

pfitzseb:
Turns out that's just how [markdown strings are rendered to plain text](https://github.com/JuliaLang/julia/blob/master/stdlib/Markdown/src/render/plain.jl#L87-L91):
````
julia> println(Base.Docs.@doc exp)
```
exp(x)
```

Compute the natural base exponential of `x`, in other words $e^x$.

# Examples

```jldoctest
julia> exp(1.0)
2.718281828459045
```

```
exp(A::AbstractMatrix)
```

Compute the matrix exponential of `A`, defined by

$$
e^A = \sum_{n=0}^{\infty} \frac{A^n}{n!}.
$$
````

Replacing `r"^\$\$$" => "```math"` or whatever should be an ok workaround for now.

> Here is another potential benefit of doing the MD=>HTML conversion in the extension: I think we can use plugins for both highlighting and math in the conversion step, and don't have to rely on the Javascript running in the webview anymore.

Not sure if I see the huge benefit of that, tbh.

davidanthoff:
> Replacing `r"^\$\$$" => "```math"` or whatever should be an ok workaround for now.

Problem is that `$$` opens and closes math blocks, but we need different opening and closing replacements here.

But, I did implement the whole conversion to HTML in the extension: https://github.com/julia-vscode/julia-vscode/pull/1455, and I think that is all around a more robust solution.

PR #1319, opened 2020-06-09
package up vscodeserver

aviatesk:
Packaging up our REPL server module, which will enable
- precompilation
- natural test suite

And several other enhancements:
- REPL will keep living even when an internal error happens when in extension development


.vscode/launch.json
```diff
@@ -7,23 +7,32 @@
...
+            "args": [
+                "--extensionDevelopmentPath=${workspaceFolder}",
+                "--extensionTestsPath=${workspaceFolder}/out/test/suite/index"
+            ],
+            "outFiles": [
+                "${workspaceFolder}/out/test/**/*.js"
             ],
-			"outFiles": ["${workspaceFolder}/out/test/**/*.js"],
-			"preLaunchTask": "npm: test-compile"
+            "preLaunchTask": "npm: test-compile"
```
davidanthoff, line 35: Same general comment as with the other ones: please try to keep diffs tidy :) I think the only thing changed here is the addition of an env variable, but figuring that out from this diff is involved and I can't be sure. Would be much easier if the diff had only the line changed that actually changed. I think in general we should try to put reformatting stuff into separate PRs.

Also, we don't need the `DEBUG_MODE` in this PR at all, right? It seems like an unrelated change?

scripts/vscodeserver/Project.toml
```diff
@@ -0,0 +1,15 @@
+name = "vscodeserver"
```
davidanthoff, line 1: Maybe `VSCodeServer`? That seems more in line with the package naming guidelines.

scripts/vscodeserver/Project.toml
```diff
@@ -0,0 +1,15 @@
...
+uuid = "9f5989ce-84fe-42d4-91ec-6a7a8d53ed0f"
+authors = ["julia-vscode extension contributors"]
```
davidanthoff, line 3: I think you can just leave that one out ;)

scripts/vscodeserver/boot_repl.jl
```diff
@@ -0,0 +1,41 @@
+# this script basially only handles `Base.ARGS`
+
+# XXX: is this needed ?
+!(Sys.isunix() || Sys.iswindows()) && error("Unknown operating system.")
```
davidanthoff, line 4: Na, can probably be taken out!

scripts/vscodeserver/boot_repl.jl
```diff
@@ -0,0 +1,41 @@
...
+        using vscodeserver
+        Pkg.activate() # back to default env
+    catch e
+        @error e
+    finally
+        redirect_stderr(old)
+        close(wr)
+        close(rd)
+    end
+end
```
davidanthoff, line 23: Could we instead just push a path onto `LOAD_PATH`? If that works we wouldn't have to change the active project at all, which might be more robust?

scripts/vscodeserver/src/vscodeserver.jl
```diff
@@ -180,114 +182,134 @@ function get_modules(toplevel = nothing, mods = Set(Module[]))
...
-            source_code = params["code"]
-            mod = params["module"]
+function serve(args...; is_dev = false)
+    conn = connect(args...)
+    conn_endpoint[] = JSONRPC.JSONRPCEndpoint(conn, conn)
+    run(conn_endpoint[])
 
-            resolved_mod = try
-                module_from_string(mod)
+    if is_dev
```
davidanthoff, line 194: Hm, I'm not so sure I like it that we have a completely different code path at dev. What is the benefit of this?

.vscode/launch.json
aviatesk: yeah, ... the thing is now #1262 _automatically_ formats these things on save (I just forgot to format those JSON files).
I should have commented the main diff here at least, and yeah, I will try to keep diffs tidy.

As for `DEBUG_MODE`, it's what I would like to introduce in this PR. Will explain in another response.

scripts/vscodeserver/src/vscodeserver.jl
aviatesk: this flag will be automatically set when we start REPL in development mode (which just uses `DEBUG_MODE` flag I set in `launch.json`):
https://github.com/julia-vscode/julia-vscode/blob/42e7402ffda79da310b8fa37d0b86b827b1ff420/scripts/vscodeserver/boot_repl.jl#L40

The main benefit here is as I mentioned:
> REPL will keep living even when an internal error happens when in extension development

In other word, while developing interactive code, we often have internal error, which currently causes this `@async` block just to be killed, and then we need to reboot REPL.

Another benefit is to use `Base.invokelatest` so that we can have every hot reload of message handlers.

---

Well, this is not exactly related to "packaging up", so I'm okay to do this in another PR if you want.

scripts/vscodeserver/Project.toml
aviatesk: yeah, sounds good. Do we want to move `vscodeserver` directly into `scripts/packages` ?

aviatesk:
I have existing test code in Juno's code base, but let's do "enriching test" (adding test, enable CI) stuff in another PR.

aviatesk:
two things remaining to consider before merging:
- whether to move `VSCodeServer` into `scripts/packages`
- separate `is_dev` stuff into another PR ?

scripts/vscodeserver/Project.toml
davidanthoff: Hm, good question. So far I had thought of the `packages` folder as stuff that is git submodules, which I think we can avoid for this story here, right? Probably doesn't matter much in the end.

scripts/vscodeserver/src/vscodeserver.jl
davidanthoff: Ok, sounds good, lets leave it like you changed it!

.vscode/launch.json
davidanthoff: Ah, right!

scripts/vscodeserver/Project.toml
aviatesk: agreed.

aviatesk:
> I can handle adding tests, I'll make it such that they run as part of our Azure Pipelines stuff.

nice, I may create a PR on top of your PR that will move Juno's test base (not so much though) here.

davidanthoff:
We might also take a look at repl.jl: I think everything in there was just copy-pasted at some point from Juno, so some of that code might have been updated in the meantime on the Juno side of things, in which case it might be nice if we would also get the newer versions :)

PR #1321, opened 2020-06-10
Move vscodeserver

davidanthoff:
@aviatesk This follows up on your suggestion to put everything into the packages folder. I think that makes sense, and I used the opportunity to also adjust the casing of folder and filenames to the new package name.

This PR also fixes the debugger crash reporting, so that makes it a bug fix.


scripts/packages/VSCodeServer/src/VSCodeServer.jl
```diff
@@ -186,31 +186,31 @@ const conn_endpoint = Ref{Union{Nothing,JSONRPC.JSONRPCEndpoint}}(nothing)
...
+function serve(args...; is_dev = false, crashreporting_pipename::Union{AbstractString,Nothing}=nothing)
     conn = connect(args...)
     conn_endpoint[] = JSONRPC.JSONRPCEndpoint(conn, conn)
     run(conn_endpoint[])
 
     if is_dev
         @async while true
             try
-                Base.invokelatest(handle_message)
+                Base.invokelatest(handle_message, crashreporting_pipename=crashreporting_pipename)
```
aviatesk, line 197: I guess we don't need this for dev mode ? otherwise my internal errors while developing will spam up the crash report :p

scripts/terminalserver/terminalserver.jl
```diff
@@ -1,6 +1,6 @@
 # this script basially only handles `Base.ARGS`
 
-Base.push!(LOAD_PATH, @__DIR__)
+Base.push!(LOAD_PATH, joinpath(@__DIR__, "..", "packages"))
```
aviatesk, line 3: how about locating this file under `scripts` as we did for `error_handler.jl` ? I don't think we will add more files to `terminalserver` directory.

scripts/packages/VSCodeServer/src/VSCodeServer.jl
davidanthoff: Yep, done.

Although it wouldn't even matter that much: all the crash reports during debug mode go into their own crash reporting target, so they are isolated from the other ones in any case. And then we have one crash report target for the insider channel and one for the release channel.

scripts/terminalserver/terminalserver.jl
davidanthoff: I think the only reason I put error_handler.jl in the root is because it is used by multiple things. I kind of like to have things in sub folders per functionality, seems a bit cleaner? If anything, I'd almost want to move `error_handler.jl` into some `shared` folder or something like that.

scripts/packages/VSCodeServer/src/VSCodeServer.jl
aviatesk: ah I never know that. I should take a more closer look at our error reports first.

scripts/terminalserver/terminalserver.jl
aviatesk: okay, understand.

PR #1329, opened 2020-06-11
fix repl again

### LINKED ISSUE #1328: `@macroexpand` ignored in Julia REPL
Julia 1.4.2, vscode-insider 0.16.3, either in the REPL or inline mode, it seems to completley ignore the presence of `@macroexpand`:

```julia
julia> @macroexpand @time 1
  0.000000 seconds
1
```

Guessing its just inadvertantly eval'ing the output of macros which return an expression, here's another version of this bug where you should be getting the expression `:(1+2)` but vscode is evaluating it, 

```julia
julia> macro return_expr()
           :(:(1+2))
       end
@return_expr (macro with 1 method)

julia> @return_expr
3
```

aviatesk:
fixes #1328 by wrapping evaluation in our own functions (i.e. `evalrepl`)
accompanied with:
- cutting off internal calls from stacktrace
- use `include_string` and more graceful stack trace

and I've found the same kind of error can happen for `Main.ans` update, so fixed it as well.

### DIFF
--- a/scripts/packages/VSCodeServer/src/repl.jl
+++ b/scripts/packages/VSCodeServer/src/repl.jl
@@ -59,6 +59,56 @@ function hideprompt(f)
     r
 end
 
+function hook_repl(repl)
+    if !isdefined(repl, :interface)
+        repl.interface = REPL.setup_interface(repl)
+    end
+    main_mode = get_main_mode(repl)
+
+    # TODO: set up REPL module ?
+    main_mode.on_done = REPL.respond(repl, main_mode; pass_empty = false) do line
+        quote
+            $(evalrepl)(Main, $line, $repl, $main_mode)
+        end
+    end
+end
+
+function evalrepl(m, line, repl, main_mode)
+    return try
+        JSONRPC.send_notification(conn_endpoint[], "repl/starteval", nothing)
+        try
+            repleval(m, line, REPL.repl_filename(repl, main_mode.hist))
+        catch err
+            display_repl_error(stderr, err, stacktrace(catch_backtrace()))
+            nothing
+        end
+    catch err
+        # This is for internal errors only.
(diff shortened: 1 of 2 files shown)


scripts/packages/VSCodeServer/src/repl.jl
```diff
@@ -59,6 +59,56 @@ function hideprompt(f)
...
+end
+
+function display_parse_error(io, err::Meta.ParseError)
+  printstyled(io, "ERROR: "; bold = true, color = Base.error_color())
+  printstyled(io, "syntax: ", err.msg; color = Base.error_color())
+  println(io)
+end
+
+# don't inline this so we can find it in the stacktrace
+@noinline repleval(m, code, file) = include_string(m, code, file)
```
pfitzseb, line 101: Why `include_string` over [`eval`](https://github.com/JunoLab/Atom.jl/blob/master/src/repl.jl#L196)?

pfitzseb: So we don't have to parse in an extra step. Not sure if I like it, but we can iterate on this in a follow-up.

aviatesk: another benefit is we can specify `file` within `include_string` so that stacktrace shows `REPL[exec_cnt]` as a location, without fixing `Core.LineNumber`s in `Expr`.
And I don't see the downside of using it ? Not too sure about that though.

pfitzseb: The only slightly annoying thing about this is that everything's wrapped in a `LoadError`:
image

pfitzseb: Fixed by https://github.com/julia-vscode/julia-vscode/pull/1331.

PR #1331, opened 2020-06-11
improve repl error printing

pfitzseb:
Removes the `LoadError` wrapper when displaying errors in REPL.

### DIFF
--- a/scripts/packages/VSCodeServer/src/repl.jl
+++ b/scripts/packages/VSCodeServer/src/repl.jl
@@ -92,9 +92,9 @@ function evalrepl(m, line, repl, main_mode)
 end
 
 function display_parse_error(io, err::Meta.ParseError)
-  printstyled(io, "ERROR: "; bold = true, color = Base.error_color())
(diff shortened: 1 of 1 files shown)


scripts/packages/VSCodeServer/src/repl.jl
```diff
@@ -92,22 +92,23 @@ function evalrepl(m, line, repl, main_mode)
 end
 
 function display_parse_error(io, err::Meta.ParseError)
-  printstyled(io, "ERROR: "; bold = true, color = Base.error_color())
-  printstyled(io, "syntax: ", err.msg; color = Base.error_color())
-  println(io)
+    printstyled(io, "ERROR: "; bold = true, color = Base.error_color())
+    printstyled(io, "syntax: ", err.msg; color = Base.error_color())
+    println(io)
```
aviatesk, line 97: this function is actually a dead code, so let's remove this ?

pfitzseb: Done.

Issue #1183, opened 2020-05-11
Documentation Pane


aviatesk:
Juno's documentation pane works in two ways (and I would like to say this is _very_ useful):
- show documentation for a symbol under a cursor: just use Julia's builtin documentation system
- search for documentation with fuzzy keyword matching: we have [DocSeeker.jl](https://github.com/JunoLab/DocSeeker.jl) for that

In order to have this in VSC, we need to have our own UI within webview, since VSC doesn't offer such an API AFAIU.

aviatesk:
ops, sorry I didn't check the existing issues

davidanthoff:
No worries :) Always better to just close something as a duplicate than not getting it in the first place!

PR #1336, opened 2020-06-12
more careful LS startup:

aviatesk:
While I tackle #1183 , I found this error and so extract here.
Currently we set `g_languageClient` when it's fired up, rather than its's ready, which may result in more internal "language client ready yet" error.

commits:
- set `g_languageClient` object on LS "ready" rather than fire
- don't create global ls start up indicator

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -21,16 +21,11 @@ import * as weave from './weave'
 let g_settings: settings.ISettings = null
 let g_languageClient: LanguageClient = null
 let g_context: vscode.ExtensionContext = null
-let g_lsStartup: vscode.StatusBarItem = null
 
 export async function activate(context: vscode.ExtensionContext) {
     await telemetry.init(context)
 
     telemetry.traceEvent('activate')
-
-    g_lsStartup = vscode.window.createStatusBarItem()
-    g_lsStartup.text = 'Starting Julia Language Server...'
-    g_lsStartup.show()
 
     telemetry.startLsCrashServer()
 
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -107,13 +102,17 @@ function changeConfig(params: vscode.ConfigurationChangeEvent) {
...
+            setLanguageClient()
         }
         startLanguageServer()
     }
 }
 
 async function startLanguageServer() {
+    const startupNotification = vscode.window.createStatusBarItem()
+    startupNotification.text = 'Starting Julia Language Server...'
+    startupNotification.show()
```
davidanthoff, line 114: I had this created at the original place so that it shows up as soon as the extension starts up. I think moving this down here might potentially mean that it will take a while until this element pops up in the UI?

Agreed that we don't need to have this as a global, but maybe it could be created in the old place, and then we do `startLanguageServer(startupNotification)`, i.e. just pass the element that was created in `activate` to the `startLanguageServer` function?

aviatesk: >  think moving this down here might potentially mean that it will take a while until this element pops up in the UI?

um, I don't feel any delay because of that; after all what we do in `activate` before calling `startLanguageServer` is registering commands/telemetry stuff and activate child functions.

I also would like to point out this notification is essentially tied to LS start up, not to extension, and we want to see that for each `startLanguageServer` call, rather than the very first `activate` call ? It can happen when we change julia path.

davidanthoff: Sounds good!

PR #1340, opened 2020-06-14
Mark julia.executablePath as machine-overridable

davidanthoff:
I think right now any user settings are automatically also used in a remote setting, which doesn't make sense. I believe that with this change, a user setting wouldn't automatically be used for a remote session, but it would still allow a user to override the setting in a workspace specific setting.

Would fix https://github.com/JuliaLang/julia/pull/36257.

### DIFF
--- a/package.json
+++ b/package.json
@@ -380,7 +380,7 @@
                     "type": "string",
                     "default": "",
                     "description": "Points to the julia executable.",
-                    "scope": "window"
+                    "scope": "machine-overridable"
                 },
                 "julia.lint.run": {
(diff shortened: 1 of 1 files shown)


aviatesk:
> `machine-overridable` - Machine specific settings that can be overridden by workspace or folder settings.

Does "machine specific settings" mean a local setting won't be used for remote session, right ? A local user-level setting of `executablePath` still keeps to work for local projects (without project/folder level setting) ?

davidanthoff:
Exactly, that is how I understand the documentation.

PR #1341, opened 2020-06-14
highlight stackframe on inline evaluation error

### LINKED ISSUE #1178: Backtrace highlighting
Whenever Juno displays a backtrace, all lines mentioned in it are also highlighted in open editors:

image

### LINKED ISSUE #1389: Trim stacktraces
A lot of stack-track error message is concerned not with the code i ran, but with the VS-Code internals that triggered it.
The REPL and Juno also have ths problem.
but in those cases it is far fewer lines.

Example:
```
Error During Test at /Users/oxinabox/JuliaEnvs/Bars.jl/test/run.jl:7
  Got exception outside of a @test
  UndefVarError: bar_day not defined
  Stacktrace:
   [1] (::var"#119#122"{DummyAgent,Date,TestClient,Stub{:dummy_simulate},Stub{:dummy_get_foos}})() at /Users/oxinabox/JuliaEnvs/Bars.jl/test/run.jl:27
   [2] apply(::var"#119#122"{DummyAgent,Date,TestClient,Stub{:dummy_simulate},Stub{:dummy_get_foos}}, ::Mocking.PatchEnv) at /Users/oxinabox/.julia/packages/Mocking/U41JO/src/patch.jl:64
   [3] apply(::Function, ::Array{Patch,1}; debug::Bool) at /Users/oxinabox/.julia/packages/Mocking/U41JO/src/patch.jl:71
   [4] apply(::Function, ::Array{Patch,1}) at /Users/oxinabox/.julia/packages/Mocking/U41JO/src/patch.jl:71
   [5] top-level scope at /Users/oxinabox/JuliaEnvs/Bars.jl/test/run.jl:26
   [6] top-level scope at /usr/local/src/julia/julia-1.4/usr/share/julia/stdlib/v1.4/Test/src/Test.jl:1113
   [7] top-level scope at /Users/oxinabox/JuliaEnvs/Bars.jl/test/run.jl:10
   [8] include_string(::Module, ::String, ::String) at ./loading.jl:1080
   [9] #invokelatest#1 at ./essentials.jl:712 [inlined]
   [10] invokelatest at ./essentials.jl:711 [inlined]
   [11] (::VSCodeServer.var"#45#48"{Dict{String,Any},String,Int64,Int64,String,Module,Bool})() at /Users/oxinabox/.vscode/extensions/julialang.language-julia-0.16.7/scripts/packages/VSCodeServer/src/VSCodeServer.jl:259
   [12] withpath(::VSCodeServer.var"#45#48"{Dict{String,Any},String,Int64,Int64,String,Module,Bool}, ::String) at /Users/oxinabox/.vscode/extensions/julialang.language-julia-0.16.7/scripts/packages/VSCodeServer/src/repl.jl:113
   [13] (::VSCodeServer.var"#44#47"{Dict{String,Any},String,Int64,Int64,String,Module,Bool,Bool})() at /Users/oxinabox/.vscode/extensions/julialang.language-julia-0.16.7/scripts/packages/VSCodeServer/src/VSCodeServer.jl:257
   [14] hideprompt(::VSCodeServer.var"#44#47"{Dict{String,Any},String,Int64,Int64,String,Module,Bool,Bool}) at /Users/oxinabox/.vscode/extensions/julialang.language-julia-0.16.7/scripts/packages/VSCodeServer/src/repl.jl:34
   [15] handle_message(; crashreporting_pipename::String) at /Users/oxinabox/.vscode/extensions/julialang.language-julia-0.16.7/scripts/packages/VSCodeServer/src/VSCodeServer.jl:237
   [16] (::Base.var"#inner#2"{Base.Iterators.Pairs{Symbol,String,Tuple{Symbol},NamedTuple{(:crashreporting_pipename,),Tuple{String}}},typeof(VSCodeServer.handle_message),Tuple{}})() at ./essentials.jl:715
   [17] #invokelatest#1 at ./essentials.jl:716 [inlined]
   [18] macro expansion at /Users/oxinabox/.vscode/extensions/julialang.language-julia-0.16.7/scripts/packages/VSCodeServer/src/VSCodeServer.jl:207 [inlined]
   [19] (::VSCodeServer.var"#40#42"{String})() at ./task.jl:358
  ```

AFAICT everything after `[7]` is VS-Code internals.
So over 60% of that message is noise.

We can can do something like what  [the `Test` stdlib does](https://github.com/JuliaLang/julia/blob/be72a571a63e134788c8bb7d997098e9b9aee019/stdlib/Test/src/Test.jl#L51-L61) and filter stuff out of the stacktrace before it is rendered.

### PR DESCRIPTION
**Pull request by aviatesk**
image

remaining TODOs:
- [x] cut off internal errors
- [x] more sensible validation of error decorations: clear all the error highlights on the last frame invalidation or something like that
- [x] utility commands as hover links (jump to last/next/prev/first frame, clear current frame, clear all the frames)
- [x] more markdown hack, clickable frame list (prepare for future rendering of other stuff like `MethodList`)

fixes:
- closes #1178 
- closes #1389

### CODE REVIEW

**Comment by pfitzseb** [comment]
Pushed a commit to improve the UI logic a bit:
- Display both results and error highlights.
- Only display error highlights for the latest evaluation.

File: src/interactive/results.ts [hunk 1]
```diff
@@ -182,19 +182,24 @@ interface Frame extends _Frame {
...
+
+    drawStackFrame(stackframe.map(frame => {
         return {
             path: frame.path,
             line: frame.line,
             err
         }
-    })
-    refreshStackTrace()
+    }))
```
aviatesk: Actually this change has one regression; results are now only registered for currently open editors, meaning we can't get error highlight for editors that are opened later.
Will make a change on this anyway.

pfitzseb: Cool, let me know when you want another review on this :)

aviatesk: Will do ;)

src/interactive/results.ts
```diff
@@ -261,17 +261,27 @@ export function removeResult(target: Result) {
     return results.splice(results.indexOf(target), 1)
 }
 
+export function removeErrorResult(target: Result) {
+    target.remove()
+    const targetIndex = getErrorResults().indexOf(target)
+    return stackFrameHighlights.highlights.splice(targetIndex, 1)
+}
```
pfitzseb, line 268: FWIW, I quite like the Juno behaviour where the stacktrace is attached to the last result. So when you clear the result that caused the error display the error trace is also cleared.

aviatesk: yeah, I'm just trying to implement that, but I'm still not sure what the best way to implement that (maybe we need to keep the last result as a global object and subscribe to its validation or something like that ?) I've not reviewed how you implemented that in Juno, but I'm very welcome if you want to implement the logic.
If so I will tackle on implementing utility commands that can be invoked as a hover links :)

pfitzseb: Done.

aviatesk: your implementation looks sooo valid... thanks !

aviatesk:
> I still think a clickable list makes more sense though. Maybe we can re-use (or write) a markdown show methods for stacktraces?

yeah, I have mostly done the exactly same idea in my machine ;) Will push later.

aviatesk:
then I will ask you for another review @pfitzseb .

scripts/packages/VSCodeServer/src/repl_protocol.jl
```diff
@@ -8,11 +8,18 @@ JSONRPC.@dict_readable struct ReplRunCodeRequestParams <: JSONRPC.Outbound
...
+end
+Frame(st::Base.StackFrame) = Frame(fullpath(string(st.file)), st.line)
+
 JSONRPC.@dict_readable struct ReplRunCodeRequestReturn <: JSONRPC.Outbound
     inline::String
     all::String
-    iserr::Bool
+    stackframe::Union{Nothing,Vector{Frame}}
 end
+ReplRunCodeRequestReturn(inline, all) = ReplRunCodeRequestReturn(inline, all, nothing)
```
aviatesk, line 22: @davidanthoff sorry for turning back the discussion, but do you think we can change the [custom](https://github.com/julia-vscode/julia-vscode/pull/1371#discussion_r443005108) ?
After writing this code, I honestly feel a bit awkward that I need to define `Frame` here, because `Frame` (and its constructor method) really depends on the logic inside eval.jl, but because I had to define this here, I had to switch between files while writing this code.
If we're okay to split these protocols into each feature file, we can just focus on one single file while developing a feature, and that would feel much better, I guess.

aviatesk:
Okay I think now this is ready for review/merge.
Now we have:
- cut off internal error stack frames
- stacktrace highlight on inline evaluation error
- clickable stacktrace in hover
- frame jump buttons in hover

image

The command string protocol (`vscode-command:<cmd>?argstart<args>argend` -> `command:<cmd>?<args>`) can be used for any other features where `MarkdownString` is supported, but I agree with the current implementation is very ugly and so I'm very welcome to hear your ideas on this.

@pfitzseb , can you take a look on this again ?

scripts/packages/VSCodeServer/src/eval.jl
```diff
@@ -131,3 +95,69 @@ function safe_render(x)
...
+        i === 1 && return line # "Stacktrace:"
+        m = match(LOCATION_REGEX, line)
+        m === nothing && return line
+        linktext = string(m[:path], ':', m[:line])
+        args = JSON.json((path = fullpath(m[:path]), line = m[:line]))
+        linkbody = vscode_cmd_string("language-julia.openFile", args)
+        linktitle = string("Go to ", linktext)
+        return "$(i-1). `$(m[:body])` at [$(linktext)]($(linkbody) \"$(linktitle)\")"
+    end |> joinlines
+end
```
aviatesk, line 154: I couldn't come up with any good idea to implement this, other than using ugly `Regex` ...
I think we can also hack into the `Base` internals, but I don't think it would look clean.

pfitzseb:
Looks great!
Pushed two small tweaks. Will give an in-depth review tomorrow.

davidanthoff:
I just tried this PR on my machine, but I can't get it to display anything. Is there something I need to configure to make this work?

aviatesk:
> I just tried this PR on my machine, but I can't get it to display anything. Is there something I need to configure to make this work?

Do you still get inline results ? Our default `julia.execution.resultType` configuration is set to `REPL` by default, which means you can't get inline results by default setting, but if you get inline results this stackframe stuff will work out of box.

Or maybe your color theme can't display error highlight well. I'm using the color below, but this is not guaranteed to show nice color in any theme:
https://github.com/julia-vscode/julia-vscode/blob/d28febf1fdb2ed9fb665249ee98395b90cc45a06/src/interactive/results.ts#L106

davidanthoff:
I have the config setting as `both`, which should work, right? Color scheme is the default light one.

aviatesk:
I confirmed the color highlighting works well in the default light theme.
Can you try this example ?
```julia
f(x) = g(x, x)

function g(x,y)
   return x // y
end

f(0) # execute this line will invoke every feature
```
The highlight won't occur if there is no stacktrace, e.g. toplevel undefined variable and such.

davidanthoff:
Hm, it doesn't show anything... Is there maybe a problem with windows file paths? I'm on Windows.

pfitzseb:
FWIW, I've been getting a bunch of weird bugs when switching branches and running the extension via F5 where VSCode was running some old version of the code or something. Usually fixed by restarting all VSCode instances.

src/interactive/repl.ts
```diff
@@ -389,12 +389,10 @@ async function evaluate(editor: vscode.TextEditor, range: vscode.Range, text: st
     if (resultType !== 'REPL') {
         if (result.stackframe) {
             results.setStackTrace(r, result.all, result.stackframe)
-            // no need to show details here because the inline error display adds them anyways:
-            r.setContent(results.resultContent(' ' + result.inline + ' ', ''))
         } else {
             results.clearStackTrace()
-            r.setContent(results.resultContent(' ' + result.inline + ' ', result.all))
```
aviatesk: @pfitzseb why reverting this ? Removing the duplication looked good to me.

pfitzseb: Because the stacktrace doesn't stay attached to the result forever (most importantly, only as long as nothing else has been evaluated).

scripts/packages/VSCodeServer/src/eval.jl
```diff
@@ -71,63 +71,115 @@ function repl_runcode_request(conn, params::ReplRunCodeRequestParams)
...
+
+function sprint_error(err)
+    sprintlimited(err, nothing, func = Base.display_error, limit = MAX_RESULT_LENGTH)
+end
+
 function render(err::EvalError)
-    str = sprintlimited(err.err, err.bt, func=Base.display_error, limit=MAX_RESULT_LENGTH)
+    bt = err.bt
+    i = find_frame_index(bt, @__FILE__, inlineeval)
+    bt = bt[1:(i === nothing ? end : i - 4)]
```
pfitzseb, line 133: Why `- 4`? Can you add a comment?

src/interactive/repl.ts
aviatesk: makes sense. Related, do you think it makes sense to keep the previous stackframe after the next evaluation ?
I would like to keep them until I hit another stackframe error (and iirc this is what Juno behaves like)

scripts/packages/VSCodeServer/src/eval.jl
```diff
@@ -163,5 +163,23 @@ function backtrace_string(bt)
...
+    for cp in codeunits(uri)
+        if cp in UNESCAPED
+            print(io, Char(cp))
+        else
+            print(io, '%')
+            @printf(io, "%2X", cp)
+        end
+    end
+    return String(take!(io))
+end
```
aviatesk: This is much better, nice.

aviatesk: Well, can you tell me any reference of this implementation ? Especially around `@printf(io, "%2X", cp)` ?

scripts/packages/VSCodeServer/src/eval.jl
aviatesk: added note

scripts/packages/VSCodeServer/src/eval.jl
pfitzseb: I have the implementation at https://gist.github.com/pfitzseb/d6a9ebee8404e12ec69132af6cf2dada.

The `@printf(io, "%2X", cp)` thing just prints `cp` as a hex (`x`) string with 2 chars (`2`) and in uppercase (`X`).

src/interactive/repl.ts
pfitzseb: Hm, ok. Not sure what I prefer, honestly.

scripts/packages/VSCodeServer/src/eval.jl
aviatesk: Got it. Let's add the test examples when we setup the test for this code.

src/interactive/repl.ts
aviatesk: Let's see how we feel on this.

aviatesk: `e6066b6`

aviatesk:
yeah, they can be addressed in follow ups, while I'm also interested in our current thinking on keybindings:
if we can use <kbd>ctrl-j</kbd> for our various commands, the frame navigation commands can be, say, <kbd>ctrl-j down</kbd> (`gotoNextFrame`), but we already have an issue of this: https://github.com/julia-vscode/julia-vscode/issues/1381

src/interactive/results.ts
```diff
@@ -123,85 +162,240 @@ export class Result {
...
+
+function setStackFrameHighlight(
+    err: string,
+    frames: Frame[],
+    editors: vscode.TextEditor[] = vscode.window.visibleTextEditors
+) {
+    stackFrameHighlights.err = err
+    frames.forEach(frame => {
+        const targetEditors = editors.filter(editor => frame.path === editor.document.fileName)
```
davidanthoff, line 274: This comparison doesn't work properly on Windows: Julia returns path with a capital drive letter, VS Code with a lower case drive letter, and we never get any match here.

We should probably in general base all of this on URIs, not on paths, and try to use the VS Code URI type as much as we can for this kind of thing. See [here](https://code.visualstudio.com/updates/v1_46#_file-path-casing) for some recent casing specific points.

pfitzseb: Fixed now. Turns out we need both this comparison (for untitled editors) _and_ a comparison of the normalized URIs (for normal files).

davidanthoff: Can't we do all the comparisons based on URIs? The URI for an unsaved file is `untitled:Untitled-1`, i.e. it doesn't use the `file:` scheme (because it isn't), but the `untitled:` scheme instead. But I guess the problem is that we can't know whether a given file name we get in a stack trace is a file that exists or an untitled file, right?

pfitzseb: Yup. We can of course have a heuristic that constructs the URI with the right scheme when the file name starts with `Untitled`, but meh...

aviatesk:
I'm pretty happy with this enhancement so I definitely want to mention about this somewhere, while I don't mind if we don't hold this until 7/1.

davidanthoff:
We can just ship now, and then mention it in the release notes for v0.17?

aviatesk:
Yeah, let's ship it now. This feature would be added on anyways.

PR #1342, opened 2020-06-14
Use kwspacing for format option

davidanthoff:
LanguageServer.jl is asking for that name, so without this PR it will never get a value for this config.


ZacLN:
https://github.com/julia-vscode/LanguageServer.jl/pull/756

Issue #1338, opened 2020-06-13
Error displaying a StaticArray


cdsousa:
I'm getting this error in the vscode julia repl:
```
julia> using StaticArrays

julia> SA[1,2]
2-element SArray{TupleERROR: {MethodError: no method matching size(::SArray{Tuple{2},Int64,1,2})
The applicable method may be too new: running in world age 27132, while current world is 27133.
Closest candidates are:
  size(::Union{LinearAlgebra.Adjoint{T,#s15} where #s15<:Union{StaticArray{Tuple{#s12},T,1} where #s12, StaticArray{Tuple{#s13,#s14},T,2} where #s14 where #s13}, LinearAlgebra.Diagonal{T,#s22} where #s22<:(StaticArray{Tuple{#s23},T,1} where #s23), LinearAlgebra.Hermitian{T,#s19} where #s19<:(StaticArray{Tuple{#s20,#s21},T,2} where #s21 where #s20), LinearAlgebra.LowerTriangular{T,#s27} where #s27<:(StaticArray{Tuple{#s28,#s29},T,2} where #s29 where #s28), LinearAlgebra.Symmetric{T,#s16} where #s16<:(StaticArray{Tuple{#s17,#s18},T,2} where #s18 where #s17), LinearAlgebra.Transpose{T,#s12} where #s12<:Union{StaticArray{Tuple{#s12},T,1} where #s12, StaticArray{Tuple{#s13,#s14},T,2} where #s14 where #s13}, LinearAlgebra.UnitLowerTriangular{T,#s33} where #s33<:(StaticArray{Tuple{#s34,#2s35},T,2} where #s35 where #s34), LinearAlgebra.UnitUpperTriangular{T,#s30} where #s30<:(StaticArray{Tuple{#s31,#s32},T,2} where #s32 where #s31), LinearAlgebra.UpperTriangular{T,#s24} where #s24<:(StaticArray{Tuple{#s25,#s26},T,2} where #s26 where #s25), StaticArray{Tuple{#s35},T,1} where #s35, StaticArray{Tuple{#s14,#s13},T,2} where #s13 where #s14, StaticArray{#s35,T,N} where N where #s35<:Tuple} where T) at /Users/csousa/.julia/packages/StaticArrays/mlIi1/src/abstractarray.jl:9 (method too new to be called from this world context.)
  size(::AbstractArray{T,N}, ::Any) where {T, N} at abstractarray.jl:38
[...]

PR #1344, opened 2020-06-15
fixes https://github.com/julia-vscode/julia-vscode/issues/1338

aviatesk:


scripts/packages/VSCodeServer/src/VSCodeServer.jl
```diff
@@ -204,7 +204,7 @@ function serve(args...; is_dev = false, crashreporting_pipename::Union{AbstractS
     else
         @async try
             while true
-                handle_message(crashreporting_pipename=crashreporting_pipename)
+                Base.invokelatest(handle_message; crashreporting_pipename = crashreporting_pipename)
```
davidanthoff, line 207: So in general we have so far used `invokelatest` very narrowly, i.e. not just wrapped a large chunk of our code in it (as this one does), but rather very specifically in places where we know that a world age problem can occur because we call out to user code. That still seems a good tactic to me.

pfitzseb: The "minimal" fix would be something like
```diff
diff --git a/scripts/packages/VSCodeServer/src/VSCodeServer.jl b/scripts/packages/VSCodeServer/src/VSCodeServer.jl
index 8eda899..1dd18fa 100644
--- a/scripts/packages/VSCodeServer/src/VSCodeServer.jl
+++ b/scripts/packages/VSCodeServer/src/VSCodeServer.jl
@@ -203,7 +203,7 @@ function serve(args...; is_dev = false, crashreporting_pipename::Union{AbstractS
     if is_dev
         @async while true
             try
-                Base.invokelatest(handle_message)
+                handle_message()
             catch err
                 Base.display_error(err, catch_backtrace())
             end
@@ -295,11 +295,12 @@ function handle_message(; crashreporting_pipename::Union{AbstractString,Nothing}
             end
         end
     elseif msg["method"] == "repl/getvariables"
-        vars = get_variables()
+        vars = Base.invokelatest(get_variables)
 
         JSONRPC.send_success_response(conn_endpoint[], msg, vars)
     elseif msg["method"] == "repl/getlazy"
-        JSONRPC.send_success_response(conn_endpoint[], msg, get_lazy(msg["params"]))
+        lazy_val = Base.invokelatest(get_lazy, msg["params"])
+        Base.invokelatest(JSONRPC.send_success_response, conn_endpoint[], msg, lazy_val)
     elseif msg["method"] == "repl/showingrid"
         try
             var = Core.eval(Main, Meta.parse(msg["params"]))
```

It's fairly hard to isolate where world-age errors can pop up, since even the communication will call functions on user objects.

I'm happy to open a PR with that change, but I'm not super convinced it's important.

davidanthoff: I think that is still too coarse :)

I think generally that we need to identify _all_ places where we are calling user provided code from our code, and then a) wrap that in a try catch block and b) use `invokelatest` for that. I think those two things correspond 1:1 with each other.

I generally think we should not wrap something like `get_variables` in a try catch block, because that would hide bugs in our own code from crash reporting, instead we should really only wrap the call to say a `show` method narrowly, and I think if we are already identifying those places for the try catch block, we might as well use the fine grained `invokelatest` in those places as well.

I think the only places that we need to handle are a) code execution and b) any `show`-like situation, right?

aviatesk: The "minimum-minimum" fix would be:
```diff
❯ git diff     
diff --git a/scripts/packages/VSCodeServer/src/VSCodeServer.jl b/scripts/packages/VSCodeServer/src/VSCodeServer.jl
index c92842a..cd5078c 100644
--- a/scripts/packages/VSCodeServer/src/VSCodeServer.jl
+++ b/scripts/packages/VSCodeServer/src/VSCodeServer.jl
@@ -53,7 +53,7 @@ function get_variables()
         startswith(n_as_string, "#") && continue
         t = typeof(x)
 
-        rendered = treerender(x)
+        rendered = Base.invokelatest(treerender, x)
 
         push!(variables, Dict(
             "type" => string(t),
@@ -204,7 +204,7 @@ function serve(args...; is_dev = false, crashreporting_pipename::Union{AbstractS
     else
         @async try
             while true
-                Base.invokelatest(handle_message; crashreporting_pipename = crashreporting_pipename)
+                handle_message(; crashreporting_pipename = crashreporting_pipename)
             end
         catch err
             Base.display_error(err, catch_backtrace())
```

But I'm still not sure we should take care of those fine cases where we need to use `Base.invokelatest`.

> I think the only places that we need to handle are a) code execution and b) any show-like situation, right?

I don't agree with this. For example, this world age error is because StaticArray.jl adds a method for `size`, which is used by our [`treerender`](https://github.com/julia-vscode/julia-vscode/blob/80c7dca2293b0cc2f421cd3f8f991856db483726/scripts/packages/VSCodeServer/src/trees.jl#L121:L121). Since `treerender` would be compiled before an user loads StaticArray.jl, we need to use `Base.invokelatest` in order to avoid that.
In general, in an interactive session, there is clear difference between "user code" and "our code", I think. We use the same running session for our interactive features and user code execution after all, so it's really impossible to avoid crashes caused by user-code, as in REPL. If s/he overloads, e.g. `print`, in a way it crashes, our interactive session also crashes. So in another word, we need to be aware that most of functions in `VSCodeServer` may accept "user-defined" objects, which may later be specialized by specific methods in user code.

Of course your approach (i.e. detect where an world age error may/can happen, and spread `Base.invokelatest` there) would be possible, but do we really need to do so ? If we wrap our toplevel handler (i.e. `handle_message`) in `Base.invokelatest`, then we can develop code without considering "this function I use in this code are likely to be overloaded by users, so I should wrap it by `Base.invokelatest`", and we need to use `Base.invokatelatest` in most of our code base, which would result in neater code. As [the docstring of `Base.invokelatest`](https://docs.julialang.org/en/v1/base/base/#Base.invokelatest) says, I think this is the perfect use case of that, no ?

I agree with we shouldn't abuse `try/catch`. They will hide an internal bug and we may oversee them otherwise.

pfitzseb: World age errors can happen every time we call user code directly or call _any_ function on an object that's introduced by the user (or some other dynamic code loading method).

For that reason we can (and will) get world age errors inside of `JSONRPC.send_success_response` when called , because that may get a user object as its argument (as will happen here, which is why @aviatesk's fix isn't quite enough when expanding objects in the workspace).

Why do you think `invokelatest` is a bad idea? Worse performance?

davidanthoff: I'm mostly a bit worried that these very high level `invokelatest` calls are not going to be enough: if new code gets evaled inside the request handler, we'll need more `invokelatest` calls after that in the request handler, I think. It is also going to be less straightforward to have these global `invokelatest` with https://github.com/julia-vscode/julia-vscode/pull/1350 (although that could probably be circumvented.

The more important aspect in my mind is to identify what we need to wrap in try catch blocks, and it just seems to me that if we do that work, we automatically will have identified any place where we might need a `invokelatest` already.

@aviatesk, you are right, calls to `size` for example also need to wrap in try catch, and I probably forgot others. But I think that is just how it is, no way around that, we just need to identify all places where we call any function with a user object, as @pfitzseb pointed out. But I don't really see why that shouldn't be possible. A bit annoying, yes, but I think worthwhile.

> For that reason we can (and will) get world age errors inside of JSONRPC.send_success_response when called , because that may get a user object as its argument

Why that? We should really only send `Dict`s today back, and in the future instances of the strongly typed response struct.

PR #1345, opened 2020-06-15
update workspace for inline eval

pfitzseb:


davidanthoff:
I don't think we need these, after all inline eval is triggered in the client, so we should be able to do whatever we want to do pre and post eval without an extra notification. That is why I had originally removed those notifications from the design. I think we _only_ need them when then user runs code from the REPL directly, i.e. when the initial interaction is not started by the extension itself.

pfitzseb:
The nice thing about this is that there only needs to be one code path on the TS site. Sure, there's two additional requests, but do we really care?

davidanthoff:
I think we probably will want to combine multiple requests together at some point: say a user presses Alt+Enter quickly in succession for three code blocks, then we would ideally only update the workspace after all three execution blocks have run, not after each one individually. It seems easier to do that if we keep this logic completely on the client, and treat the "user typed something in the REPL" case as the odd-one that needs this special handling via these notifications?

aviatesk:
ah, makes sense, those smart aggregation of requests would be fancy, I think.

davidanthoff:
Lets leave as is for now, I'll try to change this as part of https://github.com/julia-vscode/julia-vscode/pull/1350, where these notifications are a bit more tricky in any case.

PR #1350, opened 2020-06-16
Use JSONRPC msg dispatcher for REPL

davidanthoff:
~~Still has a couple of TODOs, but this migrates things to the same message dispatching framework that we use in the LS and hopefully soon in the debugger as well.~~

Ok, ready for review, it should all work. This PR also adds crash reporting for the REPL process.

### DIFF
--- a/scripts/packages/VSCodeServer/src/VSCodeServer.jl
+++ b/scripts/packages/VSCodeServer/src/VSCodeServer.jl
@@ -92,18 +92,18 @@ the following fields:
 function render(x)
     str = sprintlimited(MIME"text/plain"(), x, limit = MAX_RESULT_LENGTH)
 
-    return Dict(
-        "inline" => strlimit(first(split(str, "\n")), limit = INLINE_RESULT_LENGTH),
-        "all" => str,
-        "iserr" => false
+    return ReplRunCodeRequestReturn(
+        strlimit(first(split(str, "\n")), limit = INLINE_RESULT_LENGTH),
+        str,
+        false
     )
 end
 
 function render(::Nothing)
-    return Dict(
-        "inline" => "✓",
-        "all" => "nothing",
-        "iserr" => false
+    return ReplRunCodeRequestReturn(
+        "✓",
+        "nothing",
+        false
     )
 end
 
@@ -115,10 +115,10 @@ end
 function render(err::EvalError)
     str = sprintlimited(err.err, err.bt, func = Base.display_error, limit = MAX_RESULT_LENGTH)
 
-    return Dict(
-        "inline" => strlimit(first(split(str, "\n")), limit = INLINE_RESULT_LENGTH),
-        "all" => str,
-        "iserr" => true
+    return ReplRunCodeRequestReturn(
+        strlimit(first(split(str, "\n")), limit = INLINE_RESULT_LENGTH),
+        str,
+        true
     )
 end
 """
@@ -128,11 +128,15 @@ Calls `render`, but catches errors in the display system.
 """
 function safe_render(x)
     try
-        render(x)
+        return render(x)
     catch err
         out = render(EvalError(err, catch_backtrace()))
-        out["inline"] = string("Display Error: ", out["inline"])
-        out["all"] = string("Display Error: ", out["all"])
+
+        return ReplRunCodeRequestReturn(
+            string("Display Error: ", out.inline),
+            string("Display Error: ", out.all),
+            out.iserr = out.iserr
+        )
     end
 end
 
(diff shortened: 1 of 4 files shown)


scripts/packages/VSCodeServer/src/VSCodeServer.jl
```diff
@@ -193,126 +197,181 @@ function serve(args...; is_dev = false, crashreporting_pipename::Union{AbstractS
...
+        # Add crash reporting
+        Base.display_error(err, catch_backtrace())
     end
 end
 
-function handle_message(; crashreporting_pipename::Union{AbstractString,Nothing}=nothing)
-    msg = JSONRPC.get_next_message(conn_endpoint[])
-    if msg["method"] == "repl/runcode"
-        params = msg["params"]
+JSONRPC.@dict_readable struct ReplRunCodeRequestParams <: JSONRPC.Outbound
```
pfitzseb, line 230: Do you think it would make sense to automatically add a kwarg constructor here? IMHO it's very convenient to not have to remember argument order.

pfitzseb: Heh, nvm: https://github.com/julia-vscode/JSONRPC.jl/issues/20.

scripts/packages/VSCodeServer/src/repl_protocol.jl
```diff
@@ -0,0 +1,35 @@
...
+JSONRPC.@dict_readable struct ReplRunCodeRequestReturn <: JSONRPC.Outbound
+    inline::String
+    all::String
+    iserr::Bool
+end
+
+const repl_runcode_request_type = JSONRPC.RequestType("repl/runcode", ReplRunCodeRequestParams, ReplRunCodeRequestReturn)
+const repl_getvariables_request_type = JSONRPC.RequestType("repl/getvariables", Nothing, Vector{ReplGetVariablesRequestReturn})
+# TODO Return type
+const repl_getlazy_request_type = JSONRPC.RequestType("repl/getlazy", Int, Vector{Any})
```
davidanthoff, line 31: @pfitzseb could you try to create the types that we need for the return value here? I didn't fully understand the structure there, and was also not sure about good names. Essentially right now there are some dicts returned, and ideally we would create a type for those instead. Could you maybe open a PR with that against this branch?

src/interactive/workspace.ts
```diff
@@ -21,7 +21,7 @@ const requestTypeGetVariables = new rpc.RequestType<
     void, void>('repl/getvariables')
 
 const requestTypeGetLazy = new rpc.RequestType<
-    void,
+    number,
```
davidanthoff, line 24: @pfitzseb I was not 100% certain, but it seemed this should be number? If so, then this was just an unrelated bug that was also fixed :)

scripts/packages/VSCodeServer/src/repl_protocol.jl
davidanthoff: Actually, thinking about this a bit more: why don't we merge this PR here (it works), and then you can open a new PR against `master` at some point that uses static types for everything?

pfitzseb: Yeah, let's do that.

src/interactive/workspace.ts
pfitzseb: Huh, indeed. Weird that TS didn't complain about this.

aviatesk: That's kinda expected, since `id` filed in `WorkspaceVariable` is type annotated as `any`:
https://github.com/julia-vscode/julia-vscode/blob/5be700c8256e5e2cf66e5f93a131fcecc2f1b096/src/interactive/workspace.ts#L11

So if we change this to `number`, it would make sense to change that as well.

aviatesk:
@davidanthoff do you think we can include this in v0.16 ? Enabling crash report would be might be worth to take a risk.

src/interactive/workspace.ts
pfitzseb: Ok, no idea why I did that :)

scripts/packages/VSCodeServer/src/VSCodeServer.jl
```diff
@@ -193,126 +201,146 @@ function serve(args...; is_dev = false, crashreporting_pipename::Union{AbstractS
...
+    @async try
+        msg_dispatcher = JSONRPC.MsgDispatcher()
+
+        msg_dispatcher[repl_runcode_request_type] = repl_runcode_request
+        msg_dispatcher[repl_getvariables_request_type] = repl_getvariables_request
+        msg_dispatcher[repl_getlazy_request_type] = repl_getlazy_request
+        msg_dispatcher[repl_showingrid_notification_type] = repl_showingrid_notification
+        msg_dispatcher[repl_loadedModules_request_type] = repl_loadedModules_request
+        msg_dispatcher[repl_isModuleLoaded_request_type] = repl_isModuleLoaded_request
+        msg_dispatcher[repl_startdebugger_notification_type] = (conn, params)->repl_startdebugger_request(conn, params, crashreporting_pipename)
```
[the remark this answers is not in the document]
davidanthoff: I'm generally weary of global stuff, and have a hunch that it might be easier to write tests for all of this if we have less global state. But I'm not sure, to be honest... I don't think the startup cost would make much of difference: we would save adding like a handful of items to a dict, that can't take much time, right?

davidanthoff:
> do you think we can include this in v0.16 ? Enabling crash report would be might be worth to take a risk.

I'd say we ship it in the first patch release after v0.16? That can literally be a day after the official release, but right now I'd really like to "stick out" our test window for the insider build and not reset, given that we are three days in :)

scripts/packages/VSCodeServer/src/VSCodeServer.jl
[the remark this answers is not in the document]
pfitzseb: Yeah, let's keep these here.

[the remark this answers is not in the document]
aviatesk: yeah, now I think there is no strong motivation to make them global. sorry for the noise.

davidanthoff:
Ok, I'm going to interpret all the reviews and thumbs up as "approved" and will merge right now :) I want to get a new insider build out ASAP, and this is a biggy to get in.

PR #1351, opened 2020-06-17
fix display for zero dimensional array

marius311:
Without this, the following code errors and in fact hangs all subsequent vscode inline printing:

```julia
struct ZeroDimArray <: AbstractArray{Float64,0} end
Base.size(::ZeroDimArray) = ()
Base.print_array(io::IO, ::ZeroDimArray) = nothing

ZeroDimArray() # displaying this causes error
```

An edge case no-doubt but this is a type of object I actually have, plus it doesn't error in the REPL or Juno. 

This fixes this particular case, but its kind of worrisome that it crashed all subsequent inline printing. Is this a general problem that an error in printing isn't recoverable? (and can that be fixed?)

### DIFF
--- a/scripts/packages/VSCodeServer/src/trees.jl
+++ b/scripts/packages/VSCodeServer/src/trees.jl
@@ -31,6 +31,7 @@ const MAX_PARTITION_LENGTH = 20
 treeid() = (ID[] += 1)
 
 pluralize(n::Int, one, more = one) = string(n, " ", n == 1 ? one : more)
+pluralize(n::Tuple{}, one, more = one) = ""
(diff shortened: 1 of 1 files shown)


davidanthoff:
I think we should merge this as soon as @pfitzseb has reviewed it, but not reset our v0.16 release cycle for this, but then just push this out with the first patch release.

In general, we want the extension to crash if things go wrong so that we receive a crash report and can fix things. Currently we haven't hooked up crash reporting for the REPL part, but the PR that has that is pending.

pfitzseb:
I have an alternative in https://github.com/julia-vscode/julia-vscode/pull/1353 that the printing still makes sense:
image

davidanthoff:
Ok, then I'll close this one here, right?

PR #1355, opened 2020-06-18
change execution path correctly:

aviatesk:
Currently changing `julia.executablePath` won't work correctly, maybe because `g_settings` doesn't get updated correctly. We can correctly subscribe to the change of the configuration by watching `vscode.ConfigurationChangeEvent.affectsConfiguration`.

With this PR:
- now we can change `julia.executablePath` more than one time in a session correctly
- less global, happier we would be
- each sub feature scripts can simply subscribe to `onDidChangeConfig`, and so there is no need to maintain settings.ts (deleted)


src/extension.ts
```diff
@@ -92,14 +88,11 @@ function setLanguageClient(languageClient: vslc.LanguageClient = null) {
...
 export const onDidChangeConfig = g_onDidChangeConfig.event
-function changeConfig(params: vscode.ConfigurationChangeEvent) {
-    const newSettings = settings.loadSettings()
-    g_onDidChangeConfig.fire(newSettings)
-
-    const need_to_restart_server = g_settings.juliaExePath !== newSettings.juliaExePath ? true : false
-    if (need_to_restart_server) {
+function changeConfig(event: vscode.ConfigurationChangeEvent) {
+    g_onDidChangeConfig.fire(event)
+    if (event.affectsConfiguration('julia.executablePath')) {
```
pfitzseb, line 95: Should this also restart the REPL session?

src/jlpkgenv.ts
```diff
@@ -188,11 +187,11 @@ export async function getEnvName() {
     return path.basename(envpath)
 }
 
-export async function activate(context: vscode.ExtensionContext, settings: settings.ISettings) {
+export async function activate(context: vscode.ExtensionContext) {
     context.subscriptions.push(onSetLanguageClient(languageClient => {
         g_languageClient = languageClient
     }))
-    context.subscriptions.push(onDidChangeConfig(newSettings => { }))
+    context.subscriptions.push(onDidChangeConfig(event => { }))
```
pfitzseb, line 194: Just delete this?

aviatesk: yeah, that would be better, will do.

src/extension.ts
aviatesk: hmm, killing/starting REPL session sounds to me like a big deal, so I guess some users will complain on that, maybe ? Not sure tbh, though.

pfitzseb: True, I wouldn't want that change to mess up my current session.

PR #1357, opened 2020-06-18
LS startup frontend improvements:

aviatesk:
- clear start up notification on LS crash (currently it remains forever in status bar on LS crashes)
- click to reveal LS output channel


src/extension.ts
```diff
@@ -193,18 +193,24 @@ async function startLanguageServer() {
         }
     })
 
-    // Push the disposable to the context's subscriptions so that the
-    // client can be deactivated on extension deactivation
+    const disposable = vscode.commands.registerCommand('language-julia.showLanguageServerOutput', () => {
+        languageClient.outputChannel.show(true)
+    })
```
davidanthoff, line 198: Why not just do the normal `g_context.subscriptions.push` on this command? Yes, then the command is "around" for the lifetime of the extension, but that doesn't seem too bad and it would mean we don't have to do manual cleanup.

I'm also not entirely sure whether the code construction below will actually reliably clean-up the object, so by not using the `subscriptions` cleanup method, we might be introducing a memory leak. Here is the scenario: everything in the try part of the block below runs successfully, but then something async goes wrong and then `onReady` promise never gets successfully resolved. I _don't_ think the `catch` block will run in that case, and in that situation we would have a memory look for the command, right?

aviatesk: > Here is the scenario: everything in the try part of the block below runs successfully, but then something async goes wrong and then onReady promise never gets successfully resolved.

ah, yeah, we need to use `finally` clause instead.

aviatesk: > Registering a command with an existing command identifier twice will cause an error.

so we need to manually dispose the command if we use the same identifier for this command.

davidanthoff: Wouldn't it be easier to just register the command once in the `activate` function, and then let it be cleaned up by the normal subscription method? But also happy to merge this as if, you decide :)

aviatesk:
Okay let me merge this as is -- let's register the command into the global subscription when we come to want to expose this command to users (and so we want to use `g_languageClient` rather than local `languageClient`)

PR #1361, opened 2020-06-18
Move debugger into package

davidanthoff:
This moves the debugger into a package, and generally follows the structure for private loading that we've been using throughout the extension now.

This should generally clean things up and prepare us to move the DebugAdapter.jl package into a "proper" Julia package that could be used by other users.

This should also precompile the debugger for the F5 scenario, like @aviatesk introduced for the REPL process, which will hopefully speed things up.

Ideally we would merge this rather sooner than later because this PR includes a submodule update, and those are tricky to handle with merge conflicts.


davidanthoff:
Alright, I'm going to break all the rules here and just merge this without review. Please _do_ review after the fact, so that we can still fix any problems you see in follow up PRs.

I'm doing this because a PR with a submodule update is essentially blocking any other updates, and I'd like to update the other submodules and push out a new insider build tonight.

aviatesk:
I don't think this PR doesn't precompile `VSCodeDebugger` package, since it doesn't have any Project.toml and etc. It's still loaded as usual scripts.
But I'm still not sure when the module is used -- could you tell me how it is used in vscode ?

davidanthoff:
Ah, you are right: https://github.com/julia-vscode/julia-vscode/pull/1375.

`VSCodeDebugger` only gets used when a user starts the F5 debugger, i.e. when the user starts a new debuggee process, and is not using the debugger from the REPL. In that situation, `VSCodeDebugger` is playing a similar role to `VSCodeServer` in the REPL process: it is the top level package that we load into the debuggee process that is handling the debugger communication.

PR #1362, opened 2020-06-19
Use VS Code mechanism to start debuggee

davidanthoff:
This would significantly simplify how we start the debuggee when someone presses F5, but is currently blocked by https://github.com/microsoft/vscode/issues/100538 (unless we just want to live with the ugly UI for now).


src/juliaDebug.ts
```diff
@@ -266,48 +261,35 @@ export class JuliaDebugSession extends LoggingDebugSession {
...
+	            // kind: 'external',
+	            title: 'Julia Debugger',
+	            cwd: args.cwd,
+	            args: [
+	                this._juliaPath,
+	                '--color=yes',
+	                '--history-file=no',
+	                '--startup-file=no',
+	                `--project=${args.juliaEnv}`,
+	                join(this._context.extensionPath, 'scripts', 'debugger', 'run_debugger.jl'),
```
MariusDrulea, line 284: as I reported in other comments, having the `launch_wrapper.jl` file allows us to add custom options when we start de debugger, for instance debugging with a sysimage (julia --sysimage path_to_sysimage). With the current solution we are limited to the options presented here (color, history-file etc). We shall have this `launch_wrapper.jl`  back or we shall have a way to specify the extra arguments to julia.

MariusDrulea: https://github.com/julia-vscode/julia-vscode/issues/2552

PR #1363, opened 2020-06-19
Further split VSCodeServer into individual files

davidanthoff:
Just splits a gigantic file into multiple smaller ones, i.e. purely code moving around, nothing else.


scripts/packages/VSCodeServer/src/misc.jl
```diff
@@ -14,7 +14,7 @@ function Base.write(io::LimitIO, v::UInt8)
     io.n += write(io.io, v)
 end
 
-function sprintlimited(args...; func = show, limit::Int = 30, ellipsis::AbstractString = "…", color = false)
+function sprintlimited(args...; func=show, limit::Int=30, ellipsis::AbstractString="…", color=false)
```
aviatesk, line 17: Do we prefer not having whitespaces around arguments ?

scripts/packages/VSCodeServer/src/request_handlers.jl
```diff
@@ -0,0 +1,143 @@
...
+function repl_startdebugger_request(conn, params::String, crashreporting_pipename)
+    hideprompt() do
+        debug_pipename = params
+        try
+            DebugAdapter.startdebug(debug_pipename)
+        catch err
+            DebugAdapter.global_err_handler(err, catch_backtrace(), crashreporting_pipename, "Debugger")
+        end
+    end
+end
```
aviatesk, line 143: How about moving each request handler so that they live close to its logic, rather than aggregating them here ?
E.g.
- `repl_loadedModules_request` into module.jl
- `repl_getvariables_request` into trees.jl (which actually I prefer to rename to workspace.jl or something if we go this way)

In other word we will organize files based on feature rather than internal design. Juno also organizes handlers in this way: e.g.
https://github.com/JunoLab/Atom.jl/blob/75eebb773ec222368801968ff5d6ae1e083c4fbf/src/workspace.jl#L13

scripts/packages/VSCodeServer/src/misc.jl
davidanthoff: DocumentFormat.jl does, but I think we might just have shipped a version where we can configure that? Not sure, if we want to change, we should probably change for the whole code base in a separate PR? This code formatting was done automatically at save.

scripts/packages/VSCodeServer/src/request_handlers.jl
davidanthoff: Yes, I think that would be even nicer. I'd say in a follow up PR?

aviatesk: Yeah, maybe I will.

aviatesk:
ops sorry I missed your review on this.

aviatesk:
@pfitzseb ops sorry I missed your review on this.

PR #1366, opened 2020-06-19
Simplify the search for Julia binary

davidanthoff:
This changes how we search for the Julia binary and hopefully makes it simpler. Highlights:
- we no longer try to find out an absolute path for the Julia binary. If `julia` is enough to start Julia, then we just use that as the binary path. Technical that makes the name of the config setting a misnomer now: it should probably be `juliaCommand` or something like that. But I'd say we just leave as is.
- it reduces the number of times Julia is started during the detection phase, because we no longer start it to get the install path, but only once start it with `--version`. That should speed up this phase quite a bit.
- I think there was a situation in the old code where an exception could leave this part, and I think that explains why we sometimes get these "command not registered" errors: an exception in the detection phase breaks somewhere in `activate`, and not everything actually runs in that function.
- I removed the use of `null` and went back to `undefined`. Two reasons: the MS style guides for TypeScript blanket say one should use `undefined`, not `null`. Also, the various null coalescing operators in TypeScript return `undefined`, so it makes code simpler.
- we no longer need a dep on the `which` package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -752,8 +752,7 @@
         "uuidv4": "^6.1.0",
         "vscode-debugadapter": "^1.41.0",
         "vscode-jsonrpc": "^5.0.1",
-        "vscode-languageclient": "^6.1.3",
-        "which": "^1.3.1"
+        "vscode-languageclient": "^6.1.3"
     },
     "devDependencies": {
         "@types/cson-parser": "^4.0.4",
--- a/src/juliaexepath.ts
+++ b/src/juliaexepath.ts
@@ -1,38 +1,43 @@
 import { exec } from 'child-process-promise'
-import * as child_process from 'child_process'
 import * as os from 'os'
 import * as path from 'path'
 import * as process from 'process'
-import * as util from 'util'
 import * as vscode from 'vscode'
-import * as which from 'which'
 import { onDidChangeConfig } from './extension'
 import { setCurrentJuliaVersion, traceEvent } from './telemetry'
-const whichAsync = util.promisify(which)
 
-let actualJuliaExePath: string = null
+let g_actualJuliaExePath: string | undefined = undefined
 
-async function setNewJuliaExePath(newPath: string) {
-    actualJuliaExePath = newPath
+async function trySetNewJuliaExePath(newPath: string) {
+    if (newPath !== g_actualJuliaExePath) {
+        try {
+            const queriedVersion = (await exec(`"${newPath}" --version`)).stdout.trim()
 
-    child_process.exec(`"${newPath}" --version`, (error, stdout, stderr) => {
-        if (error) {
-            return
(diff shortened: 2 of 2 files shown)


davidanthoff:
@ZacLN, you are on Linux, right? So I tried this on Windows, you presumably on Linux, might be good if someone also tried this on a Mac. @pfitzseb  or @aviatesk, is either of you on a Mac? If not, I have a Mac in the office that I can remote into, but it is a bit of a pain...

ZacLN:
Yep, it seemed to work on linux

davidanthoff:
Hm, interesting... If you just use `julia`, how is your system configured to resolve that? Do you have the julia folder added to your `PATH`? Or something in your shell startup script that resolves this? Or something else?

pfitzseb:
What's the reason for
> we no longer try to find out an absolute path for the Julia binary. If julia is enough to start Julia, then we just use that as the binary path

Finding the absolute path seems like the more robust approach, no?

aviatesk:
> Hm, interesting... If you just use julia, how is your system configured to resolve that? Do you have the julia folder added to your PATH? Or something in your shell startup script that resolves this? Or something else?

I added `/usr/local/bin/` in my `PATH` and `/usr/local/bin/julia` is used for `julia`. IIRC this is a normal setup for macOS:
https://julialang.org/downloads/platform/#macos

davidanthoff:
> Finding the absolute path seems like the more robust approach, no?

Multiple reasons:
1. We found the absolute path by running `await exec(`"${p}" --startup-file=no --history-file=no -e "println(Sys.BINDIR)"`)`, which is fairly slow. This PR here just starts julia with the `--version` argument, which is way faster.
2. The old approach started Julia twice in the detection phase: once to get the version and once to get the absolute path, also slow (granted, that could pretty easily be combined into one).
3. I'm working on a MSIX installer for Julia, and the old approach doesn't work with that, but the new one does :) This part is pretty speculative at this point and there might be other ways to solve this. The problem is that `julia` on the `PATH` could be some wrapper program/script that actually starts the "real" julia binary. If we report the path of the Julia binary from `Sys.BINDIR` and then use that from the extension, we are essentially bypassing any such wrapper launcher/script.

@aviatesk's scenario obviously just needs to work. I'll try to dig a bit and see whether I can get around that, but if not, then we probably shouldn't merge this as is.

pfitzseb:
Oh, I thought you were using `which`/`where` to ask the OS what `julia` (or some other alias) resolves to (if not, why did we depend on the `which` npm package?). Starting multiple Julia processes is a bad idea though, agreed.

davidanthoff:
We used some duplicate mix of using the `which` package and starting Julia :) Maybe I'll just try to only use `which`...

PR #1368, opened 2020-06-19
Improve WS code and types

pfitzseb:


src/interactive/workspace.ts
```diff
@@ -22,15 +22,7 @@ const requestTypeGetVariables = new rpc.RequestType<
...
-    {
-        lazy: boolean,
-        id: number,
-        head: string,
-        haschildren: boolean,
-        value: string,
-        canshow: boolean,
-        icon: string
-    }[],
+    WorkspaceVariable[],
```
aviatesk, line 25: ops, this is much much better :O

aviatesk:
Approved, but now we have another conflict ...

pfitzseb:
Fixed that.

PR #1370, opened 2020-06-19
stop inline eval from force-showing the REPL

pfitzseb:


src/interactive/repl.ts
```diff
@@ -401,7 +401,7 @@ async function executeCodeCopyPaste(text, individualLine) {
         text = text + '\n'
     }
 
-    await startREPL(true)
+    await startREPL(true, false)
```
aviatesk, line 404: This should be `true` since we need to look into REPL for this case ?

pfitzseb: True.

davidanthoff:
Hm, I think given that we have the default config value for inline eval results as off, we shouldn't really make the default to not show the REPL. That would mean that for most users, if they press Alt+Enter, nothing shows, right?

I think we could either a) add a config value for this, or b) try to come up with a "clever" logic here. For example: only if the inline show config is set to "inline only" (i.e. no display in the REPL) would we not show the REPL, or something like that. Or maybe the REPL always shows when it is started, but if the user has hid it (i.e. if it already exists when a code block is executed), it is not shown again, or something like that.

pfitzseb:
> Or maybe the REPL always shows when it is started, but if the user has hid it (i.e. if it already exists when a code block is executed), it is not shown again, or something like that.

That's how this works.

PR #1371, opened 2020-06-19
reorganize functions more

aviatesk:
follows up #1363  -- just moves code around, but now each request handler lives closer to its logic.


scripts/packages/VSCodeServer/src/debug_cmds.jl
```diff
@@ -1,3 +1,16 @@
...
+    hideprompt() do
+        debug_pipename = params
+        try
+            DebugAdapter.startdebug(debug_pipename)
+        catch err
+            DebugAdapter.global_err_handler(err, catch_backtrace(), crashreporting_pipename, "Debugger")
+        end
+    end
+end
+
```
aviatesk, line 13: Not these functions is better to live here ... Maybe renaming `debug_cmd.jl` into `debugger.jl` or something would make more sense ?

scripts/packages/VSCodeServer/src/VSCodeServer.jl
```diff
@@ -48,12 +48,10 @@ include("../../../error_handler.jl")
 include("misc.jl")
 include("trees.jl")
 include("repl.jl")
-include("gridviewer.jl")
-include("repl_protocol.jl")
```
davidanthoff, line 52: I think we should keep the `repl_protocol.jl` file and its content as is. That pattern is used throughout the VS Code and our code base, i.e. that we have one file that defines the interface/protocol structs and message types, so that one can find them easily in one place.

So maybe make this PR such that it only moves the actual handler functions into individual files?

scripts/packages/VSCodeServer/src/debug_cmds.jl
davidanthoff: Yes, agreed!

davidanthoff:
Oh, and there is now a merge conflict :)

scripts/packages/VSCodeServer/src/VSCodeServer.jl
aviatesk: huh, sounds fair. I personally prefer to keep protocol/handler to live closer to each other, but the consistency with other parts in our code base would be more important ;)
Will revert back the change.

aviatesk:
I think I've addressed @davidanthoff 's request. 
So now this PR essentially does:
- move each handler in `request_handlers.jl` into each feature files
- reneme `debug_cmd.jl` into `debugger.jl`

PR #1374, opened 2020-06-19
fix WS regression for toplevel module

aviatesk:
follows up #1368 , currently we lacks the names for toplevel module


aviatesk:
Going to merge this since this is just a small bug fix.

PR #1375, opened 2020-06-19
Add two Project.toml files so that things precompile

davidanthoff:
With this the F5 debugger _really_ gets precompiled.

### DIFF
--- a/scripts/packages/DebugAdapter/Project.toml
+++ b/scripts/packages/DebugAdapter/Project.toml
@@ -1,0 +1,16 @@
+name = "DebugAdapter"
+uuid = "b9dc54e9-555d-4545-8d72-fb33852127ad"
+version = "0.1.0"
+
+[deps]
+JuliaInterpreter = "aa1ae85d-cabe-5617-a682-6adf51b2e16a"
+
+[extras]
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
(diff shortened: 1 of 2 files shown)


scripts/packages/DebugAdapter/Project.toml
```diff
@@ -0,0 +1,16 @@
+name = "DebugAdapter"
+uuid = "b9dc54e9-555d-4545-8d72-fb33852127ad"
+version = "0.1.0"
+
+[deps]
+JuliaInterpreter = "aa1ae85d-cabe-5617-a682-6adf51b2e16a"
```
aviatesk, line 6: This will load the "real" version of JuliaInterpreter.jl (i.e. the one registered into General), no ? 
I guess you need to checkin Manifest.toml for this package ?

davidanthoff: Actually, I think I didn't have to add this file at all because it will never be used in our setting: DebugAdapter is only being loaded via `include` in our extension, never via the normal package load mechanism. The one that really matters in the `Project.toml` for `VSCodeDebugger`. My next PR is actually moving DebugAdapter into a proper external package in any case, so this file will have a short lifetime no matter what ;)

aviatesk: okay, let's focus on the PR.

PR #1378, opened 2020-06-19
Add support for profiler results viewing

davidanthoff:
Right now it saves a `.cpuprofile` file, which can be natively opened in VS Code.

I still hope to be able to implement this without us writing a file to disc, but that is currently blocked by https://github.com/microsoft/vscode/issues/100612. Much of the code in this PR is in preparation for such a file-less implementation.

UPDATE: We'll do the first version with the file based approach, and as soon as VS Code supports the more sophisticated UI we'll use that. I've left a bunch of code in this PR that we'll only need once we move away from the file based solution. It shouldn't hurt for now, and easier than extracting it into a new PR, so I'd say we just merge it and let it hang around until we change things again here.

This does export two new things: `view_profile` and `@profview`. Are we ok with those APIs?


davidanthoff:
Oh, and in terms of merging: this clearly warrants mention in the release notes, but I think we could also ship it silently in a v0.16 build and then just mention it for the v0.17 release?

pfitzseb:
I hope you don't mind, but I've pushed a commit that automatically opens the saved `*.cpuprofile`in VSCode. Also changed the default path to Julia's `tempfile()` so we don't pollute `cwd`.

pfitzseb:
There's also something slightly wonky with the `cpuprofile`s we generate, but I haven't looked into that yet. 

One thing we should do is prune the profile so our internals don't show up. Would be nice to use e.g. `FlameGraphs.prunebranch!`/`FlameGraphs.graftchildren!`, but that's too big a dependency to take on, I think.

davidanthoff:
> There's also something slightly wonky with the `cpuprofile`s we generate, but I haven't looked into that yet.

@vchuravy implemented that conversion. We should probably try to get to the bottom of any problems there ASAP, because we can't really ship this feature if the conversion is not correct... @vchuravy, any chance that you can help us out there? We do have a nominal feature freeze on 7/1, anything that doesn't make it in by then will have to wait until after Juliacon :(

davidanthoff:
We might have used an outdated version of ChromeProfileFormat.jl in this PR, not entirely sure. I updated it now, make sure to do a `git submodule update`, and then we should maybe test it again and see whether the output has problems or not.

pfitzseb:
Things look more correct now. The flame graph is still very different from what ProvileView/Juno offers, because it has the time on the x axis.

davidanthoff:
I guess the question is whether it is good enough to ship?

pfitzseb:
AFAICT the output produced by ChromeProfileFormat.jl is correct. 

The flamegraph produced by vscode-js-profile-visualizer also looks correct, while the table is just plain wrong. It would also be cool to have a "left-heavy" mode like on https://www.speedscope.app/, which is basically what we are used to in Julia-land.

I think we can ship this as an experimental feature and hope that vscode-js-profile-visualizer improves soon.

pfitzseb:
The table display is just plain wrong, I think. Click on the small flame icon in the upper right -- that opens a much more useful flamegraph.

aviatesk:
agh, I misunderstood the extension is now bundled with vscode by default.

davidanthoff:
> The table display is just plain wrong

Did you open an issue with MS about this? That is something they need to fix, right?

jeffreyesun:
Would there be a way to reintroduce the table view? I'm not totally satisfied with completely removing the ability to view times by function in this way.

pfitzseb:
Sure, as soon as someone has time to do that (which I don't currently). Also see https://github.com/pfitzseb/ProfileCanvas.jl/issues/3.

PR #1380, opened 2020-06-20
Settings - resouce scope for formatting options

fonsp:
This changes the `"scope"` of formatting options from `"window"` to `"resource"`.

This means that repositories can _fix these settings for a repository_ by saving these settings inside `[git-root]/.vscode/settings` (normally done through UI, see screenshot). This helps open source: people who contribute to a project can call `Julia extension -> Format` on code they write, and it will match the maintainers' preferred style (and not modify other parts of the document).

(I learned about these values from [prettier-vscode](https://github.com/prettier/prettier-vscode/blob/v5.1.0/package.json#L188-L319))

Screenshot showing that these settings are currently not available at the folder/repository level:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -509,61 +509,61 @@
                     "type": "integer",
                     "default": 4,
                     "description": "Indent size for formatting.",
-                    "scope": "window"
+                    "scope": "resource"
                 },
                 "julia.format.indents": {
                     "type": "boolean",
                     "default": true,
                     "description": "Format file indents.",
-                    "scope": "window"
+                    "scope": "resource"
                 },
                 "julia.format.ops": {
                     "type": "boolean",
                     "default": true,
                     "description": "Format whitespace around operators.",
-                    "scope": "window"
+                    "scope": "resource"
                 },
                 "julia.format.tuples": {
                     "type": "boolean",
                     "default": true,
                     "description": "Format tuples.",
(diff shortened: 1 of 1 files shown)


davidanthoff:
In principle I'd be in favor of this, but we would have to support URI specific config settings in the language server before we can merge this here, and that doesn't work yet. 

I think you should already be able to store a settings file in the root of git repo, and if you open the root of the repo in VS Code the settings should apply (because we support workspace level settings). The change here would (if I understand it correctly) allow a user to use a different setting for different folders in a workspace.

fonsp:
Okay, thanks for your response!

Workspace level settings are supported right now (they are saved in the `.code-workspace` file), but a VS Code Workspace contains (multiple) repositories, given by paths on the user's machine. I think that this is why you can also add settings to the `.vscode` folder - to sync them on a repository. 

I understand the URI problem, feel free to close this PR!

fonsp:
That's a shame, it would be really nice if contributors would automatically use the same formatting settings when they open our repository in VS Code! We have this set up for our JS codebase and it works great :)

pfitzseb:
Ah, but they can, if you put a `.JuliaFormatter.toml` into your repo!

fonsp:
Ooooh nice, thanks!

PR #1388, opened 2020-06-22
Guard against user exceptions and world age errors

davidanthoff:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1338.

@aviatesk I think this is the kind of code pattern we should use to a) guard against errors in user code and b) solve our world age error problems. I _do_ think we need these very narrow `try catch` blocks no matter what to harden the REPL server against errors in user code. And I think these `try catch` blocks should be as narrow as in this example, so that they _really_ only catch errors in user code, but not errors in our code. If we do that already, we might as well put the `invokelatest` call at that location as well, right?

Unclear to me what the correct response here is if we end up in the `catch` block. Show some error node in the tree, probably?

### DIFF
--- a/scripts/packages/VSCodeServer/src/trees.jl
+++ b/scripts/packages/VSCodeServer/src/trees.jl
@@ -122,9 +122,21 @@ function treerender(x::Module)
 end
 
 function treerender(x::AbstractArray{T,N}) where {T,N}
-    treerender(LazyTree(string(typeof(x), " with $(pluralize(size(x), "element", "elements"))"), wsicon(x), length(x) == 0, function ()
-        if length(x) > MAX_PARTITION_LENGTH
-            partition_by_keys(x, sz = MAX_PARTITION_LENGTH)
+    size_of_x = try
+        Base.invokelatest(size, x)
+    catch err
+        # TODO Gracefully handle this error in user code
+    end
+
+    length_of_x = try
+        Base.invokelatest(length, x)
+    catch err
+        # TODO Gracefully handle this error in user code
+    end
+
+    treerender(LazyTree(string(typeof(x), " with $(pluralize(size_of_x, "element", "elements"))"), wsicon(x), length_of_x == 0, function ()
+        if length_of_x > MAX_PARTITION_LENGTH
+            partition_by_keys(x, sz=MAX_PARTITION_LENGTH)
         else
             [SubTree(repr(k), wsicon(v), v) for (k, v) in zip(keys(x), vec(x))]
         end


pfitzseb:
Alright, let's try this approach -- I can see why it's a good idea, even if slightly annoying to implement :)

And yes, we should show the error in the tree view whenever user code fails.

davidanthoff:
@pfitzseb next step here is that you push something to this branch that replaces the TODOs, right?

davidanthoff:
I'm also labeling this a showstopper because this bug is in the shipping version of the extension and makes the REPL almost unusable for folks, I believe.

pfitzseb:
Ok, this doesn't work. I'll push a commit that gets us half way there, but it's crazy to have to wrap every primitive into a `try/catch` and an `invokelatest`. 

The original concern against putting an `invokelatest` into the message handler was that that might fail after evaluating user code. That isn't a concern in this case because we control *all* entrypoints, and as such can wrap each of them in an `invokelatest`.

davidanthoff:
> Ok, this doesn't work.

Why not? I mean, yes, it is cumbersome, but we do need these `try catch` blocks, right? Fully on board to not add those all at once, and for now just improve the status quo incrementally.

pfitzseb:
Let me take another look at this tomorrow. The commit I pushed should get us pretty far, but there still are world age errors in `iterate` somewhere. I'd really dislike wrapping every loop in an `invokelatest`, but _maybe_ [this `collect` thing](https://github.com/julia-vscode/julia-vscode/pull/1411#discussion_r446384437) (or rather the lack thereof) is the issue.

davidanthoff:
> Just putting the handler into a try-catch (we're already doing that) gives the user the same info and still shouldn't break anything.

We need to distinguish between errors that should send a crash report (because they are ours) and errors that we need to ignore. I don't see how we could do that with a try catch block at the handler level.

aviatesk:
I firmly believe requiring us to always think about the need to wrap a piece of code into `try/catch` and add guards would be a sheer pain ...
It's highly possible the guards themselves will be a source of the errors, and then it doesn't really make sense imho.

I think there're two ways to handle this:
1. hack into a stacktrace: if the top frame in a stackfrace is in `VSCodeServer` files, then we will think of it as "our error" and report. That would be not so robust heuristic of course, but should cover most of the cases, I think
2. simply give up the error report for this REPL code: then the only thing we should do is just put a one line of `invokelatest`

davidanthoff:
Yes, it is a pain, but I think that is just what it is for infrastructure code like ours that interacts with user code and wants to be robust in the face of that. We went through exactly the same thing with SymbolServer.jl (same story with world age problems there) and yes, it was painful to identify the places where we interact with user code, and where user code can crash us, and it took time, but in the end we got a more robust code base out of that exercise.

I feel _very_ strongly that we should not take away crash reporting. That is just way too valuable to disable.

I think our strategy should be this: a) we ship often and fast, b) we don't have high level try catch blocks that hide errors (i.e. we fail fast), and c) we put `invoke_latest` calls narrowly into the codebase. That is essentially what we did with SymbolServer.jl, and if the experience with that is any guide, we'll be able to identify and fix something like 90% of the problematic cases really quickly that way and put in place the necessary guards.

PR #1409, opened 2020-06-26
Run more tests as part of our CI

davidanthoff:


davidanthoff:
I'm also just going to merge as this doesn't touch any production code, just our CI testing story.

aviatesk:
Why do we need `testenvironments` folder ? I don't see the reason for the indirection.

davidanthoff:
For the test of `DebugAdapter` we need to make sure they use the local package versions in the repo and not something that is downloaded from the repository, so we need a Manifest to specify that. For the other two I guess we could in theory activate the package itself, but doesn't that create a Manifest.toml?

aviatesk:
```
pkg> activate VSCodeServer
pkg> test
```
will create Manifest.toml in the root of VSCodeServer, and it would work as in an usual package.

davidanthoff:
Yeah, and I don't really like these `Manifest.toml` files in package folders, I'm always worried that they have some unintended side-effect later on. I think generally we should try hard to have our extension folder be as much read-only as possible (to fully get there we should at some point move the depot for the LS out of it as well).

PR #1411, opened 2020-06-26
fix treeview world age problems again

pfitzseb:
Alternative to https://github.com/julia-vscode/julia-vscode/pull/1388.

### DIFF
--- a/scripts/packages/VSCodeServer/src/trees.jl
+++ b/scripts/packages/VSCodeServer/src/trees.jl
@@ -87,7 +87,7 @@ function treerender(x)
         treerender(Text(string(typeof(x), "()")))
     else
         treerender(LazyTree(string(typeof(x)), wsicon(x), function ()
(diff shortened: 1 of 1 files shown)


scripts/packages/VSCodeServer/src/trees.jl
```diff
@@ -126,7 +126,7 @@ function treerender(x::AbstractArray{T,N}) where {T,N}
         if length(x) > MAX_PARTITION_LENGTH
             partition_by_keys(x, sz = MAX_PARTITION_LENGTH)
         else
-            [SubTree(repr(k), wsicon(v), v) for (k, v) in zip(keys(x), vec(x))]
+            collect([SubTree(repr(k), wsicon(v), v) for (k, v) in zip(keys(x), vec(x))])
```
pfitzseb, line 129: Fun fact: Without these commits user types can escape into the JSON writer, where they cause world age issues again.

davidanthoff: How so?

PR #1423, opened 2020-06-29
check docversion in module/blockrange requests

pfitzseb:
Requires https://github.com/julia-vscode/LanguageServer.jl/pull/778.

Should fix a few of the errors in the crash reports.

### DIFF
--- a/src/interactive/modules.ts
+++ b/src/interactive/modules.ts
@@ -9,9 +9,6 @@ let statusBarItem: vscode.StatusBarItem = null
 let g_connection: rpc.MessageConnection = null
 let g_languageClient: vslc.LanguageClient = null
 
-const isConnectionActive = () => g_connection !== null
-const isLanguageClientActive = () => g_languageClient !== null
-
 const manuallySetDocuments = []
 
 const requestTypeGetModules = new rpc.RequestType<void, string[], void, void>('repl/loadedModules')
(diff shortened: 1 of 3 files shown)


src/interactive/modules.ts
```diff
@@ -90,12 +91,14 @@ async function updateModuleForSelectionEvent(event: vscode.TextEditorSelectionCh
...
 
-    let loaded = false
-    if (isConnectionActive()) {
-        loaded = await g_connection.sendRequest(requestTypeIsModuleLoaded, mod)
-    }
+    if (mod) {
+        let loaded = false
+        if (isConnectionActive()) {
+            loaded = await g_connection.sendRequest(requestTypeIsModuleLoaded, mod)
+        }
```
davidanthoff, line 98: I thought we were handling these situations with `try catch` now? I think we should, the `isConnectionActive()` pattern doesn't work in async code.

pfitzseb: True.

pfitzseb:
The various `console.error(err)`s should probably be patched into crash reporting somehow (depending on whether it's an expected error or not, I suppose).

davidanthoff:
> probably be patched into crash reporting somehow

You could use `traceEvent` or `traceTrace` for now, I think we just want to know whether we still see these conditions, right?

But I'd say we can merge this now and then add this in a new PR?

davidanthoff:
I'm going to merge the underlying LS PR and this one, given that we need a submodule update in-between it is probably easier if I do the whole merge story.

Issue #1396, opened 2020-06-25
Add clean inline results commands to ...


davidanthoff:
Could we add commands that clean e.g. the inline eval results to the little `...` in the top right corner of the doc?

![image](https://user-images.githubusercontent.com/1036561/85776626-bc337680-b6d5-11ea-8ddb-41557fb653d8.png)

And maybe also the right click context menu in the editor?

aviatesk:
We can even add markdown links to clear result/results in hover as error highlight:
![image](https://user-images.githubusercontent.com/40514306/85812526-d33c9c00-b79b-11ea-8634-b533e4633c31.png)

pfitzseb:
Would also be good if we could use `Escape` as the keybinding for clearing the current inline result. That requires some set-up to allow the right `when` clause to trigger, and I couldn't figure out a good way for this yet.

aviatesk:
To subscribe to cursor movements/command invocations and then check if there is an inline result or not won't be enough ?

aviatesk:
Will we like this ?
![image](https://user-images.githubusercontent.com/40514306/86022246-0cf3f800-ba65-11ea-9d23-4c8a1b64c2af.png)

PR #1424, opened 2020-06-29
improve result interaction

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1396.

Adds the `Clear Inline Results in Editor` command to the text editor title context menu:
image

Also uses `Escape` for clearing the current inline result (without messing with other keybindings too much) *and* improves the removal logic a bit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -298,7 +298,7 @@
             },
             {
                 "command": "language-julia.clearCurrentInlineResult",
-                "key": "Escape",
+                "key": "Ctrl+I Ctrl+D",
                 "when": "editorTextFocus && !editorHasSelection && !findWidgetVisible && !markerNavigationVisible && !parameterHintsVisible && !inSnippetMode && !isInEmbeddedEditor && !suggestWidgetVisible && !onTypeRenameInputVisible && !renameInputVisible && editorLangId == julia && juliaHasInlineResult"
             },
             {
--- a/package.json
+++ b/package.json
@@ -240,6 +240,10 @@
                 {
                     "when": "jlplotpaneFocus",
                     "command": "language-julia.plotpane-delete-all"
+                },
+                {
+                    "when": "editorLangId == julia",
+                    "command": "language-julia.clearAllInlineResultsInEditor"
                 }
             ],
             "commandPalette": [
@@ -294,8 +298,8 @@
             },
             {
                 "command": "language-julia.clearCurrentInlineResult",
-                "key": "Ctrl+I Ctrl+D",
-                "when": "editorTextFocus && editorLangId == julia"
+                "key": "Escape",
+                "when": "editorTextFocus && editorLangId == julia && juliaHasInlineResult"
             },
             {
                 "command": "language-julia.clearAllInlineResultsInEditor",
--- a/package.json
+++ b/package.json
@@ -299,7 +299,7 @@
             {
                 "command": "language-julia.clearCurrentInlineResult",
                 "key": "Escape",
-                "when": "editorTextFocus && editorLangId == julia && juliaHasInlineResult"
+                "when": "editorTextFocus && !editorHasSelection && !findWidgetVisible && !markerNavigationVisible && !parameterHintsVisible && !inSnippetMode && !isInEmbeddedEditor && !suggestWidgetVisible && !onTypeRenameInputVisible && !renameInputVisible && editorLangId == julia && juliaHasInlineResult"
             },
             {
                 "command": "language-julia.clearAllInlineResultsInEditor",
--- a/src/interactive/results.ts
+++ b/src/interactive/results.ts
@@ -195,9 +195,26 @@ export function activate(context) {
             gotoNextFrame(frameArg.frame)
         }),
         vscode.commands.registerCommand('language-julia.gotoLastFrame', gotoLastFrame),
-        vscode.commands.registerCommand('language-julia.clearStackTrace', clearStackTrace)
+        vscode.commands.registerCommand('language-julia.clearStackTrace', clearStackTrace),
+        vscode.window.onDidChangeTextEditorSelection(changeEvent => updateResultContextKey(changeEvent))
     )
 }
+
+function updateResultContextKey(changeEvent: vscode.TextEditorSelectionChangeEvent) {
+    if (changeEvent.textEditor.document.languageId !== 'julia') {
+        return
+    }
+    for (const selection of changeEvent.selections) {
+        for (const r of results) {
+            if (isResultInLineRange(changeEvent.textEditor, r, selection)) {
+                vscode.commands.executeCommand('setContext', 'juliaHasInlineResult', true)
+                return
+            }
+        }
+    }
+    vscode.commands.executeCommand('setContext', 'juliaHasInlineResult', false)
+}
+
 
 export function deactivate() { }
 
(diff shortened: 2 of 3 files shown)


src/interactive/results.ts
```diff
@@ -195,10 +195,27 @@ export function activate(context) {
...
+    }
+    for (const selection of changeEvent.selections) {
+        for (const r of results) {
+            if (isResultInLineRange(changeEvent.textEditor, r, selection)) {
+                vscode.commands.executeCommand('setContext', 'juliaHasInlineResult', true)
+                return
+            }
+        }
+    }
+    vscode.commands.executeCommand('setContext', 'juliaHasInlineResult', false)
```
pfitzseb, line 215: A bit unfortunate that `setContext` isn't documented: https://github.com/microsoft/vscode/issues/46445

aviatesk: I actually used this in my documentation pane PR , and it would be appreciated if you move this into utils.ts or somewhere as I did ?
https://github.com/julia-vscode/julia-vscode/blob/939065d64e9127166321df09aab6cf32783f08ac/src/utils.ts#L7-L9

src/interactive/results.ts
```diff
@@ -351,14 +368,18 @@ export function removeAll(editor: vscode.TextEditor | null = null) {
...
-        results.filter(isvalid).forEach(removeResult)
+        results.filter(r => isResultInLineRange(editor, r, selection)).forEach(removeResult)
     })
 }
 
+function isResultInLineRange(editor: vscode.TextEditor, result: Result, range: vscode.Selection | vscode.Range) {
+    const intersect = range.intersection(result.range)
+    const lineRange = new vscode.Range(range.start.with(undefined, 0), editor.document.validatePosition(range.start.with(undefined, LINE_INF)))
+    const lineIntersect = lineRange.intersection(result.range)
+    console.log(result.document === editor.document, intersect, lineIntersect)
```
aviatesk, line 381: So `range.start.with`'s `line` parameter will use that of `range`, right ?
I'm not sure exact cases where this pass gets hit though.

pfitzseb: Consider e.g. 
```
|    3+3
```
where you selected `3+3` and evaluated that. If you then place the cursor at the `|`, you can't clear the result with the current logic.

aviatesk: I get it. Sounds nice.

package.json
```diff
@@ -240,6 +240,10 @@
                 {
                     "when": "jlplotpaneFocus",
                     "command": "language-julia.plotpane-delete-all"
+                },
+                {
+                    "when": "editorLangId == julia",
+                    "command": "language-julia.clearAllInlineResultsInEditor"
```
davidanthoff, line 246: Should we add another condition whether there are actually any inline results?

package.json
```diff
@@ -294,8 +298,8 @@
             },
             {
                 "command": "language-julia.clearCurrentInlineResult",
-                "key": "Ctrl+I Ctrl+D",
-                "when": "editorTextFocus && editorLangId == julia"
+                "key": "Escape",
+                "when": "editorTextFocus && !editorHasSelection && !findWidgetVisible && !markerNavigationVisible && !parameterHintsVisible && !inSnippetMode && !isInEmbeddedEditor && !suggestWidgetVisible && !onTypeRenameInputVisible && !renameInputVisible && editorLangId == julia && juliaHasInlineResult"
```
davidanthoff, line 302: I have to admit that `Esc` does not strike me as a good keyboard shortcut for this... First because I'm super nervous that we might have overlooked some crucial `when` condition here that we need, and second because `Esc` just never does seem to remove anything in any editor. Should we try to find something a little safer?

package.json
pfitzseb: Hm, dunno. I'm not a fan of options/commands just not being there instead of them doing nothing.

davidanthoff: Could it be enabled/disabled, depending on whether there are inline results?

package.json
pfitzseb: Well, `Escape` does clear the intellisense popup and a bunch more. Those are fairly similar in that they provide information at a certain spot in the editor -- inline results aren't part of the text document either.

We can also enumerate all conditions in VSCode core. The popular gitlens extension also [does something similar](https://github.com/eamodio/vscode-gitlens/blob/6cfd9fdedd7c6ec3bfa732af7c418bbbecdfba54/package.json#L5593-L5597).

package.json
pfitzseb: No, I don't think so.

package.json
pfitzseb: FWIW, I strongly feel that `Escape` is the right keybinding for this (as does @aviatesk, I suspect). If you really don't want this for whatever reason, we can however put this behind an option (easy to add to the `when` condition) and have a "safer" alternative keybinding enabled by default.

davidanthoff: To me there is a pretty strong difference between UI elements like intellisense popups and our annotations, I think our annotations feel very much like "part of the document", because they scroll with the doc, they stay there if you switch to a different tab and back etc. UI elements like context menus, where Esc is routinely used to dismiss them, feel very different, I think: they are very transitory and disappear with almost any UI interaction. In my mind the inline results feel much more like notebook output, which I think one would also consider part of the docs?

There seem to be a lot of extensions that add decorations of various form to documents, and as far as I can tell (maybe with the exception of GitLens), almost none of them use `Esc` as the "delete decorations" keyboard shortcut?

I think we should probably not start to enable disable keyboard shortcuts via config options: wouldn't it be easier for users to just create their own keyboard binding in that situation?

pfitzseb: So what's the concern here? That
1. it is too easy to accidentally dismiss a result, or
2. that the current definition might miss a condition, so that you might need to press Escape twice to do what you actually meant to do (first one for the result, second one for the original intent)?

The reason for potentially putting this behind a setting is that figuring out the correct `when` statement isn't trivial. Could of course put it into the docs.

FWIW, Juno has been using Escape for this since forever and I don't think we heard any complaints about it so far -- someone even asked for this keybinding in VSCode on slack the other day :)

Anyhow, do you have an alternative keybinding?

aviatesk: I honestly don't have a strong opinion on this, but I'm pretty sure that there won't be an issue to go with <kbd>esc</kbd>, given there hasn't been any complaint on this in Juno.
Is there any good alternative ? <kbd>esc</kbd> is actually used for lots of things, like notifications/clearing popups/cancelling commands, etc, and to me, "removing inline results" can pretty naturally align with them

davidanthoff: So my main concern is that I think it is not really aligned with how `Esc` is used in the UI of VS Code and many other pieces of software. `Esc` in my mind makes temporary UI elements disappear/cancel. The decorations in my mind are part of the document, because they have a lifetime that for example extends beyond a tab switch in the VS Code window (and they survive other things like window minimize/maximize etc), so in my mind they are in a different bucket of UI element than say a context menu, an active edit in a tree view or something like that, where `Esc` is a typical cancel option.

I don't have a good alternative keybinding, but I find even the current keybinding a better choice.

pfitzseb: Alright, reverted to the old keybinding.

I still maintain that `Escape` is a good choice here, but that's not a hill I'm willing to die on :)

davidanthoff: We can revisit later as well, when we hopefully revisit the whole keybinding story!

PR #1429, opened 2020-06-29
Replace isnothing with comparison

davidanthoff:
Fixes a crash reporting bug: `isnothing` doesn't exist on Julia 1.0.


aviatesk:
going to merge this because it's a small bug
fix.

PR #1430, opened 2020-06-30
add clear inline results buttons

aviatesk:
adds "clear inline results" utility buttons in each result hover:
- linewise clear
- editorwise clear

Screen Shot 2020-06-30 at 13 29 50

May look verbose, so not sure we really like this.


pfitzseb:
I'm not the biggest fan of these, tbh -- I feel like an obvious keybinding an the context menu entries are enough. 
If we go with this I'd like to move the buttons to the bottom so they don't take up so much space in the hover.

aviatesk:
yeah, I don't think I will use keybindings instead of these buttons honestly.

aviatesk:
Let's close this PR for now, I also don't like making `ReplRunCodeRequestReturn` struct to be mutable.

PR #1433, opened 2020-07-01
more and improved error tracing

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/pull/1423#issuecomment-651915275.

### DIFF
--- a/src/interactive/modules.ts
+++ b/src/interactive/modules.ts
@@ -1,4 +1,3 @@
-import { SeverityLevel } from 'applicationinsights/out/Declarations/Contracts'
 import * as vscode from 'vscode'
 import * as rpc from 'vscode-jsonrpc'
 import * as vslc from 'vscode-languageclient'
@@ -94,10 +93,7 @@ async function updateModuleForEditor(editor: vscode.TextEditor) {
         mod = await getModuleForEditor(editor)
     } catch (err) {
         if (g_languageClient) {
-            telemetry.traceTrace({
-                message: err.toString(),
-                severity: SeverityLevel.Error
-            })
+            telemetry.handleNewCrashReport('getModuleForEditor error', err.toString(), '', '')
         }
     }
 
@@ -106,10 +102,7 @@ async function updateModuleForEditor(editor: vscode.TextEditor) {
         loaded = await g_connection.sendRequest(requestTypeIsModuleLoaded, mod)
     } catch (err) {
         if (g_connection) {
-            telemetry.traceTrace({
-                message: err.toString(),
-                severity: SeverityLevel.Error
-            })
+            telemetry.handleNewCrashReport('isModuleLoaded error', err.toString(), '', '')
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/interactive/modules.ts
```diff
@@ -91,16 +93,24 @@ async function updateModuleForEditor(editor: vscode.TextEditor) {
...
         mod = await getModuleForEditor(editor)
     } catch (err) {
-        console.error(err)
-        return
+        if (g_languageClient) {
+            telemetry.traceTrace({
+                message: err.toString(),
+                severity: SeverityLevel.Error
+            })
+        }
```
davidanthoff, line 101: The idea is that this is a proper bug, right? I think we should use `handleNewCrashReport` to send an exception in that case, not a trace. We might have to add another version of it that accepts an exception object directly.

There are three benefits: 1) it shows up in failures, where it belongs :), 2) it includes stack traces, 3) it honors the opt-in for crash reporting.

Oh, and I think we should actually implement something like https://github.com/julia-vscode/julia-vscode/pull/1276, and then let the error bubble up, i.e. we should add generic crash reporting to the extension code, and then let real errors bubble up to them. So in this case, we should probably just rethrow the exception if `g_languageClient` is not `undefined`.

The reason I'm using traces for the these position validation things is that right now I'm still not sure what exactly is going wrong, i.e. whether we are making an error or not etc.

src/interactive/modules.ts
```diff
@@ -91,16 +93,24 @@ async function updateModuleForEditor(editor: vscode.TextEditor) {
...
         loaded = await g_connection.sendRequest(requestTypeIsModuleLoaded, mod)
     } catch (err) {
-        console.error(err)
-        return
+        if (g_connection) {
+            telemetry.traceTrace({
+                message: err.toString(),
+                severity: SeverityLevel.Error
+            })
+        }
```
davidanthoff, line 113: Same here, I think this should show up as an exception, not a trace, and we should add a higher level error handling like in the other PR to catch such errors.

src/interactive/modules.ts
```diff
@@ -111,8 +121,14 @@ async function chooseModule() {
     try {
         possibleModules = await g_connection.sendRequest(requestTypeGetModules, null)
     } catch (err) {
-        console.error(err)
-        vscode.window.showInformationMessage('Setting a module requires an active REPL.')
+        if (g_connection) {
+            telemetry.traceTrace({
+                message: err.toString(),
+                severity: SeverityLevel.Error
+            })
```
davidanthoff, line 128: Same :)

src/interactive/modules.ts
pfitzseb: Done. I don't think we can really get stacktraces here though. Also not sure what the latest argument is supposed to be :)

davidanthoff: I I pushed two more commits to this PR: one adds a crash reporting function that accepts a JavaScript `Error` object, and then the stack trace gets automatically extracted, the second one uses this here.

The last argument is the role, it allows us to filter on errors, essentially we use it to signal in which process the crash happened. Right now we use the values `Debugger`, `Extension`, `Language Server`, `REPL` and `Symbol Server`.

pfitzseb: Ideally we'd use the _Julia stacktrace_ here, but we can figure out a good way to do that later.

PR #1438, opened 2020-07-02
Make doc pane work offline

davidanthoff:
This is a PR against the doc viewer branch https://github.com/julia-vscode/julia-vscode/pull/1308.

This removes the use of `documenter.js` and instead uses versions of the libs that we need that ship with the extension. The little code in `documenter.js` that we actually need is moved into the HTML that we generate in the extension. There are many benefits to this:
1) everything should work offline
2) `documenter.js` has a lot of stuff that we don't need (the navigation bar, settings dialog etc.)
3) we can get rid of the whole fake documentation Julia project
4) we don't need to use requirejs anymore

I also moved the various libs into their own folder: we have this convention that each sub folder in `libs` is one third party project, so it is more consistent to have say one folder for the `fontaweseome` stuff etc. I'm also pretty sure that we'll use some of these in the notebook implementation eventually, so not having them all in the `documenter` folder makes sense from that point of view as well.

I also moved the retrieval of these third party libs into the existing update script. The one thing that I haven't managed is to update the fonts with the script yet. We could probably also remove some of the font files, they are all in there in various formats.

The code highlighting works with this, and the math display works as well as it does in `avi/documentation`, but I think there are still some problems with that in general.

@aviatesk if you are ok with this, you should just merge this PR into your branch!


aviatesk:
Looks great, thanks @davidanthoff . I will try on this branch and will merge into my branch after I confirm it works.

> but I think there are still some problems with that in general.

hm, could you elaborate what kind of problem, especially ? Is that the one that we need to address ?

aviatesk:
I confirmed this works. If we are not concerning [this comment](https://github.com/julia-vscode/julia-vscode/pull/1308#issuecomment-653638706), I'm going to go with this.

davidanthoff:
Yes, I would say we can handle the navigation bar later. I'll follow up on the problem I mentioned, somehow the math for `exp` didn't render correctly for me.

But I'd say lets merge this PR and then iterate over on the other branch.

PR #1444, opened 2020-07-03
setup environement for developement

aviatesk:
So that we can develop code in `scripts/packages` with all the power of our LS.
(of course we can have something like this in our personal workspace settings, but I thought this is worth to share with developers of this repository ?)


scripts/environments/development/Manifest.toml
```diff
@@ -0,0 +1,164 @@
...
+deps = ["Tokenize"]
+path = "../../packages/CSTParser"
+uuid = "00ebfdb7-1f24-5e51-bd34-a7502290713f"
+version = "2.4.1-DEV"
+
+[[CodeTracking]]
+deps = ["InteractiveUtils", "UUIDs"]
+git-tree-sha1 = "9c173f62af93cce8af2bd3527d160b6ddd6eaf81"
+uuid = "da1fd8a2-8d9e-5ec2-8556-3022fb5608a2"
+version = "1.0.0"
```
davidanthoff, line 16: `dev` that package to our version as well?

davidanthoff: I just pushed a commit that does this.

PR #1449, opened 2020-07-05
More simplifications for doc browser

davidanthoff:
This is a grab-bag of simplifications for the doc browser. It needs https://github.com/julia-vscode/LanguageServer.jl/pull/791 to work.

Here is what I changed:
- Instead of sending messages, we now use command URIs for links. That is much less involved and I think how this kind of problem is supposed to be done from an API point of view.
- The webview is now entirely stateless, so we don't need any state management anymore. That is also a recommendation for how one should use web views.
- I store the last position of the webview in global state, so that this is remembered across sessions. This allows us to remember the position, without having to deal with the state serialization stuff, i.e. we can still be entirely stateless.
- I added `g_` for all global variables
- I removed a couple of functions, in particular ones that were only called from one call site and only had one call in them, I think in that case the further indirect just makes the code more difficult to read and follow.
- Some random other simplifications (for example, no need to construct a new `Position` from an existing `Position`)

@aviatesk, if you agree with these here, just approve the PR over on the LS repo, and I'll merge that, and then update `julia-vscode#master` to include the LS update and then I'll merge that into this branch here, just so that we don't run into submodule problems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
         "onCommand:language-julia.debug.getActiveJuliaEnvironment",
         "onLanguage:julia",
         "onLanguage:juliamarkdown",
-        "onWebviewPanel:JuliaDocumentationBrowser",
         "workspaceContains:deps/build.jl",
         "workspaceContains:test/runtests.jl",
         "workspaceContains:benchmark/benchmarks.jl",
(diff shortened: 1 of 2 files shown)


davidanthoff:
Ok, everything related to this PR should now be merged into https://github.com/julia-vscode/julia-vscode/pull/1308 (including the LS PR).

PR #1450, opened 2020-07-06
fix undef handling in workspace

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1439.

image

### DIFF
--- a/scripts/packages/VSCodeServer/src/trees.jl
+++ b/scripts/packages/VSCodeServer/src/trees.jl
@@ -78,7 +78,7 @@ function treerender(x::Leaf)
     )
 end
 
-getfield_safe(x, f, default = "#undef") = isdefined(x, f) ? getfield(x, f) : default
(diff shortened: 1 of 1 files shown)


scripts/packages/VSCodeServer/src/trees.jl
```diff
@@ -121,12 +121,22 @@ function treerender(x::Module)
     end))
 end
 
+struct Undef end
```
aviatesk, line 124: nit, how about making a singleton instance (maybe `undef′` or `undefined`) ?

scripts/packages/VSCodeServer/src/trees.jl
```diff
@@ -121,12 +121,22 @@ function treerender(x::Module)
...
+
+function undefs(xs)
```
aviatesk, line 126: nit, I think the name is better to be something like `fillup_undefs`

PR #1452, opened 2020-07-06
add disabledDirs config

aviatesk:
corresponds to https://github.com/julia-vscode/LanguageServer.jl/pull/792

### DIFF
--- a/package.json
+++ b/package.json
@@ -388,6 +388,14 @@
                     ],
                     "description": "Highlight unknown symbols. The `symbols` option will not mark unknown fields."
                 },
+                "julia.lint.disabledDirs": {
+                    "type": "array",
+                    "default": [
+                        "docs",
+                        "test"
+                    ],
+                    "markdownDescription": "Specifies sub-directories in [a package-format directory](https://docs.julialang.org/en/v1/manual/code-loading/#Package-directories-1) where only basic lintings are enabled and most false positve lintings would be surpressed."
+                },
                 "julia.lint.call": {
                     "type": "boolean",
                     "default": true,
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -388,6 +388,14 @@
                     ],
                     "description": "Highlight unknown symbols. The `symbols` option will not mark unknown fields."
                 },
+                "julia.lint.disabledDirs": {
+                    "type": "array",
+                    "default": [
+                        "docs",
+                        "test"
+                    ],
+                    "markdownDescription": "Specifies sub-directories in [a package-format directory](https://docs.julialang.org/en/v1/manual/code-loading/#Package-directories-1) where only basic lintings are enabled and most false positve lintings would be surpressed."
```
pfitzseb, line 397:
```suggestion
                    "markdownDescription": "Specifies sub-directories in [a package directory](https://docs.julialang.org/en/v1/manual/code-loading/#Package-directories-1) where only basic linting is. This drastically lowers the chance for false positives."
```

davidanthoff:
I'm torn on this one :) Mainly because I think we'll change all of this again in our hopefully final redesign of how we handle environments. I also think we should probably at some point have a way to configure the linter in a non VS Code way so that these linting settings also apply if one were to run the linter in a stand-alone fashion (something we also don't have yet), say as part of CI. On the other hand, maybe not so bad to break this setting again in the not too distant future, and for now just support it? Not really sure, I'll vote "neutral" and let you three haggle it out ;)

pfitzseb:
I think I'd like to have this in for now.

ZacLN:
I don't think this is the right approach moving forward but don't see any reason not to include it now

PR #1455, opened 2020-07-06
Render doc MD in extension

davidanthoff:
This is another PR against @aviatesk's doc branch https://github.com/julia-vscode/julia-vscode/pull/1308.

This requires https://github.com/julia-vscode/LanguageServer.jl/pull/793 to work.

The general strategy in this PR is to have the LS return documentation in markdown format, and then convert it into HTML in the extension, rather than in the LS. Turns out that has many benefits :) Here are some highlights:
- we can do the highlighting and katex conversions as part of the conversion from MD to HTML with various markdown-it plugins, i.e. we no longer need to run _any_ JavaScript in the webview for this. That should be way better with the webview design in VS Code: every time the web view is invisible (because a different tab is selected) and then it becomes visible again, it will recreate the view, and that also means it will rerun the javascript, if I understand the design correctly. By moving that JavaScript out of the webview, we save in terms of runtime there.
- we can ship less javascript files manually: because highlight and katex now run in the extension, we can use their node packages instead. Way easier to version.
- the LS part is now generic and not tied to the VS Code extension at all, i.e. in theory other clients could also use this functionality as well now, if they implemented the necessary client stuff.
- we can do the link replacement as part of the markdown-it pass. This should be way more robust than the regex hack we had on the Julia side so far.

@aviatesk ping me if you agree with this approach, and then I'll merge the two PRs that make this up in one go.


davidanthoff:
Oh, and I also pushed a commit that makes the keyboard shortcut that @pfitzseb mentioned.

src/interactive/documentation.ts
```diff
@@ -2,6 +2,30 @@ import * as path from 'path'
 import * as vscode from 'vscode'
 import { withLanguageClient } from '../extension'
 import { getVersionedParamsAtPosition, setContext } from '../utils'
+import MarkdownIt = require('markdown-it')
```
pfitzseb, line 5: This looks weird. Should either use `require` or `import`.

aviatesk:
@davidanthoff thanks so much for this , this simplifies code much and I confirmed this works very nicely in general.
I crafted yet another follow up PR #1456 , which I hope makes this branch yet better.

src/interactive/documentation.ts
aviatesk: mostly addressed in #1456 , which for `@traptitech/markdown-it-katex`, we can't just use `import` because of its weird typing: https://github.com/traPtitech/markdown-it-katex/blob/master/index.d.ts

src/interactive/documentation.ts
```diff
@@ -80,7 +104,9 @@ async function getDocumentation(): Promise<string> {
 
     return await withLanguageClient(
         async languageClient => {
-            return languageClient.sendRequest('julia/getDocAt', getVersionedParamsAtPosition(editor, editor.selection.start))
+            const docAsMD = await languageClient.sendRequest<string>('julia/getDocAt', getVersionedParamsAtPosition(editor, editor.selection.start))
+            const docAsHTML = md.render(docAsMD)
+            return docAsHTML
```
aviatesk, line 109: I would like to do this HTML conversion in `createWebviewHTML`, but we can tweak this later.

src/interactive/documentation.ts
```diff
@@ -2,6 +2,30 @@ import * as path from 'path'
...
 import { withLanguageClient } from '../extension'
 import { getVersionedParamsAtPosition, setContext } from '../utils'
+import MarkdownIt = require('markdown-it')
+
+const md = new MarkdownIt().
+    use(require('@traptitech/markdown-it-katex'), {
+        output: 'html'
+    }).
+    use(require('markdown-it-highlightjs'), {
+        inline: true, register: {
```
aviatesk, line 12: Maybe we want to have Julia syntax highlight as default, right ? #1456 does that.

aviatesk:
agh... I merged master into `avi/documentation` and cluttered commit history by rebasing ...
I ended up merging the latest avi/documentation into this branch, and now code itself doesn't change.

davidanthoff:
Can we please never ever rebase anything? That is a workflow that really doesn't work if multiple people are working on the same repo. We don't care about commit history here at all, if it is ugly that is fine, but rebase is really a big problem for the workflow.

I'll try to sort out the branch situation here, I will probably create a new clean branch for the doc stuff that replaces everything we have.

PR #1458, opened 2020-07-07
Documentation browser

davidanthoff:
This replaces https://github.com/julia-vscode/julia-vscode/pull/1308, https://github.com/julia-vscode/julia-vscode/pull/1455 and https://github.com/julia-vscode/julia-vscode/pull/1456. I couldn't figure out how to resolve the branch/force push situation other than just creating a new branch entirely.

I think this branch/PR here should incorporate everything that was on these other branches, with a few changes:
- I removed default julia highlighting for code blocks that aren't annotated as julia. I think that is safer.
- I added footnote rendering.
- I fixed a bug where `jldoctest` code blocks with additional settings weren't highlighted correctly.
- The LS PR that is needed for this is on `master` now and in this PR here, so it should be easier to try out.
- I moved the doc code into its own folder, out of interactive. It really is its own thing, so I think that makes more sense, in particular if we hope to add additional features down the road (like a treeview content etc.)
- I undid the conversion of `g_context` and `g_languageClient` to `undefined`. Two reasons: a) the update was not correct and only halfway done (there were places in the code that still checked for `null`) and b) these unrelated changes really need to go into a different PR. It gets very complicated to manage multiple PRs over a longer time period if they start to bunch together these unrelated changes.

There are still TODOs:
- [ ] links currently don't work. I think this actually would require a lot more work to get done properly because (as far as I understand it) the information required to resolve links properly is simply not available in the data we have about packages. In particular, I think links can only be resolve if one takes the information from the `docs` folder into account, and we don't have any mechanism for that right now. I think a proper solution will require us to "precompute" documentation per package/version in the cloud, similar to what we plan to do for the symbol server store story.
- [ ] there is some problem with font downloading (DA will try to figure this one out)
- [ ] It would be nice if we could disable scripts in the webview entirely from a security point of view (DA will try to figure this one out)

In terms of workflow: please don't push to this branch. If you have suggestions for improvements, either leave a comment and let me implement it or open a PR against this branch here that I can then merge in.

My vote at this point is that we don't try to rush this in before Juliacon, but defer this to a release post Juliacon, I think everything else will be too rushed at this point.


aviatesk:
This PR looks good to me. I just opened yet another small PR against this (#1461 ), but the changes there are totally unrelated to other PRs and so I'm okay with closing  #1308, #1455 and #1456.

pfitzseb:
image

This has most of the functionality except for keybindings in the doc pane and links. Will try to fix those soonish and would also like to move to a more integrated look (i.e. use VSCode's fonts and colors instead of Documenter's).

pfitzseb:
Added some basic link handling and changed to a more consistent theme:
image

pfitzseb:
Ok, this is ready now from my end. We have
- a nicely integrated Documenter theme
- links
- search (very basic: first 10 symbols with a Levenshtein distance <= 1)
- 
image

pfitzseb:
Alright, I plan to merge this tomorrow. Let me know about any objections :)

davidanthoff (commented):
Out of curiosity: why don't we need highlight anymore?

I can't approve this PR because I'm the original author, but I'm on board having this merged.

pfitzseb:
Because there's no clean way to get highlighting to work with the VSCode theme colors, so I've removed it completely for now.

ZacLN (approved):

PR #1461, opened 2020-07-08
few more cleanups against #1458

aviatesk:
A few improvements against #1458 
- add `dom` library for TS compile, which will make CI pass again
- ~disable no-longer needed command URI option~ nvm: we will use this in the future for handling links, etc
- simplify the get/set documentation logic
- handle "no-editor" case

### DIFF
--- a/src/docbrowser/documentation.ts
+++ b/src/docbrowser/documentation.ts
@@ -114,20 +114,22 @@ Please wait a few seconds and try again once the \`Starting Julia Language Serve
 `
 async function showDocumentation() {
     // telemetry.traceEvent('command-showdocumentation')
-    const inner = await getDocumentation()
-    setDocumentation(inner)
+    const editor = vscode.window.activeTextEditor
+    if (!editor) { return }
+
+    const docAsMD = await getDocumentation(editor)
+    if (!docAsMD) { return }
+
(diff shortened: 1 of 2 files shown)


src/docbrowser/documentation.ts
```diff
@@ -137,15 +138,8 @@ async function getDocumentation(): Promise<string> {
...
-    if (!inner) { return }
-    g_forwardStack = [] // initialize forward page stack for manual search
-    showDocumentationPane()
-    const html = createWebviewHTML(inner)
-    _setHTML(html)
-}
-
-function createWebviewHTML(inner: string) {
+function createWebviewHTML(docAsMD: string) {
+    const docAsHTML = md.render(docAsMD)
```
aviatesk, line 142: Do we want to wrap this in try/catch ?

davidanthoff: I don't think so, if there is a bug in `md.render` we would want to know about it, right? Or could this throw an error because of a malformed doc string? Is there such a thing as invalid markdown?

tsconfig.json
```diff
@@ -4,7 +4,8 @@
         "target": "es6",
         "outDir": "out",
         "lib": [
-            "es6"
+            "es6",
+            "dom"
```
davidanthoff, line 8: Why this change?

PR #1470, opened 2020-07-10
Put the run and debug button in proper nav group

davidanthoff:
This is where they belong, but we needed to wait until the release of VS Code 1.47, where this starts to be supported.


davidanthoff:
Yes, it just came out!

Issue #1384, opened 2020-06-21
Enhancement proposal: show Julia Workspace Explorer only when the extension is activated


EHBaozi:
There is a LaTeX extension, [LaTeX Workshop](https://github.com/James-Yu/LaTeX-Workshop), that does exactly this: a language-specific panel that only shows up when the extension activates.

This suggestion is partially motivated by the fact that we can now freely move panels around the VSCode UI, so e.g. I can create a vertical split view with the Julia Explorer and the terminal on the right side of the window. With conditional loading of the panel, I can have a full terminal when I'm not working on a Julia file/project, and have the split view kick in automatically when the Julia extension activates.

A simple demo, using LaTeX Workshop:

![EVOnacK7t9](https://user-images.githubusercontent.com/38034505/85214494-6a03f580-b363-11ea-8961-d0f8a9499e78.gif)

davidanthoff:
I have to admit I'm not sure this is a good idea, is there any other example of panels appearing and disappearing depending on some state? We also plan to add more panels in the future, and I'm not sure how that would work out then.

EHBaozi:
The Microsoft Python extension provides a unittest pane that appears on demand. I do agree though that with multiple panes it can be quite tricky, and this really is just a nice-to-have thing.

Is there any discussion about whether vscode would have workspace-configurable layout in the future? I haven't found one but that would make this unnecessary.

EHBaozi:
Thinking about this a bit more, with more panels my original proposal above would also make less sense, since vscode doesn't have tabs within these view sections and there can only be one of them splitting with the terminal.

PR #1471, opened 2020-07-10
Load Julia Explorer pane upon extension activation

EHBaozi:
As detailed in #1384 .

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -17,12 +17,15 @@ import * as packagepath from './packagepath'
 import * as smallcommands from './smallcommands'
 import * as tasks from './tasks'
 import * as telemetry from './telemetry'
+import { setContext } from "./utils"
 import * as weave from './weave'
 
 let g_languageClient: LanguageClient = null
 let g_context: vscode.ExtensionContext = null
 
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -48,6 +49,7 @@ export async function activate(context: vscode.ExtensionContext) {
     juliaexepath.activate(context)
     await juliaexepath.getJuliaExePath() // We run this function now and await to make sure we don't run in twice simultaneously later
     repl.activate(context)
+    setContext('Julia:activated', true)
```
EHBaozi, line 52: Not sure whether this is the right place to put the call; on my machine it works fine.

pfitzseb:
If we want to go with this (which I'm not convinced of) we also need a corresponding `setContext('Julia:activated', false)` (maybe call it `Julia:active` instead?) call in `deactivate`.

EHBaozi:
David also [remarked in #1384](https://github.com/julia-vscode/julia-vscode/issues/1384#issuecomment-656924804) that this can complicate matters when the extension provides multiple panes in the future. So maybe just let this one sit in the backlog? As the functionality of the extension progresses it'll be naturally clear whether this is even desirable.

PR #1472, opened 2020-07-10
Handle errors in julia binary search

davidanthoff:
This is a minimal bug fix that solves the most immediate problem with our julia.exe binary search: currently, if something goes wrong with that search, an uncaught error can escape the `getJuliaExePath` function. Because that one is called early on in the `activate` function of the extension, that interrupts the core initialization of the extension if one of those error situations occurs. This is _really_ bad because even if the user later changes the user config setting for the julia binary path to a valid value, the extension initialization is broken and nothing works.

There is a lot more we should do, but I think those other things should come after Juliacon: a) we should provide a proper UI story when the extension can't find a julia binary, b) we need to harden every part of the extension against the case that no valid julia binary is configured. I think we'll be forced to do b) automatically once we continue with the strict mode stuff.

I'll write some inline comments that explain the situation each change her addresses.

### DIFF
--- a/src/juliaexepath.ts
+++ b/src/juliaexepath.ts
@@ -3,12 +3,10 @@ import * as child_process from 'child_process'
 import * as os from 'os'
 import * as path from 'path'
 import * as process from 'process'
-import * as util from 'util'
 import * as vscode from 'vscode'
 import * as which from 'which'
 import { onDidChangeConfig } from './extension'
 import { setCurrentJuliaVersion, traceEvent } from './telemetry'
-const whichAsync = util.promisify(which)
 
 let actualJuliaExePath: string = null
 
@@ -17,6 +15,7 @@ async function setNewJuliaExePath(newPath: string) {
 
     child_process.exec(`"${newPath}" --version`, (error, stdout, stderr) => {
         if (error) {
+            actualJuliaExePath = null
             return
         }
         const version = stdout.trim()
@@ -70,7 +69,7 @@ export async function getJuliaExePath() {
             else {
                 pathsToSearch = ['julia']
             }
-            let foundJulia = false
+
             for (const p of pathsToSearch) {
                 try {
                     const res = await exec(`"${p}" --startup-file=no --history-file=no -e "println(Sys.BINDIR)"`)
@@ -80,14 +79,10 @@ export async function getJuliaExePath() {
                     } else {
                         setNewJuliaExePath(p)
                     }
-                    foundJulia = true
                     break
                 }
                 catch (e) {
                 }
-            }
-            if (!foundJulia) {
-                setNewJuliaExePath(getExecutablePath())
             }
         }
         else {
@@ -95,7 +90,16 @@ export async function getJuliaExePath() {
                 setNewJuliaExePath(getExecutablePath().replace(/^~/, os.homedir()))
             } else {
                 // resolve full path
-                setNewJuliaExePath(await whichAsync(getExecutablePath()))
+                let fullPath: string | undefined = undefined
+                try {
+                    fullPath = await which(getExecutablePath())
+                }
+                catch (err) {
+                }
+
+                if (fullPath) {
+                    setNewJuliaExePath(fullPath)
+                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


src/juliaexepath.ts
```diff
@@ -17,6 +15,7 @@ async function setNewJuliaExePath(newPath: string) {
 
     child_process.exec(`"${newPath}" --version`, (error, stdout, stderr) => {
         if (error) {
+            actualJuliaExePath = null
```
davidanthoff, line 18: This handles a situation where we currently can up in a state where it appears that a julia binary was found, but in reality it doesn't work.

src/juliaexepath.ts
```diff
@@ -80,22 +79,27 @@ export async function getJuliaExePath() {
...
                     }
-                    foundJulia = true
                     break
                 }
                 catch (e) {
                 }
             }
-            if (!foundJulia) {
-                setNewJuliaExePath(getExecutablePath())
-            }
```
davidanthoff, line 91: This whole construction was unnecessary: The outer `if` clauses already made sure we can _only_ ever end up here if `getExecutablePath` returns `null` _and_ `actualJuliaExePath` is `null`. In that case we don't need to call `setNewJuliaExePath` at all because it is essentially a no op.

src/juliaexepath.ts
```diff
@@ -80,22 +79,27 @@ export async function getJuliaExePath() {
...
+                let fullPath: string | undefined = undefined
+                try {
+                    fullPath = await which(getExecutablePath())
+                }
+                catch (err) {
+                }
+
+                if (fullPath) {
+                    setNewJuliaExePath(fullPath)
+                }
```
davidanthoff, line 102: This is the really critical fix here: `which` throws an error if it doesn't find a program, and that error escaped into the `activate` function of the main extension.

davidanthoff:
@aviatesk yes, we should not merge this if there is any regression at all!

Could you try and see which path your system normally takes in the shipping version? [This one](https://github.com/julia-vscode/julia-vscode/blob/master/src/juliaexepath.ts#L98)?

I actually don't understand why we treat the case where `julia` is a user config setting, and the case where `julia` is auto-searched differently... That whole part of the code is weird, but right now ideally we wouldn't change anything major.

Another question: @aviatesk in theory one should never have to configure just `julia` as the config value, because if the setting is left empty, then the auto search functionality should try just plain `julia` first. Does that work on your system?

Another question: 

> ave `julia` point to the actual executable in environment path

How exactly do you do this? Add the folder with the Julia binary to your `PATH` in some startup script? Add a symbolic link somewhere? Would be good to understand the precise setup there.

aviatesk:
> Could you try and see which path your system normally takes in the shipping version? This one?

Yes, `whichAsync(getExecutablePath())` resolves to `usr/local/bin/julia` (where `getExecutablePath() == `'julia'`).

> Does that work on your system? 

If I don't set the configuration, [these lines](https://github.com/julia-vscode/julia-vscode/blob/b766a131cc2c4225fa8b81cf0781421957bf18bd/src/juliaexepath.ts#L76-L84) will be hit for `p == 'julia'` and `path.join(res.stdout.trim(), p)` resolves to the full path.

> How exactly do you do this? Add the folder with the Julia binary to your PATH in some startup script? Add a symbolic link somewhere? Would be good to understand the precise setup there.

I just followed the steps are described in the [official instruction](https://julialang.org/downloads/platform/#macos). On macOS, adding some absolute path to `usr/local/bin` is the way to setup an alias so that we can invoke `julia` in terminal, etc.

> Another question: @aviatesk in theory one should never have to configure just julia as the config value, because if the setting is left empty, then the auto search functionality should try just plain julia first.

I'm slightly against this -- some users (at least me !) may want to have a control over "which version of julia used by extension" and sometimes it's not same as the one this extension automatically finds (i.e. he may want to use a bit old version of julia for this extension while there might be an newer version installed on his machine). And, then s/he may also be so lazy to write an absolute path in the configuration JSON but rather want to just write a `julia`, while he is confident which version of julia it points to...

PR #1473, opened 2020-07-11
Add diagnostics for line in getModuleForEditor

davidanthoff:
The most common crash report error we get right now is that in the get module request in the LS we are trying to convert a position into an offset, and the LS thinks the position is invalid. There are two possible explanations for this: 1) we might still have an error in the sync engine, making the LS error incorrect, or 2) we might actually be sending incorrect positions from the client.

This adds diagnostics that should allow us to narrow down which of these two it is.

The plan would be to push this PR out as quickly as possible to everyone, and then hope that we get good feedback that allows us narrow down what is going on here.

### DIFF
--- a/src/interactive/modules.ts
+++ b/src/interactive/modules.ts
@@ -1,3 +1,4 @@
+import { SeverityLevel } from 'applicationinsights/out/Declarations/Contracts'
 import * as vscode from 'vscode'
 import * as rpc from 'vscode-jsonrpc'
 import * as vslc from 'vscode-languageclient'
@@ -56,6 +57,13 @@ export async function getModuleForEditor(editor: vscode.TextEditor, position: vs
             textDocument: vslc.TextDocumentIdentifier.create(editor.document.uri.toString()),
             version: editor.document.version,
             position: position
+        }
+
+        if (position.line > editor.document.lineCount) {
+            telemetry.traceTrace({
+                message: `getModuleForEditor is about to send a request with an invalid line. Request is asking for line ${position.line}, but doc only has ${editor.document.lineCount} lines.}`,
+                severity: SeverityLevel.Error
+            })
         }
 
         try {


pfitzseb:
How about adding a
```
validPosition = ed.document.validatePosition(position)
```
and sending that? Could then also use the (potentially) more precise check of `!(validPosition == position)`.

pfitzseb:
Right now the only place where `getModuleForEditor` is called with an unvalidated position is 
https://github.com/julia-vscode/julia-vscode/blob/df8c87b66159b65dc2a803685d330dc7075c0df7/src/interactive/modules.ts#L93

aviatesk:
I guess the problem is caused by [`vscode.window.activeTextEditor`](https://github.com/julia-vscode/julia-vscode/blob/master/src/interactive/modules.ts#L41:L41) can be `undefined` -- will make a PR against this to handle that case

davidanthoff:
I'm still not convinced that we really have an error free sync implementation between the extension and the LS, so that is why I designed this as a diagnostic, rather than a fix. I thought about the validate option as well, but it is quite involved, and I'd like to keep it really simple to understand what is going on. In particular, we _do_ have crash reports where the line is the problem, so I think this diagnostic is good enough to cover those situations.

My current working hypothesis is this: at some place in the code we record a position from the current selection. Between that moment and the call to `getModuleForEditor` there is an `await` call, and that triggers some document sync stuff. At the point in the code where we record the version in the extension, the position we recorded at the beginning is already invalid. I think that story would explain what we are seeing, but it might also just be incorrect and there could be a problem in our text sync implementation. I think this diagnostic here is good enough to decide which of those cases is happening.

aviatesk:
> Between that moment and the call to getModuleForEditor there is an await call, and that triggers some document sync stuff. At the point in the code where we record the version in the extension, the position we recorded at the beginning is already invalid.

In this specific `getModuleForEditor` case, document version and position are recorded at [the same time](https://github.com/julia-vscode/julia-vscode/blob/521470f31cb3f240f70fe9df06e46c1180e015a9/src/interactive/modules.ts#L55-L59) (cause we only use default `position` argument in this file actually), and so I think `position` is maybe always valid. My hunch is that the position sometime gets invalidated _after_ we send a `getModuleAt` message to LS.

davidanthoff:
[Here](https://github.com/julia-vscode/julia-vscode/blob/521470f31cb3f240f70fe9df06e46c1180e015a9/src/interactive/repl.ts#L323) is an example where the position gets recorded, then an `await` call happens, then `getModuleForEditor` is called. I think in theory the position could get invalid during the `await startREPL(true, false)` call, right?

Maybe a better fix here would be to just eliminate all of these situations right away, i.e. just make sure we never have an `await` between the selection recording and the `getModuleForEditor` call, and see whether the bug disappears in that case...

davidanthoff:
https://github.com/julia-vscode/julia-vscode/pull/1475 is an alternative to this PR that might be a better idea.

aviatesk:
> Here is an example where the position gets recorded, then an await call happens, then getModuleForEditor is called. I think in theory the position could get invalid during the await startREPL(true, false) call, right?

aha, I forgot all about the fact that we use this function in repl.ts. I agree with the alternative and would like to try that first.

aviatesk:
hmm I just revisited crash report and it seems like the error still happens on v0.17.5 ...

pfitzseb:
That's not super surprising. #1475 doesn't fix all cases where things might get out-of-sync.

aviatesk:
I'm not when they can get out-of-sync, exactly, like when there are lots of concurrent messages that the LS message dispatcher needs to handle (i.e. positions modified while LS handles other preceding messages) ? Then `validPosition = ed.document.validatePosition(position)` won't be helpful, I guess.

davidanthoff:
I've got one more idea (https://github.com/julia-vscode/julia-vscode/pull/1496), if that doesn't work I think we'll have to add something like this PR here to get to the bottom of this.

PR #1478, opened 2020-07-12
enable run/debug julia file via commands

aviatesk:
Currently these commands will result in an error when we call them from command (not via icon).
Will use active editor as fallback resource.

### DIFF
--- a/package.json
+++ b/package.json
@@ -215,14 +215,14 @@
                 "title": "Julia: Select Current Module"
             },
             {
-                "command": "language-julia.debugEditorContents",
-                "title": "Debug Julia File",
-                "icon": "$(debug-alt)"
+                "command": "language-julia.runEditorContents",
+                "title": "Julia: Run File",
+                "icon": "$(play)"
             },
             {
-                "command": "language-julia.runEditorContents",
-                "title": "Run Julia File",
-                "icon": "$(play)"
+                "command": "language-julia.debugEditorContents",
+                "title": "Julia: Debug File",
+                "icon": "$(debug-alt)"
             }
         ],
         "menus": {
--- a/src/debugger/debugFeature.ts
+++ b/src/debugger/debugFeature.ts
@@ -52,12 +52,10 @@ export class JuliaDebugFeature {
     public dispose() { }
 }
 
-function getActiveUri(
+const getActiveUri = (
     uri: vscode.Uri | undefined,
     editor: vscode.TextEditor | undefined = vscode.window.activeTextEditor
-) {
-    return uri ? uri.fsPath : editor ? editor.document.fileName : undefined
-}
+) => uri ? uri.fsPath : editor ? editor.document.fileName : undefined
 
 export class JuliaDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
 
--- a/src/debugger/debugFeature.ts
+++ b/src/debugger/debugFeature.ts
@@ -16,12 +16,11 @@ export class JuliaDebugFeature {
                 return pkgenvpath
             }),
             vscode.commands.registerCommand('language-julia.runEditorContents', (resource: vscode.Uri | undefined) => {
-                const editor = vscode.window.activeTextEditor
-                if (!(resource || editor)) {
+                const program = getActiveUri(resource)
+                if (!program) {
                     vscode.window.showInformationMessage('No active editor found.')
                     return
                 }
-                const program = resource === undefined ? editor.document.fileName : resource.fsPath
                 vscode.debug.startDebugging(undefined, {
                     type: 'julia',
                     name: 'Run Editor Contents',
@@ -35,12 +34,11 @@ export class JuliaDebugFeature {
 			*/)
             }),
             vscode.commands.registerCommand('language-julia.debugEditorContents', (resource: vscode.Uri | undefined) => {
-                const editor = vscode.window.activeTextEditor
-                if (!(resource || editor)) {
+                const program = getActiveUri(resource)
+                if (!program) {
                     vscode.window.showInformationMessage('No active editor found.')
                     return
                 }
-                const program = resource === undefined ? editor.document.fileName : resource.fsPath
                 vscode.debug.startDebugging(undefined, {
                     type: 'julia',
                     name: 'Debug Editor Contents',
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -214,15 +214,15 @@
...
             },
-            {
-                "command": "language-julia.debugEditorContents",
-                "title": "Debug Julia File",
-                "icon": "$(debug-alt)"
-            },
             {
                 "command": "language-julia.runEditorContents",
-                "title": "Run Julia File",
+                "title": "Julia: Run File",
```
aviatesk, line 219: I think it's better to have a consistent naming for our commands, but let me know if you have any thought on having a different format for these commands, @davidanthoff

src/debugger/debugFeature.ts
```diff
@@ -42,6 +52,11 @@ export class JuliaDebugFeature {
     public dispose() { }
 }
 
+const getActiveUri = (
+    uri: vscode.Uri | undefined,
+    editor: vscode.TextEditor | undefined = vscode.window.activeTextEditor
+) => uri ? uri.fsPath : editor ? editor.document.fileName : undefined
```
pfitzseb, line 58: Maybe write this as `function getActiveUri(...) {...}` ?

aviatesk: sure, if you prefer that :)

davidanthoff:
I would actually prefer if we could hide these commands for the user. They are really duplicates of the generic F5 and Ctrl+F5, right? Is there some way that we can mark these so that they don't show up in the list of commands a user has access to, but still work for the buttons atop the document?

pfitzseb:
https://code.visualstudio.com/api/extension-guides/command#controlling-when-a-command-shows-up-in-the-command-palette should work for that, I think.

davidanthoff:
Ah yes! So I think I would just remove it from the command palette permanently. Does that make sense to others?

aviatesk:
> They are really duplicates of the generic F5 and Ctrl+F5, right?

I don't think so ? <kbd>f4</kbd> and <kbd>ctrl-f5</kbd> will first look for project-based configuration, e.g. what we setup for this repository in `.vscode` folder, while those `Julia: Run File` and `Julia: Debug File` commands will work on active-editor basis.
So for example, if you want to `Run File` in a `julia-vscode` repository, you have to use these commands instead of <kbd>ctrl-f5</kbd>

We can still hide the commands, though. Then we have to use buttons rather than via command palette or our own personal custom keybindings, but I'm not sure why we want to restrict that possibility ?

davidanthoff:
Ah, also a good point... I think I'm just worried about providing too many different entry points for fairly similar things. And in this case, if one doesn't have a specific launch config, then F5 will do the same thing. And if one has a launch config, one can configure it such that it starts the currently active file, so it just overall feels like too many options to do the same thing? Not 100% sure, though.

aviatesk:
> I think I'm just worried about providing too many different entry points for fairly similar things.

Fair. I guess `Julia: Execute File` and `Julia: Run File` can also be confusing, while they actually behave really differently.

>  And if one has a launch config, one can configure it such that it starts the currently active file

As far as I understand, these single-file debugging icon & commands are provided so that an user doesn't need to interact with the launch configurations nor even with <kbd>f5</kbd> (according to [the release note](https://code.visualstudio.com/updates/v1_47#_single-file-debugging)), so I thought that <kbd>f5</kbd> and these single-file debugging icon/commands have somewhat different meaning.

I'm not sure, tbh. I usually live with REPL and don't use one-shot running/debugging for julia code, so I don't have a strong opinion on these commands anyway. But I know lots of people will try these commands and it may be better that they don't need to think about what <kbd>f5</kbd> does, but just remember `Julia: Run File`.

Could I here your idea as well, @pfitzseb ?

pfitzseb:
I think we should probably have (and expose, document, and promote) our own commands for running/debugging the current file (corresponding to the in-editor buttons); the `F5`/`Ctrl-F5` story seems conceptually different.
I'd also argue that the `Run` button should not invoke the debugger though and instead call `Julia: Execute File`.

davidanthoff:
Yeah, the whole situation is definitely confusing...

I've been working on adding debugging to the notebook PR, following the example designs from MS, and there are some neat UI ideas in there that we might want to adopt down the road that maybe would clear things up a bit, although I'm still not sure whether they are actually an improvement or make things even more confusing ;) The general idea there seems to be that you toggle a debug mode for a notebook, and depending of that state, all the ways to execute code will either run things in the debugger or not.

At some level we could adopt something like that for our REPL: we could have a "toggle" for the REPL that puts the whole REPL into a debug mode, and then if it is in that mode, _all_ code execution commands that run stuff in the REPL would run things in the debugger. That would allow us to get rid of `@enter` and `@run` entirely, and the debugger would also naturally work with things like Alt+Enter etc.

That would still leave us with a fair number of other modes, of course...

One other point, @pfitzseb: your number 2 on the list doesn't run things in the interpreter, it really works like `include`. And for number 1 the key thing is that it runs things in the REPL process.

The way I think about `runEditorContents` and `debugEditorContents` is that these are shortcuts for F5 and Ctrl+F5 where the `program` setting of the launch config is hard coded to the currently open file.

aviatesk:
I think the discussion seems to have diverged -- introducing something like "debug model button" @davidanthoff proposed should be introduced in another PR (, which I somewhat agree with as far as we can have an recognizable & accessible UI for that).

Let's go back to the original discussion and finish this PR (, which fixes certain bugs that currently exist) ? I think we can finish this PR once we decide whether should we expose those `Julia: Run/Debug File` commands to command palette for now. Then we can decide 1.) whether to change these one-shot execution/debugging into in-process ones, and 2.) whether to introduce the global state to control "execution mode" and vanish every "debug-specific" commands, later (maybe after Juliacon).

> The way I think about `runEditorContents` and `debugEditorContents` is that these are shortcuts for F5 and Ctrl+F5 where the `program` setting of the launch config is hard coded to the currently open file.

True. My current thinking is that it's highly possible that we change these commands' behaviour in the future, so let's hide these commands from commands palette for now in order to minimize the confuses ?

davidanthoff:
My vote at this point is that we fix this after Juliacon. I don't think this is a super critical bug, and right now I think we are better of keeping the code churn really low.

mkitti:
> One other point, @pfitzseb: your number 2 on the list doesn't run things in the interpreter, it really works like `include`. And for number 1 the key thing is that it runs things in the REPL process.

Could "Julia: Execute File" be "Julia: Execute File in REPL" or "Julia: Include File in REPL" to make that *key thing* obvious?

pfitzseb:
Well, the REPL isn't involved at all in that command, it's just that we don't start a new process.

mkitti:
I/O is mapped to the terminal window that is named "Julia REPL". I recognize that this REPL actually maps to a background Julia process, but this is not apparent to the casual user.

Issue #1474, opened 2020-07-11
Crash when running distributed code line by line


LudwigBoess:
I just encountered a crash of the REPL when running code that uses `DistributedArrays.jl` line by line.

Steps to reproduce:

Run this code line by line
```julia
using Distributed
using DistributedArrays
addprocs(8)
@everywhere using DistributedArrays
@everywhere using DistributedArrays.SPMD
d_in=d=DArray(I->fill(myid(), (map(length,I)...,)), (nworkers(), 2), workers(), [nworkers(),1])
d_out=ddata();  # ! this line kills the terminal
```

Version info:

```
Julia 1.4.0
Julia vscode extension 0.17.1
Visual Studio Code 1.47.0
```

Issue #1468, opened 2020-07-10
Error not shown inline


xgdgsc:
The inline result of error isn' t complete and helpful like:
```
ERROR: Failed to precompile ExprStore [top-level] to /home/sgao/.julia/compiled/v1.4/ExprStore.ji.
Stacktrace:

error(::String) at ./error.jl:33
compilecache(::Base.PkgId, ::String) at ./loading.jl:1272
_require(::Base.PkgId) at ./loading.jl:1029
require(::Base.PkgId) at ./loading.jl:927
require(::Module, ::Symbol) at ./loading.jl:922
include_string(::Module, ::String, ::String) at ./loading.jl:1080
```
while in REPL there' s the helpful **LoadError** message:
```
[ Info: Precompiling ExprStore [top-level]
ERROR: LoadError: UndefVarError: Grammar not defined
Stacktrace:
 [1] top-level scope at /home/sgao/Documents/alpha-store/db/ExprStore.jl:58
 [2] include(::Module, ::String) at ./Base.jl:377
 [3] top-level scope at none:2
 [4] eval at ./boot.jl:331 [inlined]
 [5] eval(::Expr) at ./client.jl:449
 [6] top-level scope at ./none:3
in expression starting at /home/sgao/Documents/alpha-store/db/ExprStore.jl:58
```

pfitzseb:
Ah, I think we're unconditionally and recursively unwrapping `LoadError's.

aviatesk:
Let's keep this open until we really address this.

PR #1481, opened 2020-07-13
Fix error handling

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1474 (in that we now correctly throw a display error) and fixes https://github.com/julia-vscode/julia-vscode/issues/1468 (by only unwrapping at most one `LoadError`).

### DIFF
--- a/scripts/packages/VSCodeServer/src/eval.jl
+++ b/scripts/packages/VSCodeServer/src/eval.jl
@@ -96,7 +96,7 @@ function safe_render(x)
         return ReplRunCodeRequestReturn(
             string("Display Error: ", out.inline),
             string("Display Error: ", out.all),
-            out.iserr
+            out.stackframe
         )
     end
 end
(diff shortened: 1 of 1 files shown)


aviatesk:
LGTM, but as for #1468 , this doesn't fix it --  In the "try to precompile -> get `LoadError` about why it fails" case, the error is only sent to stderr (maybe by Pkg.jl) and we still get general `ERROR: Failed to precompile...` error (same as current behavior).

We can hack into stderr with something like [this](https://github.com/JunoLab/Atom.jl/blob/8945b9a1a08040bbe5dcab146ac0964844083d92/src/environments.jl#L3-L16), but I'm not sure it's worth to do.

pfitzseb:
Hm, I'm curious why this works in Juno then:
image
The implementations should be basically identical.

aviatesk:
Screen Shot 2020-07-13 at 23 16 34
Now I get totally lost cause it looks like above on my Juno (Julia@v1.4)...

pfitzseb:
image

Weeeeeeird.

aviatesk:
lol

PR #1496, opened 2020-07-16
Fix another getModuleForEditor race condition

davidanthoff:
The only other potential place that I could find where there might be an `await` between recording the selection position and recording the version of the document was in the default handling of `getModuleForEditor` itself. It is an async function, and it is not clear to me at which point the default argument is evaluated in those situations: pre or post the await call happening. So this could be another source of the discrepancies that we seem to be having.

I also changed the argument to be the doc itself, maybe that also helps, who knows.

If that doesn't fix it, I'm kind of out of ideas :)

### DIFF
--- a/src/interactive/modules.ts
+++ b/src/interactive/modules.ts
@@ -48,13 +48,13 @@ export function activate(context: vscode.ExtensionContext) {
     context.subscriptions.push(statusBarItem)
 }
 
-export async function getModuleForEditor(editor: vscode.TextEditor, position: vscode.Position = editor.selection.start) {
-    let mod = manuallySetDocuments[editor.document.fileName]
+export async function getModuleForEditor(document: vscode.TextDocument, position: vscode.Position) {
+    let mod = manuallySetDocuments[document.fileName]
 
     if (mod === undefined) {
         const params: VersionedTextDocumentPositionParams = {
-            textDocument: vslc.TextDocumentIdentifier.create(editor.document.uri.toString()),
-            version: editor.document.version,
+            textDocument: vslc.TextDocumentIdentifier.create(document.uri.toString()),
+            version: document.version,
             position: position
         }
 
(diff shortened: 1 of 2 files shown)


aviatesk:
Default arguments in JavaScript are [evaluated at call time](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) and `async` is just a syntax sugar that basically wraps the return value into `Promise` (as far as I understand), so tbh I'm not sure this helps...
But I also have no critical idea to fix the issue. Maybe there is something going wrong under the TS hood, so I'm okay to publish this for a try. Maybe do you have another idea, @pfitzseb ?

davidanthoff:
I think while it is evaluated at call time, it is actually evaluated _inside_ the promise. At least, if there is a bug in the function that is called for the default argument, then the promise returned is rejected, rather than an exception thrown directly. So if the function of the default argument is called inside the promise, then I believe there is the potential that lots of other things might have run between the moment when our function was called and when the default argument is evaluated. So I think this PR here _could_ fix things.

In any case, should we merge it and try it? I think there is no potential here to make things worse.

src/interactive/modules.ts
```diff
@@ -48,13 +48,13 @@ export function activate(context: vscode.ExtensionContext) {
     context.subscriptions.push(statusBarItem)
 }
 
-export async function getModuleForEditor(editor: vscode.TextEditor, position: vscode.Position = editor.selection.start) {
-    let mod = manuallySetDocuments[editor.document.fileName]
+export async function getModuleForEditor(document: vscode.TextDocument, position: vscode.Position) {
+    let mod = manuallySetDocuments[document.fileName]
```
pfitzseb, line 53:
```suggestion
_position = document.validatePosition(position)
if _position !== position
    position = _position
    telemetry.traceTrace(`Invalid position ${position} about to be passed to the LS; corrected to ${_position}.`)
end
```
Maybe also add validation here?

davidanthoff: yes, if this here doesn't help, then we should add a lot more diagnostics like that. But lets give this here one chance :)

PR #1499, opened 2020-07-18
Change default for missingrefs to none

davidanthoff:
@ZacLN  are you ok with this? I think we just have too many complaints on the forums about false positives, so maybe we should just disable this by default for now, until we fix the environment story.

### DIFF
--- a/package.json
+++ b/package.json
@@ -411,7 +411,7 @@
                 },
                 "julia.lint.missingrefs": {
                     "type": "string",
-                    "default": "all",
+                    "default": "none",
                     "enum": [
                         "none",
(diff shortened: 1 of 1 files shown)


ZacLN:
Happy with this as discussed on the last call (I attended!) though I would query whether we really had that many queries on this? (is this on Slack/zulip?)

davidanthoff:
I've seen probably most of these on discourse? Maybe I'm wrong, but I just have the (admittedly very subjective) feeling that I've answered a lot of things over the last couple of weeks with "we will fix the module handling after Juliacon" ;)

Issue #1500, opened 2020-07-18
`Julia: Execute File` doesn't recognize a module


aviatesk:
MRE would be to invoke this command for e.g. VSCodeServer/src/trees.jl.
I tracked down this is because this commands [uses `vscode.Position(0, 0)`](https://github.com/julia-vscode/julia-vscode/blob/60c68df04420c6c7101129888ec37910ae3fc7db/src/interactive/repl.ts#L228) for the current position in a document and [`get_expr`](https://github.com/julia-vscode/LanguageServer.jl/blob/91155cc204dd895df7b9e2f5e8c7edaed6ed44df/src/utilities.jl#L176-L217) return `nothing` for this case and so finally `getModuleForEditor` always returns `Main` for that case.

The super easy fix is to change the default position into `vscode.Position(0, 1)` or whatever, but that won't be ideal, taking an empty editor into consideration.
Maybe @ZacLN do you have an idea to fix this case ? If it's not easy to fix, I think we can go with the easy way for now.

pfitzseb:
That's basically https://github.com/julia-vscode/LanguageServer.jl/issues/771, I think. Unless we can get a quick fix for that, let's go with `document.validatePosition(vscode.Position(0, 1))`, imho.

PR #1505, opened 2020-07-20
quick fix for #1500

aviatesk:
let's go this way for now, I don't think we need to be so strict on handling the empty editor case


aviatesk:
okay, let me merge this for now. I would like to keep #1500 open until we really address the LS side problem.

PR #1509, opened 2020-07-21
create 'Julia: New Julia File' command

aviatesk:
Adds an utility command to create an new Julia file:
- when invoked in general context, creates an untitled editor with its language set to julia
- when invoked from file explorer context, shows the prompt for an user to enter the target path and then opens the created file setting its language to julia


src/smallcommands.ts
```diff
@@ -39,7 +41,51 @@ function applyTextEdit(we) {
...
+            vscode.workspace.openTextDocument(targetUri)
+                .then(document => {
+                    return vscode.languages.setTextDocumentLanguage(document, 'julia')
+                })
+                .then(document => {
+                    vscode.window.showTextDocument(document)
+                })
+        } catch (err) {
+            vscode.window.showErrorMessage(`Failed to create ${targetUri.fsPath}`)
+        }
```
aviatesk, line 75: Not sure this part will be really useful (`File: New File` command with file extension to be `.jl` is more intuitive way to create an julia file imho). But this way allows us to create a file with whatever file extension while setting its language to be julia.
It would be great if vscode offers some API to interact with the Explorer UI, but I couldn't find. that.

davidanthoff:
Maybe we should hook into https://code.visualstudio.com/updates/v1_59#_contribution-point-for-new-file-menu?

Issue #1491, opened 2020-07-15
Stop showing user errors for workspace in REPL


pfitzseb:
See https://discourse.julialang.org/t/julia-vscode-new-version-issue-with-jump/43134.

Minimal fix would be to just remove the various `@error` statements, but we could also move them into the tree view itself. Not sure if that's appropriate before JuliaCon though because it smells like a feature :)

davidanthoff:
Yeah, lets do after Juliacon, this is weird but not really breaking anyone's workflow, right?

pfitzseb:
It does break most REPL-centric workflows. The workspace is updated after each code execution action, so you get spammed with ~2 pages of errors no matter what you do, potentially interleaved with your intended output (because two tasks print at the same time).

davidanthoff:
Ah, didn't realize that. I think then we should remove the `@error` statements?

PR #1514, opened 2020-07-24
hide internal errors in WS

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1491.

### DIFF
--- a/scripts/packages/VSCodeServer/src/trees.jl
+++ b/scripts/packages/VSCodeServer/src/trees.jl
@@ -198,6 +198,9 @@ function getvariables()
             tree.canshow = can_display(x)
             push!(variables, tree)
         catch err
+            # FIXME: This should end up in the tree view as an "error child".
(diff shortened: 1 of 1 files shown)


scripts/packages/VSCodeServer/src/trees.jl
```diff
@@ -198,8 +198,8 @@ function getvariables()
             tree.canshow = can_display(x)
             push!(variables, tree)
         catch err
-            printstyled("Internal Error: ", bold = true, color = Base.error_color())
-            Base.display_error(err, catch_backtrace())
+            # printstyled("Internal Error: ", bold = true, color = Base.error_color())
+            # Base.display_error(err, catch_backtrace())
```
aviatesk, line 202: Maybe attach the link to [the issue](https://github.com/julia-vscode/julia-vscode/issues/1491) here for the future reference ?

davidanthoff:
I think we could merge now or post Juliacon, @pfitzseb you should make that call :)

PR #1515, opened 2020-07-24
fix up connection/module handling:

aviatesk:
- don't try to send request when ls/repl connection haven't been established
- enable telemetry correctly for any consumer functions of them

Currently these code spams the console with error logs (when `g_launguageClient` or `g_connection` isn't established, especially on startup) and this PR should suppress them.
Obviously this won't fix our recent `get_offset` errors though.

### DIFF
--- a/src/interactive/modules.ts
+++ b/src/interactive/modules.ts
@@ -49,24 +49,23 @@ export function activate(context: vscode.ExtensionContext) {
 }
 
 export async function getModuleForEditor(document: vscode.TextDocument, position: vscode.Position) {
-    let mod = manuallySetDocuments[document.fileName]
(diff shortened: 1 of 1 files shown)


davidanthoff:
Should we do this post Juliacon? I think we should enter a phase now where we only fix bugs that really affect a lot of users in a very annoying way, and leave other bugs around for now, just so that we slow down the code churn rate pre Juliacon.

aviatesk:
yes, this error is not user-facing at all, and this PR certainly touches a bit many parts of the code. Let's change the milestone to "next minor release"

aviatesk:
So now can we proceed with this ?

PR #1516, opened 2020-07-24
update the module status bar item correctly

aviatesk:
Currently our status bar is initialized as `Main` whenever REPL is active or an active workspace item is Julia editor or not. This PR fixes them by using the same function everytime we need to update status bar item.


davidanthoff:
Post Juliacon because this is really only a cosmetic bug?

aviatesk:
This PR is after #1515 , so yes, let's do after JuliaCon.

pfitzseb (approved):

PR #1522, opened 2020-07-26
always use bracketed paste for executeCodeCopyPaste

pfitzseb:
This works fine on Windows, so no need to special case that.

### DIFF
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -432,12 +432,7 @@ async function executeCodeCopyPaste(text: string, individualLine: boolean) {
     let lines = text.split(/\r?\n/)
     lines = lines.filter(line => line !== '')
     text = lines.join('\n')
-    if (individualLine || process.platform === 'win32') {
-        g_terminal.sendText(text + '\n', false)
-    }
-    else {
-        g_terminal.sendText('\u001B[200~' + text + '\n' + '\u001B[201~', false)
-    }
+    g_terminal.sendText('\u001B[200~' + text + '\n' + '\u001B[201~', false)
 }
 
 function executeSelectionCopyPaste() {


davidanthoff:
We definitely tried this in previous versions and it broke. There are many reasons why this might work now: it could be that newer VS Code versions handle this better, it could be that newer Windows handle this better, it could be that this works if the conpty options is selected but not without it. I think before we merge this we would have to test this on older Windows versions to make sure we don't introduce a regression, and also test this with both the latest master versions of Julia (where a lot of the handling of this has changed in Julia) and older Julia versions.

pfitzseb:
Yeah, that makes sense. Do you remember the failure modes?

davidanthoff:
> Do you remember the failure modes?

No. I just remember that we added it at some point, then it broke things and we removed it. Then we added it again sometime later, it broke things again and we reverted again :)

IljaK91:
This would be good for using OhMyREPL.jl with inside VSCode.

KristofferC:
Do we dare to try this again? If someone tells me how to test a PR, I can try on windows as well.

pfitzseb:
Try https://github.com/julia-vscode/SymbolServer.jl/pull/186#issuecomment-672873621.

I tried this on a fairly recent Windows version and it seemed to work fine there.

davidanthoff:
One thing we did discover lately is that Julia itself disables some sort of bracket paste support on Windows. I'm not sure that is relevant for this PR here, but there was something...

The way to test this is: clone this repo, checkout this PR, run the npm script `initialsetup` (you can just click on the run button for that in VS Code in the `NPM SCRIPTS` part), and then you just press F5 to start a VS Code instance in the debugger that is based on this PR here.

PR #1553, opened 2020-08-03
don't leak internals into ARGS

pfitzseb:


scripts/terminalserver/terminalserver.jl
```diff
@@ -1,23 +1,27 @@
...
+        "USE_PLOTPANE=true" in args && Base.Multimedia.pushdisplay(VSCodeServer.InlineDisplay())
+    end
 
-let
-    conn_pipeline, telemetry_pipeline = Base.ARGS[1:2]
-    VSCodeServer.serve(conn_pipeline; is_dev = "DEBUG_MODE" in Base.ARGS, crashreporting_pipename = telemetry_pipeline)
+    let
+        conn_pipeline, telemetry_pipeline = args[1:2]
+        VSCodeServer.serve(conn_pipeline; is_dev = "DEBUG_MODE=true" in args, crashreporting_pipename = telemetry_pipeline)
+    end
```
aviatesk, line 26:
```suggestion
    conn_pipeline, telemetry_pipeline = args[1:2]
    VSCodeServer.serve(conn_pipeline; is_dev = "DEBUG_MODE=true" in args, crashreporting_pipename = telemetry_pipeline)
```

we don't need this `let` block anymore.

scripts/terminalserver/terminalserver.jl
```diff
@@ -1,23 +1,29 @@
...
-    try
-        @eval using Revise
-        Revise.async_steal_repl_backend()
-    catch err
+let
+    args = [popfirst!(Base.ARGS) for _ in 1:5]
+    # load Revise ?
+    if "USE_REVISE=true" in args
+        try
+            @static if VERSION ≥ v"1.5"
```
davidanthoff, line 13: I see the logic, but doesn't this depend on the version of Revise? If I'm on Julia 1.5 and have an older version of Revise.jl installed, we still need the `async_steal_repl_backend` call, right?

aviatesk: yes, we can have something like
```julia
using Pkg

function revise_installed()
    function check(deps)
        return filter(collect(values(deps))) do pkg_info
            return pkg_info.name == "Revise" && 
                pkg_info.version ≥ v"2.7"
        end |> !isempty
    end

    # current
    check(Pkg.dependencies()) && return true

    # fallback
    p = normpath(Pkg.envdir(), string('v', VERSION.major, '.', VERSION.minor), "Project.toml")
    env = Pkg.Types.EnvCache(p)
    ctx = Pkg.Types.Context(; env)
    check(Pkg.dependencies(ctx)) && return true

    return false
end
```

but why I didn't include this into the patch was it may slow down our startup time.

davidanthoff: I'll try to revive my "ship Revise.jl out of the box" PR, so I think this right now is a great fix for the moment.

Issue #1530, opened 2020-07-30
Plot animations not being show in Julia Plots pane as in Juno


arthursoprana:
Hello,

I've been recently testing the `julia-vscode` extension and noticed that for animation plots generated with `@gif` or `@animate` macros the resulting `.gif` animation is not being shown in the **Julia Plots** pane as it occurs in Juno (Atom). Is this an expected behavior?

Consider the following example:

```julia
using Plots

@userplot CirclePlot
@recipe function f(cp::CirclePlot)
    x, y, i = cp.args
    n = length(x)
    inds = circshift(1:n, 1 - i)
    linewidth --> range(0, 10, length = n)
    seriesalpha --> range(0, 1, length = n)
    aspect_ratio --> 1
    label --> false
    x[inds], y[inds]
end

n = 150
t = range(0, 2π, length = n)
x = sin.(t)
y = cos.(t)

@gif for i ∈ 1:n
    circleplot(x, y, i)
end
```

When running the code above, the only way I could see the resulting animation was by manually opening the generated `.gif` image in vscode. I am currently using the extension in Windows and didn't test for the linux case yet.

Thanks in advance!

Arthur

Issue #1257, opened 2020-05-29
Config option to open the plot viewer in the second editor group by default


heliosdrm:
So that when the first plot is created, the plot viewer doesn't hide the code.
Context: https://discourse.julialang.org/t/open-vscode-plot-pane-in-2nd-editor-group/40403

heliosdrm:
Close by #1554

PR #1554, opened 2020-08-03
plot pane tweaks

pfitzseb:
Now opens plot pane in a different column instead of the current one. Also adds an unhandled disposable to the global subscriptions.

### DIFF
--- a/src/interactive/plots.ts
+++ b/src/interactive/plots.ts
@@ -42,7 +42,17 @@ export function showPlotPane() {
     const plotTitle = g_plots.length > 0 ? `Julia Plots (${g_currentPlotIndex + 1}/${g_plots.length})` : 'Julia Plots (0/0)'
     if (!g_plotPanel) {
         // Otherwise, create a new panel
-        g_plotPanel = vscode.window.createWebviewPanel('jlplotpane', plotTitle, { preserveFocus: true, viewColumn: vscode.ViewColumn.Beside }, { enableScripts: true })
+    g_plotPanel = vscode.window.createWebviewPanel(
+        'jlplotpane',
+        plotTitle,
+        {
+            preserveFocus: true,
+            viewColumn: vscode.ViewColumn.Beside
+        },
+        { 
+            enableScripts: true
+        }
+    )
         g_plotPanel.webview.html = getPlotPaneContent()
         vscode.commands.executeCommand('setContext', c_juliaPlotPanelActiveContextKey, true)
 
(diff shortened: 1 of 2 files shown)


pfitzseb:
Now also fixes https://github.com/julia-vscode/julia-vscode/issues/1530 by adding the display machinery for `image/gif`.

src/interactive/plots.ts
```diff
@@ -42,7 +42,7 @@ export function showPlotPane() {
     const plotTitle = g_plots.length > 0 ? `Julia Plots (${g_currentPlotIndex + 1}/${g_plots.length})` : 'Julia Plots (0/0)'
     if (!g_plotPanel) {
         // Otherwise, create a new panel
-        g_plotPanel = vscode.window.createWebviewPanel('jlplotpane', plotTitle, { preserveFocus: true, viewColumn: vscode.ViewColumn.Active }, { enableScripts: true })
+        g_plotPanel = vscode.window.createWebviewPanel('jlplotpane', plotTitle, { preserveFocus: true, viewColumn: vscode.ViewColumn.Beside }, { enableScripts: true })
```
aviatesk, line 45:
```suggestion
    g_plotPanel = vscode.window.createWebviewPanel(
        'jlplotpane',
        plotTitle,
        {
            preserveFocus: true,
            viewColumn: vscode.ViewColumn.Beside
        },
        { 
            enableScripts: true
        }
    )
```

Just cosmetic change.

src/interactive/plots.ts
```diff
@@ -42,7 +42,17 @@ export function showPlotPane() {
     const plotTitle = g_plots.length > 0 ? `Julia Plots (${g_currentPlotIndex + 1}/${g_plots.length})` : 'Julia Plots (0/0)'
     if (!g_plotPanel) {
         // Otherwise, create a new panel
-        g_plotPanel = vscode.window.createWebviewPanel('jlplotpane', plotTitle, { preserveFocus: true, viewColumn: vscode.ViewColumn.Active }, { enableScripts: true })
+    g_plotPanel = vscode.window.createWebviewPanel(
+        'jlplotpane',
+        plotTitle,
+        {
+            preserveFocus: true,
+            viewColumn: vscode.ViewColumn.Beside
```
davidanthoff, line 50: I think that is what we had in a previous iteration and then people didn't like that either and we changed it to the current setting :)

I think maybe we should do the same thing here that I did for the doc browser [here](https://github.com/julia-vscode/julia-vscode/pull/1458/files#diff-95d0de50f6e5fa04a90b3e4cd8e82cc8R81), i.e. remember the last location?

heliosdrm:
Is this a solution for #1257?

pfitzseb:
It doesn't add a config option, but it changes the default, yes.

src/interactive/plots.ts
pfitzseb: Done.

davidanthoff (approved):

PR #1579, opened 2020-08-06
progress bar integration

pfitzseb:
Peek 2020-08-06 19-37

ToDo:
- [x] REPL integration.
- [x] Potentially upstream a few ProgressLogging.jl changes (that aren't in this PR, actually): https://github.com/JunoLab/ProgressLogging.jl/pull/34
- [x] Support for multiple (and/or nested) progress bars.


pfitzseb:
Alright, this should be in a fairly good spot now. Reviews would be appreciated :)

scripts/packages/VSCodeServer/Manifest.toml
```diff
@@ -0,0 +1,65 @@
...
+
+[[Sockets]]
+uuid = "6462fe0b-24de-5631-8697-dd941f90decc"
+
+[[UUIDs]]
+deps = ["Random", "SHA"]
+uuid = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
+
+[[Unicode]]
+uuid = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
```
davidanthoff, line 65: Remove this file.

scripts/packages/VSCodeServer/src/eval.jl
```diff
@@ -19,63 +19,75 @@ function repl_runcode_request(conn, params::ReplRunCodeRequestParams)
...
 end
 
+function fix_displays()
+    for d in reverse(Base.Multimedia.displays)
+        if d isa InlineDisplay
+            popdisplay(InlineDisplay())
+        end
+    end
+    pushdisplay(InlineDisplay())
+end
```
davidanthoff, line 89: Not clear what this is? Shouldn't be part of this PR, right?

scripts/packages/VSCodeServer/src/progress.jl
```diff
@@ -0,0 +1,27 @@
+struct VSCodeLogger <: Logging.AbstractLogger end
+
+function Logging.handle_message(j::VSCodeLogger, level, message, _module,
+                                group, id, file, line; kwargs...)
+    progress = ProgressLogging.asprogress(level, message, _module, group, id, file, line; kwargs...)
```
davidanthoff, line 5: Is this the only thing we are using from `ProgressLogging`? Essentially a function that takes a message from the logging system and puts it into a struct? Might it make sense to actually copy that code into here and then no longer ship `ProgressLogging` at all? Really the interface we are relying is the format of the log message, not some types in this package, right?

In general it looks to me like ProgressLogging.jl is the package that packages should use that want to report progress, so maybe it would be nice to move the stuff that a `AbstractLogger` needs out of that package?

scripts/packages/VSCodeServer/src/progress.jl
```diff
@@ -0,0 +1,27 @@
...
+        return nothing
+    else
+        previous_logger = Logging.global_logger()
+        if (Base.invokelatest(Logging.min_enabled_level, previous_logger) <= Logging.LogLevel(level) ||
+            Base.CoreLogging.env_override_minlevel(group, _module)) &&
+            Base.invokelatest(Logging.shouldlog, previous_logger, level, _module, group, id)
+            Logging.handle_message(previous_logger, level, message, _module,
+                            group, id, file, line; kwargs...)
+        end
+    end
```
davidanthoff, line 17: I don't understand that whole `else` block, could you add some comments that just give a rough explanation of what it is doing? :)

davidanthoff (changes requested):
Awesome!!

I think my main question is whether we can do this without shipping ProgressLogging.jl...

scripts/packages/VSCodeServer/src/eval.jl
pfitzseb: Ah, yeah, that was an effort to fix our display logic that snaked (I dunno why I always try to type "snuck" here...) in here by accident.

scripts/packages/VSCodeServer/src/progress.jl
pfitzseb: Hm, not sure. I just pushed a commit that simply copies the relevant code from ProgressLogging.jl here, but it turns out that that's a fairly substantial part of the package.
That said, it makes https://github.com/JunoLab/ProgressLogging.jl/pull/34 unnecessary and it's probably a good thing to keep those "hacks" here...

scripts/packages/VSCodeServer/src/progress.jl
```diff
@@ -0,0 +1,209 @@
...
+
+Logging.shouldlog(::VSCodeLogger, level, _module, group, id) = true
+
+Logging.catch_exceptions(::VSCodeLogger) = true
+
+function Logging.min_enabled_level(::VSCodeLogger)
+    min(Base.invokelatest(Logging.min_enabled_level, Logging.global_logger()), Logging.LogLevel(-1))
+end
+
+module ProgessBase
```
c42f, line 29:
```suggestion
module ProgressBase
```

scripts/packages/VSCodeServer/src/progress.jl
c42f: > Really the interface we are relying is the format of the log message, not some types in this package, right?

No, quite the opposite. The intent is that the `Progress` type is the definitive "type for reporting progress", and that the other pattern matching which is done in `asprogress` is a hack to support the older `progress` keyword.

c42f:
Very cool to see this functionality close to working!

The vendoring of most of `ProgressLogging` concerns me, somewhat. As I noted above, the intent of `ProgressLogging.Progress` was to be a fairly definitive type for representing progress (with `asprogress` as a backward compatibility hack).

I'm sure you want to avoid a ProgressLogging dependency for reason though, so I'd like to understand why that is.

The logging system needs some way to connect message semantics emitted by logging frontends like `@progress` to a semantically relevant action taken by the logging backend (displaying progress bars in the UI). Currently we were considering that this would be based on matching of message types, hence the `ProgressLogging.Progress` type. But it's also possible this is the wrong direction to take. Thoughts?

davidanthoff:
We generally ship all the code that is needed for the VS Code extension in a vendored form, so that things a) always work on a user system without any additional installs, and b) we never run into problems that the version of some package that we need and that we ship might conflict with a different version of that package that the user might want to use in their project/environment. Essentially one of the requirements for us is that everything works, even if the user is using Julia 1.0 with package versions from that era.

So using specific types from packages for interfacing user code with our VS Code is very tricky in that model, because we can always run into version compatibilities when we do that.

So I think one way around this is the duck typing approach that @pfitzseb started here, which I actually kind of like :) Another would approach would be to put the type that we use for progress reporting into a package that is absolutely, 100% guaranteed to never ever get a 2.0 release (that is essentially how we handle the table integration with TableTraits.jl). I think that would only work if such a package is way, way more minimal than the current ProgressLogging.jl, it might really just have that one type defined in it or something like that. There are still some really tricky design questions with this approach, though, how we would ship that package in the extension then and make sure any other environment that uses ProgressLogging.jl also gets the version we ship etc. So I think generally I'm not too keen on this approach. The final idea I have of course would be to put this into Base or the stdlib :) I know folks want to not move stuff there, but this seems like such a fundamental part that it might be warranted, actually?

Also open to hear other ideas!

c42f:
Right, the duck typing / pattern matching approach makes some sense:
* On the downside, we're replacing an explicit type and package version dependency with an implicit structural dependency. So when/if `ProgressLogging.Progress` is updated by the user it may break the julia-vscode logging backend.  But assuming users always install the latest julia-vscode this can always be mitigated by updating julia-vscode to pattern match against multiple versions of `ProgressLogging.Progress`. So it's not too bad.
* On the upside, it removes any dependence between the user's installed package versions and the vscode implementation which seems extremely important.

> Another would approach would be to put the type that we use for progress reporting into a package that is absolutely, 100% guaranteed to never ever get a 2.0 release

I don't think we've reached that level of stability in the logging backends. So I agree that this is not the way to go for now.

> The final idea I have of course would be to put this into Base or the stdlib :) I know folks want to not move stuff there, but this seems like such a fundamental part that it might be warranted, actually?

I really want to do this. But I feel like this might be a strictly worse version of depending on a "ProgressBase" package because it gives us even less flexibility to improve the design and ties the release schedule to Base.

Anyway, thinking through your options I've come around to the idea that duck typing is the way to go for now.  Even though it's quite ugly, it leaves a *lot* more flexibility to improve the system in the future.

@tkf do you have any thoughts on this?

pfitzseb:
Would be cool if we could get consensus on whether we want to go with the duck typing approach.

> On the downside, we're replacing an explicit type and package version dependency with an implicit structural dependency. So when/if ProgressLogging.Progress is updated by the user it may break the julia-vscode logging backend. But assuming users always install the latest julia-vscode this can always be mitigated by updating julia-vscode to pattern match against multiple versions of ProgressLogging.Progress. So it's not too bad.

That doesn't seem like a downside to me :) Or at least having a structural dependency doesn't. We could (should?) probably define a minimal set of required fields and pass all others along to the frontend.

davidanthoff:
I've understood the discussion here as everyone agreeing that for now the duck typing approach is the way to go and we can merge this?

davidanthoff (approved):

c42f:
> I've understood the discussion here as everyone agreeing that for now the duck typing approach is the way to go and we can merge this?

Agreed! It seems there's unfinished business here, but this is a nice flexible option for now.

> That doesn't seem like a downside to me :) Or at least having a structural dependency doesn't.

To explain why I think structural dependency is a little strange:
* Julia has nominal types, not structural types
* Interfaces are defined in terms of generic functions rather than pattern matching

In a language with structural types and pervasive pattern matching, a structural dependency would make complete sense but in Julia I think it's a bit unusual. Having said that, logging has some particular needs which make structural typing more compelling:
* Log data has a stronger need for serialization and persistence than most data structures.
* We desire backward compatibility for serialized logs.

Perhaps this suggests we "should have" a stdlib `Logging` interface to map from Julia's nominal type system into structural types for use in log serialization. This make sense for, eg, JSON log serialization, but it's not clear this really helps all logging backends. For example `TerminalLogger` will probably still want `show(io, MIME"text/plain"(), logged_value)` to map into a textural form.

tkf:
I'm sorry I'm late to the party. (@c42f thanks for the ping!) I need to catch up with the discussion and think about this. But my gut feeling is that some packages are not adequate for vendoring because, for example, you sometimes really need _nominal_ typing or you need specific constants (like `ProgressLogging._id_var`).  I don't think it's just ProgressLogging.jl.  For example, to provide table UI, don't you need to actually load Tables.jl/TableTraits.jl?  Shouldn't we be relying on SemVer to actually interop well with the user code?

davidanthoff:
> For example, to provide table UI, don't you need to actually load Tables.jl/TableTraits.jl?

We jump through some hoops to be able to integrate with tabular data sources while _not_ loading any package ourselves and thus constraining users to specific versions of some package. You can see the code for that [here](https://github.com/julia-vscode/julia-vscode/blob/d3f98a00f498d1e3765a342dc65f316d1ab4064a/scripts/packages/VSCodeServer/src/display.jl#L156). That works well because a) TableTraits.jl is super, super minimal (really just two functions) and b) I control it so I can guarantee that this will never ever break :). The idea is essentially that we let the user load the package, and when it is loaded our integration starts to kick in.

> Shouldn't we be relying on SemVer to actually interop well with the user code?

I think in theory yes, but I think we would need to use it in a very different way than how it is typically used in the package manager right now. The core requirement we have for the extension is to always work with any packages and any package versions that a user might have. So we could never use the "I restrict things to compatible versions" aspect of SemVer from our extension side of things: we never want to restrict what versions a user can use.

I think it would be a different story if we had something like the existing [`pkgload`](https://github.com/julia-vscode/julia-vscode/blob/d3f98a00f498d1e3765a342dc65f316d1ab4064a/scripts/packages/VSCodeServer/src/display.jl#L162) structure that then was able to handle incompatible versions of packages, for example it could test whether TableTraits v1 or TableTraits v2 (which will never exist, I hope) was loaded and then make sure things work with either version.

So I guess that whole structure we use for TableTraits.jl and DataValues.jl right now could be extended to handle ProgressLogging.jl as well... But maybe we start with this PR as it is? I don't think we are locking ourselves into any design with this, and then we can still reconsider down the road if we run into problems?

tkf:
It's not clear to me if this PR handles `ProgressString` and in particular nested progress bars.  Supporting it with duck typing could be possible but I think it becomes trickier.

Isn't TableTraits-like solution easier to implement?  I imagine it'd be something like

```julia
const progresslogging_pkgid = Base.PkgId(
    Base.UUID("33c8b6b6-d38a-422a-b730-caa89a2f386c"),
    "ProgressLogging"
)

"""
    try_process_progress(f, args...; kwargs...) -> nothing or Some(ans)

Try to process logging record by a function `f(::Progress) -> ans`.  Arguments
`args` and `kwargs` are the ones passed to ```Logging.handle_message`.  Return
`Some(ans)` if it is a progress record.
"""
function try_process_progress(f, args...; kwargs...)
    m = get(Base.loaded_modules, progresslogging_pkgid, nothing)
    m === nothing && return nothing
    # TODO: maybe check the version of ProgressLogging here.
    Base.invokelatest() do
        progress = m.asprogress(args...; kwargs...)
        progress === nothing && return nothing
        return Some(f(progress))
    end
end
```

and then use it as

```julia
try_process_progress(level, message, _module, group, id, file, line; kwargs...) do progress
    JSONRPC.send_notification(conn_endpoint[], "repl/updateProgress", progress)
end isa Some && return nothing
```

?  Or maybe even

```julia
progress = try_process_progress(identity, level, message, _module, group, id, file, line; kwargs...)
if progress !== nothing
    JSONRPC.send_notification(conn_endpoint[], "repl/updateProgress", progress)
else
    ...
end
```

?

I'm guessing the latter (i.e., using `progress::Progress` outside `invokelatest`) is OK as long as we don't define `getproperty` etc. in ProgressLogging.jl.

If the above snippet is enough, it sounds like much easier solution.  Or are there some aspects that I'm missing?

pfitzseb:
Fair point. I was originally hesitant to use that approach because it means the non-ProgressLogging way of doing things (constructing logging messages manually) doesn't work. On second thought that doesn't seem like too much of a loss though.

PR #1580, opened 2020-08-07
package server setting

pfitzseb:
Now that Julia 1.5 is out it probably makes sense to allow the package server to be set from within VSCode. Setting it in `startup.jl` takes precedence, of course.

### DIFF
--- a/package.json
+++ b/package.json
@@ -635,6 +635,12 @@
                     "default": false,
                     "description": "Print executed code in REPL.",
                     "scope": "window"
+                },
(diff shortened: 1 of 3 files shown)


pfitzseb:
Oh, I forgot to commit those changes.

And no, you can only have one package server set, so this should be fine.

davidanthoff:
I think we can ship this in a patch release? Seems small enough to me.

pfitzseb:
Sure, the sooner the better :)

Issue #1593, opened 2020-08-12
[FR] Add keybind to restart the Julia REPL in Terminal as here was option in Juno


racinmat:
Currently, the only option to restart the Julia REPL is to kill Julia and then open it, which causes the terminal window to close, and then open. This is a bit annoying, I'd be grateful if there was the way to restart the REPL in a way that terminal window would stay open.

PR #1594, opened 2020-08-12
add stopREPL command

pfitzseb:
Fixes the part of https://github.com/julia-vscode/julia-vscode/issues/1593 we can fix (persistent terminals without a wrapper process require upstream support).

### DIFF
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -113,6 +113,12 @@ async function startREPL(preserveFocus: boolean, showTerminal: boolean = true) {
     }
 }
 
+function killREPL() {
+    if (g_terminal) {
+        g_terminal.dispose()
+    }
+}
+
 function debuggerRun(code: string) {
     const x = {
         type: 'julia',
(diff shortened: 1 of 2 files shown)


src/interactive/repl.ts
```diff
@@ -113,6 +113,12 @@ async function startREPL(preserveFocus: boolean, showTerminal: boolean = true) {
     }
 }
 
+function killREPL() {
+    if (g_terminal) {
+        g_terminal.dispose()
```
davidanthoff, line 118: This will presumably then trigger some even handler that sets `g_terminal` to `undefined`, right?

pfitzseb: To `null`, but yes, [this one](https://github.com/julia-vscode/julia-vscode/blob/2d4d60dba1b3e74ad8742214c673d94aa0209a83/src/interactive/repl.ts#L507). Not sure if I shouldn't explicitly do that here though.

davidanthoff: That seems good enough, I think only having one point where this is set to `null` is actually nicer :)

PR #1604, opened 2020-08-14
Changed activity bar from filled to outline

### LINKED ISSUE #1601: Use an outline icon for the activity bar
In [v1.37](https://code.visualstudio.com/updates/v1_37#_new-product-icons) vscode transitioned from filled to outline icons. Julia's Explorer looks out of place with a filled icon. Example:

image

If approved I can send a PR.

shilangyu:
Fixes #1601 

The new svg was crafted based on the vscode spec and on `images/julia-dots.svg`. After that change `images/julia-dots.svg` is not used anywhere, it can removed.

Before (left) and after (right) in light and dark themes:

image image

### DIFF
--- a/images/julia-dots-outline.svg
+++ b/images/julia-dots-outline.svg
@@ -3,9 +3,8 @@
 	circle {
 		stroke: #d7dae0;
 		fill: transparent;
-		stroke-width: 18;
-		stroke-linejoin: round;
-		r: 67;
+		stroke-width: 20;
+		r: 65;
 	}
 </style>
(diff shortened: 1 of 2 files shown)


pfitzseb:
Nice!
Seems like the line width is slightly smaller than that of the built-in icons though? Could you increase it a bit if that's really the case?

davidanthoff:
This is awesome!

shilangyu:
How about now?

image image

pfitzseb:
Looks good, thanks!

PR #1644, opened 2020-09-02
Show error msg if both insider and regular installed

davidanthoff:


ZacLN (approved):

src/extension.ts
```diff
@@ -23,6 +23,11 @@ let g_languageClient: LanguageClient = null
 let g_context: vscode.ExtensionContext = null
 
 export async function activate(context: vscode.ExtensionContext) {
+    if (vscode.extensions.getExtension('julialang.language-julia') && vscode.extensions.getExtension('julialang.language-julia-insider')) {
+        vscode.window.showErrorMessage('You have both the Julia Insider and regular Julia extension installed at the same time, which is not supported. Please uninstall one of the two extensions.')
```
pfitzseb, line 27:
```suggestion
        vscode.window.showErrorMessage('You have both the Julia Insider and regular Julia extension installed at the same time, which is not supported. Please uninstall or disable one of the two extensions.')

```

Issue #1659, opened 2020-09-12
julia-vscode doesn't handle multiprocessing correctly


k4rt33k:
I am trying to start Julia interpreter in VSCode with additional arguments `-p 4` and getting this error.
```
ERROR: LoadError: On worker 2:
ArgumentError: Package VSCodeServer [9f5989ce-84fe-42d4-91ec-6a7a8d53ed0f] is required but does not seem to be installed:
 - Run `Pkg.instantiate()` to install all recorded dependencies.

_require at ./loading.jl:993
require at ./loading.jl:922
#1 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Distributed/src/Distributed.jl:78
#105 at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Distributed/src/process_messages.jl:290
run_work_thunk at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Distributed/src/process_messages.jl:79
run_work_thunk at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Distributed/src/process_messages.jl:88
#98 at ./task.jl:333
```
The issue is with terminalserver.jl as it doesn't use `@everywhere` for changing the `LOAD_PATH` and for `using VSCodeServer`. I'll make a PR with my fixes that I tested on Julia 1.3 and 1.5.

davidanthoff:
Can you explain a bit more what you are trying to do? I think in general we don’t want to load `VSCodeServer` in worker processes and we also don’t want users to directly use the stuff that is in there, it is really an implementation detail.

If the goal is to use the debugger for multi process situation, then that will require quite a lot of extra design...

k4rt33k:
As a part of my research workflow, I regularly use `pmap` to parallelize some time consuming tasks. I just want to be able to run these functions from within vscode. I'm only loading `VSCodeServer` `@everywhere` because VSCode REPL was throwing an error and not starting up when I used `-p 4` as additional arguments in the extension settings.

It is very likely that I might be going about it the wrong way.

chelate:
This may be redundant for you, but I found this information hard to find, so I might as well post it here. To get pmap working I set In **Settings Julia: Num Threads: 8.**

Then you just have to making sure all the workers know about all the functions they need to run your simulations.  Typically I'm running code from inside a package-like folder.  I then run in my VS-code terminal
```
using Distributed
addprocs(7)
[...]

PR #1660, opened 2020-09-12
#1659 multiprocessing fix

k4rt33k:
Address the issue reported in https://github.com/julia-vscode/julia-vscode/issues/1659

### DIFF
--- a/scripts/terminalserver/terminalserver.jl
+++ b/scripts/terminalserver/terminalserver.jl
@@ -1,8 +1,8 @@
 # this script basially only handles `ARGS`
 
-
-Base.push!(LOAD_PATH, joinpath(@__DIR__, "..", "packages"))
-using VSCodeServer
+using Distributed
+@everywhere Base.push!(LOAD_PATH, joinpath(@__DIR__, "..", "packages"))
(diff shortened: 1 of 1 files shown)


rustyconover:
Lets try this instead:

```julia
@everywhere Base.push!(LOAD_PATH, joinpath(@__DIR__, "..", "packages"))
@everywhere if myid() == 1 using VSCodeServer end
@everywhere pop!(LOAD_PATH)
```

This assumes the master process has an id of 1.

k4rt33k:
@rustyconover Works for me. 
@davidanthoff Does this solution seem reasonable to you?

davidanthoff:
Wouldn't this still load `VSCodeServer` on all machines? I thought the problem is that if one runs `using VSCodeServer` on worker 1, then it always tries to load it on all other workers as well?

pfitzseb:
No, `using` should be process-local, afaiu. Which is exactly what we want, so I'm not sure why #1659 happens (some other code might try to load it or something, I guess?).

davidanthoff:
I think this was recently fixed with a different PR from @pfitzseb .

Issue #1537, opened 2020-08-01
"Rerun last cell" command feature request


Sixzero:
I have a specific feature request, which could greatly increase productivity.

I am looking for the: "Rerun last cell" command.

The problem I am facing is that, I work with 2 files, one is the main.jl lets say, the other is algo.jl. The code should be able to be run from main.jl, but when I edit algo.jl I want to rerun the code, so I HAVE to switch back to main.jl, which is timeconsuming. Intellij usually have a "Run configuration" what you can rerun anytime, what we need to have here, is "Rerun last cell" or "Rerun cell in the run history", there can be some other solutions, but we should be able to start our script just from anywhere and anytime. 

Currently I solve this issue, by creating a little fewliners, which can start the script from algo.jl, what I find bad in this solution, is that, I have many wasted line at the end of algo.jl and also I have to create such fewliners for every script I work on and it is not main.jl.

I am interested how hard it would be to make a "Rerun last cell" command. 

Please let me know if anyone have a solution which could replace this solution, or be a better version of it.

davidanthoff:
This is a really interesting suggestion!

I wonder whether we should hook into the launch configuration for this... We could add some options that allow you to create a launch config that runs stuff in the REPL, with options like you describe.

Sixzero:
I am thinking about to make this as a solution, probably I will have a little time for this. 

"Launch config" could also be a good idea, since there are many users coming from intellij IDE-s. 

The reason why I proposed the "Repeat last cell exec" was because I thought it is easy to just repeat the last thing VScode fired to REPL. Sounds like a really few lines of code to add. 

So basically to implement such a feature what I think I should do:
 - Remember which cell was executed last time. Here I want to know if vscode identifies cells by an ID or it fires the content of a cell to REPL directly by examining the code in the file?
 - Also second question is that, if the content or the ID of the cell can be cached temporarily, so we can reuse it later on? 
 - You mentioned hooks, so basically there exists already some launch configuration in VScode?

davidanthoff:
Another idea how we could solve this: if we had a command history panel, one could just double click an old command there and it would get re-executed. And then we could add a command that executes the most recent command in the history. Would that work?

We could just add the panel parallel to the existing Workspace panel.

PR #1664, opened 2020-09-14
Rerun last cell productivity enhancement feature

Cvikli:
The Request was: #1537 

For me this feature is the biggest timesaver. 

Can someone check it? 

This is my first PR, sorry if I did something wrong!

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
         "onCommand:language-julia.weave-save",
         "onCommand:language-julia.show-plotpane",
         "onCommand:language-julia.executeCell",
+        "onCommand:language-julia.executeLastCell",
         "onCommand:language-julia.executeJuliaBlockInREPL",
         "onCommand:language-julia.selectBlock",
         "onCommand:language-julia.debug.getActiveJuliaEnvironment",
@@ -193,6 +194,10 @@
             {
                 "command": "language-julia.executeCellAndMove",
                 "title": "Julia: Execute Code Cell And Move"
+            },
+            {
+                "command": "language-julia.executeLastCell",
+                "title": "Julia: Execute Last Code Cell"
             },
             {
                 "command": "language-julia.selectBlock",
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -24,6 +24,8 @@ let g_languageClient: vslc.LanguageClient = null
 
 let g_terminal: vscode.Terminal = null
 
+let last_cell_code = null;
+
 export let g_connection: rpc.MessageConnection = undefined
 
 function startREPLCommand() {
@@ -301,6 +303,18 @@ function isCellBorder(s: string) {
     return g_cellDelimiters.some(regex => regex.test(s))
 }
 
+async function executeLastCell() {
+    if (last_cell_code != null) {
+        const [startpos, endpos, code] = last_cell_code;
+        const ed = vscode.window.activeTextEditor;
+        const module: string = await modules.getModuleForEditor(
+            ed.document,
+            startpos
+        );
+        await evaluate(ed, new vscode.Range(startpos, endpos), code, module)
+    }
+}
+
 async function executeCell(shouldMove: boolean = false) {
     telemetry.traceEvent('command-executeCell')
 
@@ -338,6 +352,8 @@ async function executeCell(shouldMove: boolean = false) {
         vscode.window.activeTextEditor.selection = new vscode.Selection(nextpos, nextpos)
         vscode.window.activeTextEditor.revealRange(new vscode.Range(nextpos, nextpos))
     }
+
+    last_cell_code = [startpos, endpos, code]
 
     await evaluate(ed, new vscode.Range(startpos, endpos), code, module)
 }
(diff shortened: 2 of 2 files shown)


davidanthoff (commented):
Nice, you certainly got the mechanics right here :)

I wonder whether saving the position of the previous cell is a bit brittle, though? I think one could easily end up in a situation where that becomes invalid... Maybe saver to just store the code itself as a string and then rerun that? Of course at that point we lose the location information for that piece of code, which would be a shame... Not sure what the best trick here is...

I also wrote about another option in the underlying issue: we could add a command history UI element. I think that would subsume this approach here and might be a bit more general? But we could certainly start with something like this here and then later expand to a command history.

@pfitzseb any thoughts on the location vs code storage?

Cvikli:
Yes, exactly, invalidating is certainly a problem, but I didn't want to solve it yet. I vote for later. :D

One solution could be:
Caching the "last cell's pointer"(first and last #%%) and follow the code modification, keep it till it invalidates with deleting the cell's #%%. Whenever someone try to use the "Execute last cell" without "Selected cell"/"Invalidated cell" we could send a notification about "Cell isn't selected or invalidated"? (This is definitely a little bit hard at first for me.)

Till then we should maybe rename it to "Execute cached last cell"? So everybody gets it, that it is THE cached cell? 
Btw this tooltip could be good also when there were no cache yet at all. That is something that I should do. (But I have to figure out how the pull request work yet and how to add modification later. :D )

Cvikli:
@davidanthoff I invalidate the command in case of code mismatch, to avoid any problem arising for invalidation and it gives appropriate user feedback now. 

Btw I tried to specify the empty range. But setting zeropos to the ranges gives invalid linenumbers and linenumbers seems to be very important. 

So:
 - Renamed it to Execute Last Cached Cell, so it is evident how it works. 
 - I gave some tooltip to handle empty cell case.
 - Made a code recheck between the position and then I send invalidation message if there were some modification. 
Problem solved I hope! 

Later on someone could make some cell tracking, but I think this is enough most of the time and the other case everyone will get the right message. 

P.s. One more reason to go with this solution, if someone work on the file, that contains the cell, he should use "Execute Cell" so... actually there should be even less case who will face with limitation of the command. We could wait someone to send the request.

davidanthoff:
@julia-vscode/core any thoughts on this? I'm a bit worried that this might generally be too brittle/specific? And we should just go with the generic command history view instead?

pfitzseb:
Implementation wise this looks fine to me.

I kinda agree that this is a bit of a weird special case. Maybe a `Rerun last Julia code execution command` would be more generically useful and not as big of a project as a full command history?

Cvikli:
I would rename it to `Rerun last Julia code cell command` because "cell" word should be included in the description due to the command works on cells. Is that okay?

pfitzseb:
Sorry, my comment wasn't about the command name, it was about the functionality. Rerunning the last code execution command, no matter whether it was executing a file, cell, or code block, seems more generally useful, no?
I don't think I understand how the workflow this enables looks like, so maybe I'm missing something?

Cvikli:
Oh sorry.

I have a "main.jl" file, I go cell by cell. 
90% of the time I work in separated files. Each time I want to see the result of my edit I click back to main.jl and press "Execute cell". 
For me this function is like ultimate coding functionality that merge the julia speed + revise and vscode REPL which gives me incredible speed at development and gives the feeling of coding with heart. 

No interruption in the flow.

Sixzero:
Thanks for the implementation! 

I believe as for a first try, going with "Rerun last executed cell", could be a really good solution instead of a history of last executed cells. Sometimes having less is more.

If there will be real need for a history then I think there will be people who implement that feature, and I already see when I would use such a feature(in case I debug a modularized file from two entrance point).
But I am already proud of us, for being one more feature ahead of the "python-vscode" extension. :+1:

Cvikli:
@pfitzseb  and @davidanthoff, is it possible to get this in the next update?

Cvikli:
I made some changes during the weekend. 

I see, I have to use it for a while yet, to be sure that everybody will love the results. 

Now I am create this feature to be as convenient as possible. 
But in a week, I would suggest you guys to check out because at the current stage it is really good. 
For example I can go line by line with the laptop with this current feature, I will try to do a demo in the next month or idk, I want to show you why this is the most important feature of Julia. ;)

pfitzseb:
Alright, let's get this merged. Can you fix the merge conflict (by merging master into this branch) or do you want me to do that?

Cvikli:
I would like to do it. Also I want a final check then! 😊

Cvikli:
Ok, I think it is ready! ;)

I made a lot of testing during the last week and that is why I uploaded some more because some of these feature is essential to keep the speed. 

I realised I have to put a semicolon on the "rerun" command, because testing line by line(s) the code would flood the output with the result of the rerun "line(s)":
So I added a setting:
julis.executeion.executeLastCodeWithSemicolon:
"description": "Execute last code run with a closing semicolon to only print what we explicitely print during the rerun."
So, I tested it a lot and developed a lot with this function and it is really great I can tell you that!

One more time I describe the flow:
0. Run the whole program or the part of the code you want.
1. Run the code cell or selected code the part you want to see again from the "main.jl".
2. Examine the results.
3. Modify the code anywhere in the codebase (revise.jl will take care of the updates)
4. "Execute last code" 
5. Go to step 2.

I really believe this can be really beneficial for everyone, I hope for the best!

Sixzero:
I think the pull requests diverged a little from its main goal, but I still find it incredibly useful! 
I build my own julia-vscode version just to be able to use this feature! 
Could we get it reviewed and approved?

Issue #1556, opened 2020-08-03
Soft scope for inline evaluation?


pfitzseb:
Should inline evaluation use soft-scope rules? Maybe only when running code in `Main`?

Current behaviour for inline evaluation:
![image](https://user-images.githubusercontent.com/6735977/89201513-e1a78000-d5b1-11ea-9d1e-25c20140c66a.png)
For REPL evaluation (because we use `include_string` for that as well):
![image](https://user-images.githubusercontent.com/6735977/89201592-04399900-d5b2-11ea-8b7c-189292280e2f.png)

jlperla:
Maybe see if wrapping in https://github.com/stevengj/SoftGlobalScope.jl calls does what you need without much work?

jlperla:
Also, what about running a script? Isn't consistency with that also important?

pfitzseb:
AFAICT we're using hard-scoping *everywhere* in VSCode atm.

jlperla:
I see. So even the shift-enter to the REPL is using hardscope. Then the issue is consistency with the 1.5 repl rather than internal consistency.

One option to consider, which maintains the consistency I like, is to have a toggle with "usesoftscope" which just always passes things through the macro if it is on. It would make the REPL consistent with the 1.5 repl. It would also make the rest of code in vscode consistent with the REPL (basically an IJulia compatible mode)...

But there are many downsides, of course. Although I would always use the option personally.

jlperla:
@davidanthoff @pfitzseb I am walking away from this now that I have brought it up.

 I don't see any obvious answers here, but having the vscode repl different from the base repl doesn't smell right.

 maybe @stevengj has some thoughts since he championed this with the REPL and patched IJulia. I am sure the broader perspective on where this stuff goes next...

stevengj:
See also #540.   Especially now that this is the default in the Julia REPL, I feel like it should be the default for the terminal and other interactive contexts as well.

jlperla:
I agree completely. And to me, doing "run script" in a gui is an interactive context. So having an opt-out "usesoftscope" option which applies to absolutely everything would make lots of sense

PR #1665, opened 2020-09-15
softscope on Julia >=1.5

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1556.

This implements the following behaviour on Julia 1.5+ (everything stays the same on older versions):
- The REPL always uses softscoping.
- Inline evaluation always uses softscoping.
- Evaluating the whole file with `Julia: Execute File` always uses hardscoping.

I thought about making this user configurable, but I don't think that's a good idea.

### DIFF
--- a/scripts/packages/VSCodeServer/src/repl.jl
+++ b/scripts/packages/VSCodeServer/src/repl.jl
@@ -94,7 +94,10 @@ function evalrepl(m, line, repl, main_mode)
 end
 
 # don't inline this so we can find it in the stacktrace
-@noinline repleval(m, code, file) = include_string(m, code, file)
+@noinline function repleval(m, code, file)
+    args = VERSION >= v"1.5" ? (REPL.softscope, m, code, file) : (m, code, file)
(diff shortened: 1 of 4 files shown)


scripts/packages/VSCodeServer/src/repl.jl
```diff
@@ -94,7 +94,10 @@ function evalrepl(m, line, repl, main_mode)
 end
 
 # don't inline this so we can find it in the stacktrace
-@noinline repleval(m, code, file) = include_string(m, code, file)
+@noinline function repleval(m, code, file)
+    args = VERSION >= v"1.5" ? (REPL.softscope, m, code, file) : (m, code, file)
+    return include_string(args...)
```
aviatesk, line 99: lucky we're already using `include_string` ;)

aviatesk (approved):

jlperla:
Great! What about running the debugger? I assume that is consistent with the inline execution and repl because it is an interactive thing?

Issue #514, opened 2018-06-15
no easy opt out of telemetry


gdkrmr:
The telemetry dialogue does not provide a button "I do not want to send telemetry", the only thing you can do is to ignore the dialogue, which will pop up at every restart of VSCode, or go to the options and and change the value of `julia.enableTelemetry` to `false`. 
It took me 10 minutes to do figure that out, is annoying and gives a bad impression about the plugin's respect for user privacy (it is about actions not statements).

And btw, does the plugin send telemetry if you ignore the dialogue?

davidanthoff:
Take a look at our privacy policy: https://github.com/JuliaEditorSupport/julia-vscode/wiki/Privacy-Policy. I do think we are respecting user privacy with the approach we take, i.e. we are never sending anything unless you explicitly opt-in.

gdkrmr:
I really don't doubt that, but having an annoying pop up that has no way of making it go away permanently in an obvious way reminds of "free" android/windows apps that made me switch to open source in the first place.

gdkrmr:
Another way to make things more transparent could be to make the actual data (anonymized of course) or some condensed statistics accessible to everyone, and maybe my own data to me, so I can see where my computer stands performance wise.

This way the user gets an actual feeling on what data is being collected and what for instead of an ominous "we are collecting your data", I am sure very few people actually read the privacy policy. Also data scientists (the Julia target group) will have much more fun with data than with a boring privacy policy ;-)

gdkrmr:
Just for the record, I still think you should provide an opt-out button on that pop-up.

MartinOtter:
vs-code and your extension is really nice, so don't misunderstand me: For our web-site we are currently also struggling with the right balance between getting helpful data and privacy and therefore we had to understand the GDPR regulations that hold since May 2018 in Europe: Usually, the approach is to provide once a default option "no telemetery" and "telemetry" is only switched on if you actively make a click (fulfilled by julia-vscode). The vs-code approach to always ask for "telemetry" at start-up (if "telemetry" is not selected) seems to be on the border. You are of course right that vs-code itself might have a much much larger problem to fulfill the GDPR.

[...]

PR #1676, opened 2020-09-24
Improve telemetry dialogues

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/514.

image

I think we have enough users opting into telemetry without a nag message now, so let's go with this as a more user-friendly alternative.


davidanthoff:
Lets talk about this in our next call.

davidanthoff (approved):

PR #1690, opened 2020-10-05
interrupts for inline evaluation

pfitzseb:


package.json
```diff
@@ -318,6 +322,11 @@
                 "key": "alt+Enter",
                 "when": "editorTextFocus && editorLangId == julia"
             },
+            {
+                "command": "language-julia.interrupt",
+                "key": "Ctrl+Shift+C",
```
pfitzseb, line 327: This technically overwrites a default kebinding (`Open New External Terminal`), but that doesn't seem like a big deal to me.

scripts/packages/VSCodeServer/src/VSCodeServer.jl
```diff
@@ -69,23 +69,25 @@ include("debugger.jl")
...
         msg_dispatcher[repl_loadedModules_request_type] = repl_loadedModules_request
         msg_dispatcher[repl_isModuleLoaded_request_type] = repl_isModuleLoaded_request
-        msg_dispatcher[repl_startdebugger_notification_type] = (conn, params)->repl_startdebugger_request(conn, params, crashreporting_pipename)
+        msg_dispatcher[repl_startdebugger_notification_type] = (conn, params) -> repl_startdebugger_request(conn, params, crashreporting_pipename)
 
         while true
             msg = JSONRPC.get_next_message(conn_endpoint[])
 
-            if is_dev
+            @async if is_dev
```
pfitzseb, line 90: This *shouldn't* cause any problems, but is required for us to process multiple requests/notifications at the same time.

pfitzseb (commented):
~~Current shortcomings should be limited to the fact that it's not possible to interrupt code in the REPL from an editor and vice versa.~~

pfitzseb:
Alright, every way to evaluate code now goes through the same mechanism. As a consequence, the `language-julia.interrupt` command can interrupt the REPL as well, and `Ctrl-C` in the REPL now invokes that command.

package.json
davidanthoff: Hm, don't we have the other keyboard shortcut already? I think overwriting default keybindings is probably something we should avoid...

scripts/packages/VSCodeServer/src/VSCodeServer.jl
davidanthoff: I have to admit this makes me quite nervous because if affects absolutely everything, i.e. now the whole REPL code here needs to be reentrant safe, which is certainly not something we thought about when we wrote it initially...

But I also don't have a fantastic other idea ;) We could go back to the original model, where the code execution message is a notification that returns right away once the code running has been started, and then the server sends a notification back when the code has finished running, that way we wouldn't have these long-running code execution requests. I think that would limit the whole reentrant story to code execution, right?

davidanthoff (commented):
Cool!

I can't pretend to fully understand all of this, but I did have one comment on making the whole msg loop async.

package.json
pfitzseb: Yes, but it's very convenient not to have to focus the REPL to interrupt code, imho. Either is fine though, so I can remove this.

davidanthoff: I kind of feel we need a generic solution for this... Here is one idea: we could have a setting that says "enable conflicting julia keybindings", and that toggles one of these context values, and then we could use that in `when` clauses for all the keybindings that we want to introduce but that conflict with existing keybindings. The default for the setting would be `false`, but it would make it very easy for a user to toggle that setting, and then they would automatically get a large number of "better" keybindings, at the cost of them overwriting existing ones.

pfitzseb: Yes, I like that option. I wanted to rework our keybindings anyways, so I can implement that option while I'm at it. Ok to do that in a follow up?

theogf:
In reference to the 
```
"command": "language-julia.interrupt",
"key": "Ctrl+Shift+C",
```
maybe it's preferable to have a key combination like in Juno. For example : Ctrl + J, Ctrl + C

pfitzseb:
Alright, removed the controversial keybinding. We're also only `@async`ing runcode requests, which is enough for this feature.

davidanthoff (approved):

davidanthoff:
Test are failing, though.

PR #1707, opened 2020-10-18
Set JULIA_VSCODE_LANGUAGESERVER=1 for language server processes.

fredrikekre:
This makes it possible to use `direnv` with the VSCode extension by setting `Julia: Executable Path` to the following script:
```bash
!#/bin/bash
export DIRENV_BASH=/bin/bash
exec /bin/direnv exec "${JULIA_VSCODE_WORKSPACE}" julia "$@"
```

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -147,7 +147,8 @@ async function startLanguageServer() {
         env: {
             JULIA_DEPOT_PATH: path.join(g_context.extensionPath, 'scripts', 'languageserver', 'julia_pkgdir'),
             JULIA_LOAD_PATH: process.platform === 'win32' ? ';' : ':',
-            HOME: process.env.HOME ? process.env.HOME : os.homedir()
+            HOME: process.env.HOME ? process.env.HOME : os.homedir(),
+            JULIA_LANGUAGESERVER: '1'
         }
     }
 
--- a/src/juliaexepath.ts
+++ b/src/juliaexepath.ts
@@ -13,7 +13,11 @@ let actualJuliaExePath: string = null
 async function setNewJuliaExePath(newPath: string) {
     actualJuliaExePath = newPath
 
-    child_process.exec(`"${newPath}" --version`, (error, stdout, stderr) => {
+    const env = {
+        JULIA_LANGUAGESERVER: '1'
+    }
+
+    child_process.exec(`"${newPath}" --version`, { env: env }, (error, stdout, stderr) => {
         if (error) {
             actualJuliaExePath = null
             return


pfitzseb:
This needs to check whether `workspaceFolders` is defined, as per the docs:
```
workspaceFolders: ReadonlyArray<WorkspaceFolder> | undefined

List of workspace folders or undefined when no folder is open. Note that the first entry corresponds to the value of rootPath.
```

I'm also not sure if I like anyone messing with how we start the LS, but there's probably not much we can do about it :) This is also unlikely to be used by anyone who doesn't know what's going on, so whatever.

fredrikekre:
Thanks, I will fix that. Can it happen that the language server/Julia REPL can be started if it is undefined? Perhaps if you open a single file?

> I'm also not sure if I like anyone messing with how we start the LS, but there's probably not much we can do about it :)

I don't need this for the LS startup, but currently there is only a single option for specifying the Julia binary so I don't have much choice.

pfitzseb:
What do you think about setting a `IS_JULIA_LANGUAGESERVER` env variable instead? Then your startup script can use pwd for user processes and just pass through for the LS?

fredrikekre:
Fine with me if you think the system can handle possible different binaries running the user process and the language server.

src/juliaexepath.ts
```diff
@@ -13,7 +13,11 @@ let actualJuliaExePath: string = null
 async function setNewJuliaExePath(newPath: string) {
     actualJuliaExePath = newPath
 
-    child_process.exec(`"${newPath}" --version`, (error, stdout, stderr) => {
+    const env = {
+        JULIA_VSCODE_WORKSPACE: vscode.workspace.workspaceFolders[0].uri.fsPath
+    }
+
+    child_process.exec(`"${newPath}" --version`, { env: env }, (error, stdout, stderr) => {
```
fredrikekre, line 20: What is this value used for? Does it expect the Julia version for LS or the version for the REPL process? I tried to trace it and it seems like it is only used for telemetry? So I guess you expect this to be the version of julia that LS runs on?

pfitzseb: I think the main reason for this is to check whether the path is correct; getting the version is just a side effect.

fredrikekre:
Updated the PR

src/extension.ts
```diff
@@ -147,7 +147,8 @@ async function startLanguageServer() {
         env: {
             JULIA_DEPOT_PATH: path.join(g_context.extensionPath, 'scripts', 'languageserver', 'julia_pkgdir'),
             JULIA_LOAD_PATH: process.platform === 'win32' ? ';' : ':',
-            HOME: process.env.HOME ? process.env.HOME : os.homedir()
+            HOME: process.env.HOME ? process.env.HOME : os.homedir(),
+            JULIA_VSCODE_LANGUAGESERVER: '1'
```
fredrikekre, line 151: Perhaps just `JULIA_LANGUAGESERVER` is enough?

pfitzseb: Yeah, might be better to use something generic so other clients could set the same env var.

fredrikekre: Done.

pfitzseb (approved):

PR #1721, opened 2020-10-28
provide ways to refresh LS

aviatesk:
fixes https://github.com/julia-vscode/LanguageServer.jl/issues/841
requires https://github.com/julia-vscode/LanguageServer.jl/pull/842

@pfitzseb @davidanthoff please feel free to change the behavior by pushing into this branch; I've been a bit off from the code base and this approach may contain something that should be fixed.

/cc @zerefwayne ask me any question on those code anyway :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
         "commands": [
             {
                 "command": "language-julia.refreshLanguageServer",
-                "title": "Julia: Refresh Language Server"
+                "title": "Julia: Re-Index Language Server Cache"
             },
             {
                 "command": "language-julia.restartLanguageServer",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -107,6 +107,14 @@
             }
         ],
         "commands": [
+            {
+                "command": "language-julia.refreshLanguageServer",
+                "title": "Julia: Refresh Language Server"
```
pfitzseb, line 112:
```suggestion
                "title": "Julia: Re-Index Language Server Cache"
```

pfitzseb (approved):

PR #1728, opened 2020-10-30
default play button to "run file"

pfitzseb:
We've gotten a bunch of reports of people being annoyed/confused about the play button starting a new process/taking a long time/not using the correct launch config (this last one is because of a VSCode bug, I think).

This switches it to the in-process "Run File" command, which imho people should be using anyways. It also changes the names for "Run File"/"Debug File" to clarify that those start a new process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -233,12 +233,12 @@
             },
             {
                 "command": "language-julia.runEditorContents",
-                "title": "Julia: Run File",
+                "title": "Julia: Run File in new process",
                 "icon": "$(play)"
             },
             {
                 "command": "language-julia.debugEditorContents",
-                "title": "Julia: Debug File",
+                "title": "Julia: Debug File in new process",
                 "icon": "$(debug-alt)"
             }
         ],
@@ -271,7 +271,7 @@
                     "command": "language-julia.plotpane-delete-all"
                 },
                 {
-                    "command": "language-julia.runEditorContents",
+                    "command": "language-julia.executeFile",
                     "when": "resourceLangId == julia",
                     "group": "1_run@10"
                 },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -233,12 +233,12 @@
             },
             {
                 "command": "language-julia.runEditorContents",
-                "title": "Julia: Run File",
+                "title": "Julia: Run File in new process",
```
aviatesk, line 236:
```suggestion
                "title": "Julia: Run File in New Process",
```

package.json
```diff
@@ -233,12 +233,12 @@
...
                 "icon": "$(play)"
             },
             {
                 "command": "language-julia.debugEditorContents",
-                "title": "Julia: Debug File",
+                "title": "Julia: Debug File in new process",
```
aviatesk, line 241:
```suggestion
                "title": "Julia: Debug File in New Process",
```

aviatesk (commented):
Sounds good to me.

aviatesk (approved):

david-hofmann:
Dear developers! I have a comment regarding the behavior of the play button. My team and I were puzzled when we noticed that a line-by-line evaluation of the Julia code works fine but when running the whole script at once it threw an error. It took a while until we realized that the code that it ran was simply not the visible code but the file's content (and the current status wasn't saved).

This behavior of the play (or run) button doesn't seem intuitive. I figure it is because that button essentially executes what is spelled out in `launch.json` which (in its default form) runs the file on the harddrive. 
However, for instance the Python IDE Spyder saves a file when clicking the run button and thus executes the visible version. I find that behavior more intuitive if it is too complicated to run the code that is present in working memory and thus the one that one sees on screen. I don't see an immediate disadvantage of automatically saving when clicking the run button. Would be glad to learn if there is some!

Thanks!

pfitzseb:
Thanks for the report, will be fixed in #2431.

PR #1729, opened 2020-10-30
fix plot pane setting

pfitzseb:
This makes the plot pane setting apply instantly without needing to restart the Julia process. It's also much more robust against other packages trying to inject their own displays above ours.
Also simplifies a bunch of `display`/`displayable` definitions.

### DIFF
--- a/scripts/packages/VSCodeServer/src/display.jl
+++ b/scripts/packages/VSCodeServer/src/display.jl
@@ -19,7 +19,7 @@ function sendDisplayMsg(kind, data)
     JSONRPC.send_notification(conn_endpoint[], "display", Dict{String,String}("kind" => kind, "data" => data))
 end
 
-function display(_::InlineDisplay, m::MIME, x)
+function display(d::InlineDisplay, m::MIME, x)
     mime = string(m)
     if PLOT_PANE_ENABLED[] && mime in DISPLAYABLE_MIMES
         payload = stringmime(m, x)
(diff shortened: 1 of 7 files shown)


scripts/packages/VSCodeServer/src/display.jl
```diff
@@ -1,105 +1,50 @@
...
+function sendDisplayMsg(kind, data)
+    JSONRPC.send_notification(conn_endpoint[], "display", Dict{String,String}("kind" => kind, "data" => data))
 end
 
-displayable(d::InlineDisplay, ::MIME{Symbol("text/html")}) = true
-
-function display(d::InlineDisplay, ::MIME{Symbol("juliavscode/html")}, x)
-    payload = stringmime(MIME("juliavscode/html"), x)
-    sendDisplayMsg("juliavscode/html", payload)
+function display(_::InlineDisplay, m::MIME, x)
```
aviatesk, line 22:
```suggestion
function display(d::InlineDisplay, m::MIME, x)
```

cause we may throw an err in `throw(MethodError(display, (d, m, x)))`

scripts/packages/VSCodeServer/src/display.jl
```diff
@@ -1,105 +1,50 @@
 struct InlineDisplay <: AbstractDisplay end
```
aviatesk, line 1: nit, it feels like `InlineDisplay` is now better to be renamed to `VSCodeDisplay` or such.

aviatesk (approved):

scripts/packages/VSCodeServer/src/display.jl
pfitzseb: Yeah, true. Not too bad though because it's not exported and `VSCodeServer.InlineDisplay` makes some sense ;)

PR #1731, opened 2020-11-01
set up separate listeners for workspace and progress update

aviatesk:


src/interactive/workspace.ts
```diff
@@ -87,7 +82,7 @@ export function activate(context: vscode.ExtensionContext) {
     context.subscriptions.push(vscode.commands.registerCommand('language-julia.showInVSCode', showInVSCode))
     context.subscriptions.push(onInit(connection => {
         g_connection = connection
-        connection.onNotification(notifyTypeReplFinishEval, replFinishEval)
+        connection.onNotification(notifyTypeReplFinishEval, async () => await updateReplVariables())
```
pfitzseb, line 85:
```suggestion
        connection.onNotification(notifyTypeReplFinishEval, updateReplVariables)
```

src/interactive/repl.ts
```diff
@@ -180,6 +180,10 @@ const g_onInit = new vscode.EventEmitter<rpc.MessageConnection>()
 export const onInit = g_onInit.event
 const g_onExit = new vscode.EventEmitter<Boolean>()
 export const onExit = g_onExit.event
+const g_onStartEval = new vscode.EventEmitter<null>()
+export const onStartEval = g_onStartEval.event
+const g_onFinishEval = new vscode.EventEmitter<null>()
+export const onFinishEval = g_onFinishEval.event
```
aviatesk, line 186: `rpc.onMessageConnection.onNotification` only allow one single hook to be registered, so we need to setup our own event hooks for them.

scripts/packages/VSCodeServer/src/eval.jl
```diff
@@ -96,6 +96,8 @@ function repl_runcode_request(conn, params::ReplRunCodeRequestParams)
                         @eval Main ans = $(QuoteNode(ans))
                     catch err
                         EvalError(err, catch_backtrace())
+                    finally
+                        JSONRPC.send_notification(conn_endpoint[], "repl/finisheval", nothing)
```
pfitzseb, line 100: Together with https://github.com/julia-vscode/julia-vscode/blob/master/src/interactive/repl.ts#L303 this will trigger the WS update twice, no?

aviatesk: yes, so I removed those calls (see: https://github.com/julia-vscode/julia-vscode/blob/57755e25779c30a90a10cdc314cdbf1465507bfc/src/interactive/repl.ts#L259-L298 and https://github.com/julia-vscode/julia-vscode/blob/57755e25779c30a90a10cdc314cdbf1465507bfc/src/interactive/repl.ts#L353-L392)

pfitzseb (approved):

PR #1733, opened 2020-11-01
evaluation indicator for inline evaluation

aviatesk:
image

- built on top of #1731 
- will be suppressed when progress indicator is enabled
- click to jump to the code location

### DIFF
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -5,6 +5,7 @@ import * as path from 'path'
 import * as vscode from 'vscode'
 import * as rpc from 'vscode-jsonrpc'
 import * as vslc from 'vscode-languageclient'
+import { Command } from 'vscode-languageclient'
 import { onSetLanguageClient } from '../extension'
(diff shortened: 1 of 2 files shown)


src/interactive/repl.ts
```diff
@@ -540,22 +546,46 @@ export async function replStartDebugger(pipename: string) {
...
+        }),
+        onStartEval(info => {
+            evaluationIndicator.tooltip = `click to jump to the code`
+            if (info) {
+                evaluationIndicator.command = Command.create('openFile', 'language-julia.openFile', {
+                    path: info.filename,
+                    line: info.line + 1,
+                })
+            }
+            evaluationIndicator.show()
```
aviatesk, line 580: it might be more useful if click on the indicator interrupts the execution ?

pfitzseb: Yeah. What do you think about https://github.com/julia-vscode/julia-vscode/pull/1735 as an alternative to this?

aviatesk:
close in favor of #1735

PR #1735, opened 2020-11-02
interrupt evaluation from progress monitor

pfitzseb:
Alternative to https://github.com/julia-vscode/julia-vscode/pull/1733 -- this shows an indeterminate progress indicator when evaluating anything inline. Also shows a `Cancel` button, which throws an interrupt at Julia (can't guarantee that'll end up in the actual `@progress` loop, but whatever).


aviatesk (approved):
nice, this is better both in functionality and implementation.

PR #1744, opened 2020-11-04
cmd: add package to workspace

pfitzseb:


aviatesk (approved):

pfitzseb:
Turns out this doesn't actually work this way. The `workbench.action.addRootFolder` command *also* doesn't take a URI... :/

PR #1745, opened 2020-11-04
add 'Activate Parent Environment' command

aviatesk:
This is especially useful when invoked from active editor, etc.

### DIFF
--- a/scripts/packages/VSCodeServer/src/misc.jl
+++ b/scripts/packages/VSCodeServer/src/misc.jl
@@ -139,3 +139,24 @@ function activate_uri(conn, uri)
     Pkg.activate(uri)
     return nothing
 end
+
+function activate_from_dir(conn, dir)
+    path = search_up_file("Project.toml", dir)
+    if path === nothing
+        @warn "No Project.toml found for `$(dir)`"
+        return
+    end
+    Pkg.activate(path)
+    return dirname(path)
+end
+
+function search_up_file(basename, dir)
+    parent_dir = dirname(dir)
+    return if (parent_dir == dir || # ensure to escape infinite recursion
+        isempty(dir)) # reached to the system root
+        nothing
+    else
+        path = joinpath(dir, basename)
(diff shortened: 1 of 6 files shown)


scripts/packages/VSCodeServer/src/misc.jl
```diff
@@ -139,3 +139,24 @@ function activate_uri(conn, uri)
...
+end
+
+function search_up_file(basename, dir)
+    parent_dir = dirname(dir)
+    return if (parent_dir == dir || # ensure to escape infinite recursion
+        isempty(dir)) # reached to the system root
+        nothing
+    else
+        path = joinpath(dir, basename)
+        isfile(path) ? path : search_up_file(basename, parent_dir)
```
pfitzseb, line 160: Probably better to use a "safe" `isfile` (e.g. the usual `isfile′(path) = try isfile(path) catch err false end`).

scripts/packages/VSCodeServer/src/misc.jl
```diff
@@ -139,3 +139,24 @@ function activate_uri(conn, uri)
     Pkg.activate(uri)
     return nothing
 end
+
+function activate_from_dir(conn, dir)
+    path = search_up_file("Project.toml", dir)
+    if path === nothing
+        @warn "No Project.toml found for `$(dir)`"
```
pfitzseb, line 146: I think I'd prefer to show this in the VSCode UI. Maybe return a `(error = true, dir = dir)` NT here instead and show a notification in VSCode?

pfitzseb (changes requested):

scripts/packages/VSCodeServer/src/misc.jl
aviatesk: yeah, good point

scripts/packages/VSCodeServer/src/misc.jl
aviatesk: addressed

pfitzseb (approved):

PR #1747, opened 2020-11-05
pushfirst/popfirst LOAD_PATH

pfitzseb:
Otherwise we can end up loading the wrong version of VSCodeServer, depending on how `LOAD_PATH` was set.

### DIFF
--- a/scripts/debugger/run_debugger.jl
+++ b/scripts/debugger/run_debugger.jl
@@ -1,7 +1,7 @@
 # ENV["JULIA_DEBUG"] = "all"
 
-Base.push!(LOAD_PATH, joinpath(@__DIR__, "..", "packages"))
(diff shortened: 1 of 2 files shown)


fredrikekre:
Same here: https://github.com/julia-vscode/julia-vscode/blob/df7e0dfb5ec3cb2018f4842527beb05c27b07a9b/scripts/debugger/run_debugger.jl#L3-L5

fredrikekre:
I'll just note that this trick: https://github.com/julia-vscode/julia-vscode/blob/df7e0dfb5ec3cb2018f4842527beb05c27b07a9b/.vscode/settings.json#L30 will not work after this, but IMO it is better to foolproof the extension rather than simplify for developers.

pfitzseb:
Yep. It doesn't make much sense to have that for development anyways; much easier to modify the code in `scripts/packages` directly.

davidanthoff (approved):

Issue #1752, opened 2020-11-06
Alt+Enter doesn't send input to REPL


JeffBezanson:
I have Ctrl+Enter set to paste the input line in the REPL, where it also then gets added to the history. However, Alt+Enter doesn't do that. Is there a separate setting for that, or does this need to be added?

aviatesk:
<kbd>alt-enter</kbd> doesn't send code to REPL, but evaluates it via `include_string`. 
Appending the evaluated code to REPL history would still be possible.

PR #1754, opened 2020-11-10
add inline-eval'ed code to repl history

pfitzseb:
if `codeInREPL` is set.

Fixes https://github.com/julia-vscode/julia-vscode/issues/1752.

I thought about putting this behind another setting, but it feels fairly intuitive to reuse `codeInREPL` for this.


aviatesk (commented):
look good to me

aviatesk (approved):

PR #1756, opened 2020-11-10
better env watching

pfitzseb:
Not sure if we want this, but it's very convenient to also watch the global environment for changes and re-index if necessary.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2,6 +2,7 @@
 // The module 'vscode' contains the VS Code extensibility API
 // Import the module and reference it with the alias vscode in your code below
 import { SeverityLevel } from 'applicationinsights/out/Declarations/Contracts'
+import { unwatchFile, watchFile } from 'async-file'
 import * as os from 'os'
 import * as path from 'path'
 import * as vscode from 'vscode'
@@ -21,6 +22,7 @@ import * as weave from './weave'
 
 let g_languageClient: LanguageClient = null
 let g_context: vscode.ExtensionContext = null
+let g_watchedEnvironmentFile: string = null
 
 export async function activate(context: vscode.ExtensionContext) {
     if (vscode.extensions.getExtension('julialang.language-julia') && vscode.extensions.getExtension('julialang.language-julia-insider')) {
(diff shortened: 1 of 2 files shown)


aviatesk:
I guess the PR currently only changes vscode's file watcher to node's polling, right ? Not sure which part does:

> also watch the global environment for changes

(I assume this means we will watch for the currently active project and also for global environment (like v1.6) if the current active one is not the global one)

pfitzseb:
`vscode.workspace.createFileSystemWatcher` only works for files in the workspace, which usually isn't true for e.g. `~/.julia/environments/v1.5/Manifest.toml`.

aviatesk:
Makes sense. And I think we don't need to try to watch both env files for now; currently the LS is only able to handle one single environment anyway.

src/extension.ts
```diff
@@ -216,16 +218,19 @@ async function startLanguageServer() {
...
+
     // automatic environement refreshing
-    const targetFile = (await jlpkgenv.getProjectFilePaths(jlEnvPath)).manifest_toml_path
-    const disposableWatcher = vscode.workspace.createFileSystemWatcher(targetFile) // TODO: needs to fix backslashes on Windows ?
-    g_context.subscriptions.push(
-        disposableWatcher,
-        disposableWatcher.onDidChange(_ => {
+    g_watchedEnvironmentFile = (await jlpkgenv.getProjectFilePaths(jlEnvPath)).manifest_toml_path
+    // polling watch for robustness
+    watchFile(g_watchedEnvironmentFile, { interval: 10000 }, (curr, prev) => {
```
aviatesk, line 228: I'm not familiar with file systems honestly, but as far as I read the [doc of `fs.watchFile`](https://nodejs.org/api/fs.html#fs_fs_watchfile_filename_options_listener), I think using `fs.watch` is more robust and preferred way ?

> Using fs.watch() is more efficient than fs.watchFile and fs.unwatchFile. fs.watch should be used instead of fs.watchFile and fs.unwatchFile when possible.

Could you tell why the polling is more robust for this case ?
Well, there doesn't seem to be any "unwatch API" for that, I guess we can also discard the last callback if we've started to watch for a different file.

pfitzseb: [`fs.watch` sounded kinda scary to me](https://nodejs.org/api/fs.html#fs_caveats). I'm not sure if it really matters either way though.

> Well, there doesn't seem to be any "unwatch API" for that, I guess we can also discard the last callback if we've started to watch for a different file.

```
const watcher = fs.watch(...)
watcher.close()
```
does the trick in that case.

aviatesk: yeah, I just saw the section; let's go with `fs.watchFile` then

aviatesk (approved):

PR #1757, opened 2020-11-10
search project file in JS

aviatesk:
so that we don't need to wait for REPL to boot

~~requires https://github.com/julia-vscode/LanguageServer.jl/pull/846~~ no longer required, we use extension code instead


pfitzseb:
Hmm. Not sure how I feel about this.

Should we maybe do all path handling straight in VSCode instead?

aviatesk:
> Should we maybe do all path handling straight in VSCode instead?

This would be strictly better; why didn't I come up with this ...

Done.

src/interactive/repl.ts
```diff
@@ -585,21 +590,33 @@ async function activateHere(uri: vscode.Uri) {
...
             console.log(err)
         }
     }
 }
 
+async function searchUpFile(target: string, from: string): Promise<string> {
+    const parentDir = path.dirname(from)
+    // XXX: on windows ?
+    if (parentDir === from || // ensure to escape infinite recursion
+        from === '/'          // reached to the system root
```
aviatesk, line 611: I'm not sure what this condition is enough for windows; any idea on this ?
```js
$ node
> const path = require('path')
> path.dirname('\\')
```
on Windows code would show what is needed here.

pfitzseb: I'm getting 
```
> const path = require('path')
undefined
> path.dirname('\\')
'\\'
```
on Windows, so I think we do need an additional check

pfitzseb (approved):

pfitzseb:
Can you fix the conflicts, @aviatesk?

src/interactive/repl.ts
pfitzseb: Nvm, that means the `parentDir === from` check is enough...

aviatesk: Thanks for checking for me !
Now I think we actually even don't need this check:
https://github.com/julia-vscode/julia-vscode/blob/9a1a6af8a973d2e9fee3f21f1a115fb7627358b3/src/interactive/repl.ts#L621

because
```js
 node                             
Welcome to Node.js v14.6.0.
Type ".help" for more information.
> const path = require('path')
undefined
> path.dirname('/')
'/'
```

Sorry for the noise.

PR #1778, opened 2020-11-19
Note that julia.environmentPath needs a VS Code reload to take effect

mortenpi:
I was trying to get the REPL to start in a specific directory, so I was trying out different settings for `julia.environmentPath`. But I was just restarting the REPL, so it looked like `julia.environmentPath` was not having any effect. Thanks to @aviatesk on Slack I now realize what's going on (I incorrectly assumed that the environment settings applied just to the REPL), but it might be good to mention something like this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -598,7 +598,7 @@
                         "null"
                     ],
                     "default": null,
-                    "description": "Path to a julia environment.",
+                    "description": "Path to a julia environment. VS Code needs to be reloaded for changes to take effect.",
                     "scope": "window"
                 },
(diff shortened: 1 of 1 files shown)


pfitzseb (approved):
Thanks! Let's get this in for now, although I think we can do better and remove the need for a reload.

aviatesk:
Well, if we change the environment via our commands rather than directly manipulating the configuration file, the configuration will be reflected within the active window; it's still better we can watch for the changes in the configuration files though.

PR #1779, opened 2020-11-19
allow string arg in executeFile

pfitzseb:
Allows keybindings like
```
    {
        "key": "ctrl+shift+t",
        "command": "language-julia.executeFile",
        "args": "/home/pfitzseb/test.jl"
    }
```

### DIFF
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -274,7 +274,7 @@ function clearProgress() {
     }
 }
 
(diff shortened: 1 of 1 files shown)


aviatesk (approved):
I didn't know that vscode allows an user to specify custom arguments to each command; how powerful !

pfitzseb:
Yeah, same, until someone on slack pointed this out.

PR #1785, opened 2020-11-20
jupyter compatible keybindings

pfitzseb:
This switches our code-eval keybindings to something close to Jupyter and makes sure our kbds aren't used in jupyter notebooks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -362,14 +362,29 @@
         },
         "keybindings": [
             {
-                "command": "language-julia.executeJuliaCodeInREPL",
-                "key": "ctrl+Enter",
-                "when": "editorTextFocus && editorLangId == julia"
+                "key": "ctrl+enter",
+                "command": "language-julia.executeCodeBlockOrSelection",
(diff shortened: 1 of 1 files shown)


aviatesk (approved):
lgtm, I would like to wait for approve from @davidanthoff before merging tho

aviatesk:
Could you take a look at this, @davidanthoff ?
Well I think we can move forward with this anyway; afaiu the keybindings are just compatible with the Jupyter extension's.

davidanthoff:
I would do this once the notebook stuff in base VS Code has shipped and we are sure that we know what their final choices will be.

davidanthoff:
Or wait, is this not actually updating our keybindings, but just making sure that they are not used in the notebook UI? In that case, I'm in favor, of course.

pfitzseb:
I was sneaky -- it does both. I actually thought the November release would finalize the Notebook API, but it turns out it didn't. It does seem *very* unlikely that the notebook keybindings will change though.

PR #1797, opened 2020-11-25
fix inline plot display with plotpane disabled

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1795.

### DIFF
--- a/scripts/packages/VSCodeServer/src/display.jl
+++ b/scripts/packages/VSCodeServer/src/display.jl
@@ -105,7 +105,7 @@ function Base.display(d::InlineDisplay, x)
             end
         end
     else
-        with_no_default_display(() -> display(x))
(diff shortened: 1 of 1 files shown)


scripts/packages/VSCodeServer/src/display.jl
```diff
@@ -87,6 +104,8 @@ function Base.display(d::InlineDisplay, x)
                 return display(d, mime, x)
             end
         end
+    else
+        with_no_default_display(() -> display(x))
```
aviatesk, line 108:
```suggestion
        return with_no_default_display(() -> display(x))
```

or remove the succeeding `throw` call.

aviatesk (changes requested):

aviatesk (approved):

PR #1798, opened 2020-11-26
improve LS startup notification

pfitzseb:


aviatesk (approved):
lgmt, but why is this needed ?

pfitzseb:
Because you could end up with multiple status bar items under certain circumstances. Would probably make sense to track down when that happens, but this at least looks better...

aviatesk:
I see, let's go merge for now ;)

PR #1806, opened 2020-12-04
launch oop debugger with startup file

pfitzseb:


davidanthoff:
This kind of has the potential to mess things up, right? I think we can try it, but a more robust implementation might be to not pass the flag, but instead run the startup script after the whole debugger has been initialized. And maybe have an option for this in the launch config? But for now I'd say we can give it a shot as is.

pfitzseb:
Ah, fair point. I'll change this to load startup.jl in our launch script instead.

davidanthoff (approved):
This looks good. I could also imagine that we move the call to `load_julia_startup` to even later, essentially just before the message queue starts to run. But maybe lets try this one first here, my worry might well be a bit academic :)

PR #1812, opened 2020-12-07
fix bt cropping for inline results

pfitzseb:
Turns out I broke this in the last release.
image

### DIFF
--- a/scripts/packages/VSCodeServer/src/misc.jl
+++ b/scripts/packages/VSCodeServer/src/misc.jl
@@ -18,7 +18,7 @@ function find_first_topelevel_scope(bt::Vector{<:Union{Base.InterpreterIP,Ptr{Cv
             linfo = frame.linfo
             if linfo isa Core.CodeInfo
                 linetable = linfo.linetable
-                if isa(linetable, Vector{Core.LineInfoNode}) && length(linetable) ≥ 1 && first(linetable).method === Symbol("top-level scope")
+                if isa(linetable, Vector) && length(linetable) ≥ 1 && first(linetable).method === Symbol("top-level scope")
                     return true
                 end
             end
--- a/scripts/packages/VSCodeServer/src/misc.jl
+++ b/scripts/packages/VSCodeServer/src/misc.jl
@@ -11,8 +11,21 @@ function find_frame_index(bt::Vector{<:Union{Base.InterpreterIP,Ptr{Cvoid}}}, fi
     end
     return
 end
-
-
+function find_first_topelevel_scope(bt::Vector{<:Union{Base.InterpreterIP,Ptr{Cvoid}}})
+    for (i, ip) in enumerate(bt)
+        st = Base.StackTraces.lookup(ip)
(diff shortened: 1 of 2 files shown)


aviatesk (commented):
I prefer looking for the toplevel frame to crop a stacktrace, but it seems that our stacktrace rendering is still broken ? 

> On this branch
Screen Shot 2020-12-08 at 21 48 42

It only seems to relate to frontend code (and so we don't necessarily address that on this PR) though. The cropping of stacktrace works nicely.

pfitzseb:
Someone complained about this on Slack too, but I can't repro that on either Linux ro Windows...

scripts/packages/VSCodeServer/src/misc.jl
```diff
@@ -11,8 +11,21 @@ function find_frame_index(bt::Vector{<:Union{Base.InterpreterIP,Ptr{Cvoid}}}, fi
...
-
+function find_first_topelevel_scope(bt::Vector{<:Union{Base.InterpreterIP,Ptr{Cvoid}}})
+    for (i, ip) in enumerate(bt)
+        st = Base.StackTraces.lookup(ip)
+        ind = findfirst(st) do frame
+            if frame.linfo isa Base.CodeInfo
+                if frame.linfo.linetable[1].method == Symbol("top-level scope")
+                    return true
+                end
+            end
```
aviatesk, line 22: nit, just improve robustness:
```suggestion
            linfo = frame.linfo
            if linfo isa Core.CodeInfo
                linetable = linfo.linetable
                if isa(linetable, Core.LineInfoNode) && length(linetable) ≥ 1 && first(linetable).method === Symbol("top-level scope")
                    return true
                end
            end
```

aviatesk:
> Someone complained about this on Slack too, but I can't repro that on either Linux ro Windows...

okay, I will try to investigate that this week.

Let's merge this PR for now putting that aside.

scripts/packages/VSCodeServer/src/misc.jl
```diff
@@ -11,8 +11,23 @@ function find_frame_index(bt::Vector{<:Union{Base.InterpreterIP,Ptr{Cvoid}}}, fi
...
-
-
+function find_first_topelevel_scope(bt::Vector{<:Union{Base.InterpreterIP,Ptr{Cvoid}}})
+    for (i, ip) in enumerate(bt)
+        st = Base.StackTraces.lookup(ip)
+        ind = findfirst(st) do frame
+            linfo = frame.linfo
+            if linfo isa Core.CodeInfo
+                linetable = linfo.linetable
+                if isa(linetable, Core.LineInfoNode) && length(linetable) ≥ 1 && first(linetable).method === Symbol("top-level scope")
```
aviatesk, line 21:
```suggestion
                if isa(linetable, Vector{Core.LineInfoNode}) && length(linetable) ≥ 1 && first(linetable).method === Symbol("top-level scope")
```

aviatesk (approved):

PR #1815, opened 2020-12-10
Fix ReadMe Getting Started list numbering

briochemc:
Currently the "Getting Started" instructions do not look right when I open the extension in VS code. I think this is because the list of instructions is not indented correctly. This PR fixes that.

For the record here is what it currently looks like for me:
Screen Shot 2020-12-10 at 1 41 53 pm
where you can see the numbering is thrown off by the lack of indentation.


pfitzseb:
Thanks!

pfitzseb (approved):

Issue #1278, opened 2020-06-03
[FR] A few missing inline-related commands


marius311:
(From my comment on Slack regarding 0.16.0, split into a few issues)

It would be nice to have bindable commands for:
* [ ] Move up/down by a cell 
* [x] Choose current module (currently could only find how to do so by clicking on the status bar area)
* [ ] `workbench.action.terminal.clear` moves focus to the terminal which is often unwanted as you're doing inline evaluation. might be nice if julia-vscode provided its own clear command that kept focus on the editor.

(edited to reflect things that are done)

pfitzseb:
I think we have all of these now.

PR #1828, opened 2020-12-20
Cell move commands

yha:
This adds commands to move to next and previous cell (part of #1278).
I assigned the shortcuts `ctrl+alt+[` and `ctrl+alt+]` to match the vscode python extension, although the behavior matches Juno (moving to previous cell moves to its last line, rather than the first).
Another option is alt+arrows, like Juno, but vscode uses those for "move line".

### DIFF
--- a/package.json
+++ b/package.json
@@ -392,16 +392,6 @@
                 "when": "editorTextFocus && editorLangId == julia"
             },
             {
-                "command": "language-julia.moveCellUp",
-                "key": "ctrl+alt+[",
-                "when": "editorTextFocus && editorLangId == julia"
-            },
-            {
-                "command": "language-julia.moveCellDown",
-                "key": "ctrl+alt+]",
-                "when": "editorTextFocus && editorLangId == julia"
-            },
-            {
                 "command": "language-julia.clearCurrentInlineResult",
                 "key": "Ctrl+I Ctrl+D",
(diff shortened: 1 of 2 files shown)


pfitzseb:
Very nice, great work!

I'm not the biggest fan of those keybindings, mostly because there's no way to type them on my keyboard (and probably most non-US/UK keyboards).

One idea I had was `Alt-J Alt-C Alt-J`/`Alt-J Alt-C Alt-K` for moving cells and `Alt-J Alt-B Alt-J`/`Alt-J Alt-B Alt-K` for moving blocks, but maybe that's too cute... Alternatively, we could just not add a default keybinding and/or push the decision back until we do a keybinding overhaul.

yha:
I didn't realize these shortcuts are not available on some keyboards. That does seem like a big disadvantage.

> One idea I had was `Alt-J Alt-C Alt-J`/`Alt-J Alt-C Alt-K` for moving cells and `Alt-J Alt-B Alt-J`/`Alt-J Alt-B Alt-K` for moving blocks, but maybe that's too cute...

I think these are just a bit too long rather than too cute. Three keystrokes is a bit much for a movement command.
It seems that `ctrl/shift+alt+PgUp/PgDown` are available. Maybe these could be used for move cell/block up/down. 

> Alternatively, we could just not add a default keybinding and/or push the decision back until we do a keybinding overhaul.

Is there a planned overhaul? If so, I think I'll just remove the keybindings from this PR for now.

pfitzseb:
Well, there is https://github.com/julia-vscode/julia-vscode/pull/1785 which could/should probably be expanded to make all of our keybindings consistent. Removing the keybindings is probably the easiest way to get this merged quickly :P

yha:
OK, I've removed the keybindings for now

pfitzseb (approved):

Issue #1753, opened 2020-11-09
v1.0.8 Connection to server got closed. Server will not be restarted.


steven-varga:
For some reasons it doesn't work, at all. Anyone has similar experience?  

```
Distributor ID:	Linuxmint
Description:	Linux Mint 20
Release:	20
Codename:	ulyana
```

```
Julia Version 1.5.2
Commit 539f3ce943 (2020-09-23 23:17 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i5-5300U CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libimf
  LLVM: libLLVM-9.0.1 (ORCJIT, broadwell)
```

steven-varga:
It is labelled as 'needs more information', is there a pointer how to get debug (the 'more') information, or perhaps to debug the system?

8uurg:
Under the `output` tab in vscode there should be a listing for the Julia Language Server in the dropdown, which should contain any crashes that have occurred. Alternatively, enabling [crash reporting](https://github.com/julia-vscode/julia-vscode/wiki/Privacy-Policy) for the extension could help as well.

steven-varga:
@8uurg  thank you for your quick reply! Below is the printout, if it helps anyone... yes I enabled the crash reporting.  

```
[Info  - 9:38:23 p.m.] Connection to server got closed. Server will restart.
[Info  - 9:38:23 p.m.] Connection to server got closed. Server will restart.
[Info  - 9:38:23 p.m.] Connection to server got closed. Server will restart.
[Info  - 9:38:23 p.m.] Connection to server got closed. Server will restart.
[Error - 9:38:24 p.m.] Connection to server got closed. Server will not be restarted.
```

pfitzseb:
You can try setting `julia.trace.server` to something `messages` or `verbose` and then check the `Julia Language Server trace` output.

That said, it's very weird you're seeing silent crashes. Are you by any chance using a non-default sysimg?

steven-varga:
There must be some single point explanation, this projects is many years old, it must do something right, whatever that maybe  :) but for now it certainly fails with default settings on ubuntu 20.04 based system, with little output. No custom Julia image used, except Julia is linked against Intel MKL. Starting VSCode from console -- where proper  MKL is guaranteed -- makes no difference

Here is the server trace, top few lines, if it helps anyone.
```
[...]

PR #1839, opened 2020-12-30
fixes: #1753

steven-varga:
Intel MKL linked julia requires MKL related path variables present, not preserving the original environment will lead to silent, hard to debug, crashes. In the PR
the original environment is augemnted with julia vs-code environment, allowing Intel MKL compiled julia be used as a back end.

Im addition this PR may fix other pending issues related to JULIA_DEPOT_PATH, by preserving the old_depot_path, and adding a closing ':' token to allow julia to use its default
DEPOT_PATH mechanism.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -151,12 +151,12 @@ async function startLanguageServer() {
     const serverArgsDebug = ['--startup-file=no', '--history-file=no', '--depwarn=no', `--project=${envForLSPath}`, 'main.jl', jlEnvPath, '--debug=yes', telemetry.getCrashReportingPipename(), oldDepotPath, g_context.globalStoragePath]
     const spawnOptions = {
         cwd: path.join(g_context.extensionPath, 'scripts', 'languageserver'),
-        env: {
-            JULIA_DEPOT_PATH: path.join(g_context.extensionPath, 'scripts', 'languageserver', 'julia_pkgdir'),
+        env: Object.assign(process.env, {
+            JULIA_DEPOT_PATH: path.join(g_context.extensionPath, 'scripts', 'languageserver', 'julia_pkgdir') + ':' + oldDepotPath,
             JULIA_LOAD_PATH: process.platform === 'win32' ? ';' : ':',
             HOME: process.env.HOME ? process.env.HOME : os.homedir(),
             JULIA_LANGUAGESERVER: '1'
-        }
+        })
     }
 
     const jlexepath = await juliaexepath.getJuliaExePath()
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -65,10 +65,10 @@ async function startREPL(preserveFocus: boolean, showTerminal: boolean = true) {
             return jlarg2
         }
 
-        const env = {
+        const env = Object.assign(process.env, {
             JULIA_EDITOR: get_editor(),
             JULIA_NUM_THREADS: inferJuliaNumThreads()
-        }
+        })
 
         const pkgServer: string = vscode.workspace.getConfiguration('julia').get('packageServer')
         if (pkgServer.length !== 0) {


src/extension.ts
```diff
@@ -151,12 +151,12 @@ async function startLanguageServer() {
     const serverArgsDebug = ['--startup-file=no', '--history-file=no', '--depwarn=no', `--project=${envForLSPath}`, 'main.jl', jlEnvPath, '--debug=yes', telemetry.getCrashReportingPipename(), oldDepotPath, g_context.globalStoragePath]
     const spawnOptions = {
         cwd: path.join(g_context.extensionPath, 'scripts', 'languageserver'),
-        env: {
-            JULIA_DEPOT_PATH: path.join(g_context.extensionPath, 'scripts', 'languageserver', 'julia_pkgdir'),
+        env: Object.assign(process.env, {
+            JULIA_DEPOT_PATH: path.join(g_context.extensionPath, 'scripts', 'languageserver', 'julia_pkgdir') + ':' + oldDepotPath,
```
pfitzseb, line 155: Are you sure appending `oldDepotPath` is required for this to fix your problem? This process shouldn't ever load random packages, and Base Julia shouldn't care about the DEPOT_PATH at all, no?

steven-varga: In my understanding the launched `julia` process needs to be identical to any other `julia` REPL launched on a given system to able to interact with other system libraries. My experience with not including `JULIA_DEPOT` path was all packages got re-installed into the directory where `code-server` had been running from -- re-introducing the original JULIA_DEPOT and a `:` colon reverted back to the original, expected behaviour of julia REPL; now launched from `code-server`

Is including `oldDepotPath` against something I am not aware of?

pfitzseb: This process isn't user visible and should only load packages present in `scripts/packages`, so I don't think the behaviour you were seeing really was because of this.

steven-varga: OK, let me know when you identified the reason.

pfitzseb: Ok, then let's just remove this for now so the fix for the MKL issue can be merged. We haven't gotten any reports about `DEPOT_PATH` issues with the LS process and I'm not seeing any myself.

src/extension.ts
```diff
@@ -151,12 +151,12 @@ async function startLanguageServer() {
     const serverArgsDebug = ['--startup-file=no', '--history-file=no', '--depwarn=no', `--project=${envForLSPath}`, 'main.jl', jlEnvPath, '--debug=yes', telemetry.getCrashReportingPipename(), oldDepotPath, g_context.globalStoragePath]
     const spawnOptions = {
         cwd: path.join(g_context.extensionPath, 'scripts', 'languageserver'),
-        env: {
+        env: Object.assign(process.env, {
```
davidanthoff, line 154: I'm quite nervous about this one, to be honest... One of the key design decisions around LS was that we isolate it completely from user configuration settings that might hang around on a system, and with this one here now we are pulling _everything_ in that a user might have configured in terms of env vars for every user, which I think we should not do. One of the reasons why we haven't had many problems about configuration is that we have been so strict about that.

Could we do a more targeted approach here, or something that is more configurable? For example, we could have an configuration option `advancedLSenv`, which is a string, empty be default, that gets added here. Or we have a bool configuration option `advancedPassSystemEnvToLS` that is `false` by default, or something like that.

src/interactive/repl.ts
```diff
@@ -65,10 +65,10 @@ async function startREPL(preserveFocus: boolean, showTerminal: boolean = true) {
             return jlarg2
         }
 
-        const env = {
+        const env = Object.assign(process.env, {
```
davidanthoff, line 68: I think this one is good, we probably should have done that a long time ago :)

davidanthoff (changes requested):

src/extension.ts
steven-varga: @davidanthoff  thanks for the observation. I am not familiar with your code base, including any of the mechanism to drive it. I find it useful and performing well -- with the exception of **silent crashes** with other engineered solutions such as Intel MKL. These modifications including the `:` colon with Julia Depot fixes my problems -- also passes ALL TEST cases you have on your CI.

As a consumer to this open sourced excellent work I uploaded the diffs, with good intentions: to help others from silent crashes. The change was only possible as @pfitzseb  [showed me](https://discourse.julialang.org/t/vscode-julia-doesnt-find-intel-mkl-shared-objects/52232/8?u=steven_varga) how to compile these packages -- as I am not familiar with the framework/workflow. In my opinion, more qualified people such as you and @pfitzseb should use it or not as you wish. 

In a [different thread](https://github.com/julia-vscode/julia-vscode/pull/1839/files/9b4e3b34b9fc8159a2a4bdc077cadc2e923ad868#r552692845) @pfitzseb asked to remove the addition of ':' based on similar argument. Combining the two modifications, we got zero line.

src/interactive/repl.ts
pfitzseb: Yeah, I was wondering about that too. This should be a no-op, except if VSCode messes with `process.env` -- [the docs](https://code.visualstudio.com/api/references/vscode-api#TerminalOptions) say that the environment should be inherited automatically
>env?: Object with environment variables that will be added to the VS Code process.
> strictEnv?: boolean
> Whether the terminal process environment should be exactly as provided in TerminalOptions.env. When this is false (default), the environment will be based on the window's environment and also apply configured platform settings like terminal.integrated.windows.env on top. When this is true, the complete environment must be provided as nothing will be inherited from the process or any configuration.

And it is, in most cases -- on cursory inspection the integrated REPL shows a normal environment. We did get reports about weird subtle differences between a shell env and the env in the REPL though, which I've always chalked up to us not starting Julia through a shell.

That said, doing this shouldn't hurt anything.

src/extension.ts
pfitzseb: As below, this *should* be a no-op. So we could just run with it if it actually fixes the MKL issue.

PR #1870, opened 2021-01-20
Add support for tagging package versions

davidanthoff:
This makes the `tag` functionality from PkgDev.jl available through the extension.

I'm pretty sure there are still some corner cases that this doesn't handle properly, and we could probably also improve on the UI, but on the other hand it really, really simplifies tagging a new version :) And one major benefit of using this from VS Code rather than directly as a package is that we can get the authentication info from VS Code, so that is a lot less brittle than the normal way in PkgDev of trying to get the credentials from the git credential manager.

Not sure, should we maybe just merge and leave it with the `experimental` label and then fix things as they come up?

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
         "onCommand:language-julia.cdHere",
         "onCommand:language-julia.activateHere",
         "onCommand:language-julia.activateFromDir",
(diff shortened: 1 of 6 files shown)


ZacLN (approved):
Looks good to me

pfitzseb (approved):

PR #1872, opened 2021-01-22
Live unit testing

davidanthoff:
This adds a new task that does "live unit testing". Essentially it runs the tests in your package, and whenever one edits the tests or the package and saves, it reruns the test suite.

I think there really is only one thing to figure out:
- [x] Activate the correct test environment for the package that needs testing.


pfitzseb (approved):

davidanthoff:
FYI @timholy, this of course uses Revise.jl to work :)

PR #1873, opened 2021-01-23
Add option for live test file

davidanthoff:
If one has a test suite that genuinely takes a long time to run, then this options allows one to specify a different test file that has some unit tests that might run quicker specifically for live testing.


davidanthoff:
I'm going to merge without review, I want to push an insider build out before I go to bed :)

PR #1876, opened 2021-01-23
Update node packages

davidanthoff:
This updates all node packages to the latest version, and all of these updates are with version jumps that are breaking.

Not really sure how we should test this... But for some of the packages we are using we were on versions that had security problems and are no longer maintained, so I think at some point we have to update these and then just fix what comes up...

EDIT: Actually, this doesn't even compile, so I'll try to fix that first.
EDIT2: And now it compiles.


pfitzseb (approved):
Cool! Haven't tried this, but should be good to go if the extension loads and you tried a few features. ;)

PR #1882, opened 2021-01-24
better plotpane image scaling

jkrumbiegel:
With CairoMakie I'm usually annoyed that SVG's are not scaled and therefore shown way too big / cut off. Also, png images are scaled to fit but only horizontally, so if the plot pane is rather wide but not tall, images are cut off at the bottom.

This tries to fix both of those problems.

SVG's are put into an image tag if they appear to be valid for that purpose (have the xmlns tag). This image tag is then scaled such that the image either vertically or horizontally fits into the plot pane if the image is bigger than either dimension.

If the image is shorter in both dimensions than the available space, it is not scaled but shown at its true size (via `object-fit: scale-down`).

Here are two examples showing how a big svg is scaled:

```julia
let
    svg = String(read(download("https://jkrumbiegel.github.io/assets/2020-11-15-makielayout/example_1.svg")))
    display(MIME"image/svg+xml"(), svg)
end
```

grafik

grafik

### DIFF
--- a/src/interactive/plots.ts
+++ b/src/interactive/plots.ts
@@ -134,21 +134,48 @@ export function plotPaneDelAll() {
     }
 }
 
+// wrap a source string with an <img> tag that shows the content
+// scaled to fit the plot pane unless the plot pane is bigger than the image
+function wrap_imagelike(srcstring: string) {
+    const html = `
+    <html style="padding:0;margin:0;">
+        <body style="padding:0;margin:0;">
+            <div style='width: 100%; height: 100vh'>
+                <img style='display:block; height: 100%; width: 100%; object-fit: scale-down; object-position: 0 0;' src='${srcstring}'>
+            </div>
+        </body>
+    </html>`
+    return html
+}
+
 export function displayPlot(params: { kind: string, data: string }) {
     const kind = params.kind
     const payload = params.data
 
     if (kind === 'image/svg+xml') {
-        g_currentPlotIndex = g_plots.push(payload) - 1
+        const has_xmlns_attribute = payload.includes("xmlns=");
+        var plotPaneContent: string;
+        if (has_xmlns_attribute) {
+            // the xmlns attribute has to be present for data:image/svg+xml to work
+            // encodeURIComponent is needed to replace all special characters from the SVG string
+            // which could break the HTML
+            plotPaneContent = wrap_imagelike(`data:image/svg+xml,${encodeURIComponent(payload)}`);
+        } else {
+            // otherwise we just show the svg directly as it's not straightforward to scale it
+            // correctly if it's not in an img tag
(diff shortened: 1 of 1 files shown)


pfitzseb:
> the xmlns attribute has to be present for data:image/svg+xml to work

Really? I was fairly certain that data URIs work for all SVGs, but maybe that was by accident.

jkrumbiegel:
Yeah I tried it and it made a difference for me. I can't find the original link anymore where I read it but here the answer also mentions it https://stackoverflow.com/questions/18467982/are-svg-parameters-such-as-xmlns-and-version-needed

src/interactive/plots.ts
```diff
@@ -134,21 +134,48 @@ export function plotPaneDelAll() {
...
+}
+
 export function displayPlot(params: { kind: string, data: string }) {
     const kind = params.kind
     const payload = params.data
 
     if (kind === 'image/svg+xml') {
-        g_currentPlotIndex = g_plots.push(payload) - 1
+        const has_xmlns_attribute = payload.includes("xmlns=");
+        var plotPaneContent: string;
```
pfitzseb, line 157:
```suggestion
        let plotPaneContent: string;
```

src/interactive/plots.ts
```diff
@@ -134,21 +134,48 @@ export function plotPaneDelAll() {
...
+    return html
+}
+
 export function displayPlot(params: { kind: string, data: string }) {
     const kind = params.kind
     const payload = params.data
 
     if (kind === 'image/svg+xml') {
-        g_currentPlotIndex = g_plots.push(payload) - 1
+        const has_xmlns_attribute = payload.includes("xmlns=");
```
pfitzseb, line 156:
```suggestion
        const has_xmlns_attribute = payload.includes('xmlns=');
```
The check here isn't particularly robust, but that's fine I think.

src/interactive/plots.ts
```diff
@@ -134,21 +134,48 @@ export function plotPaneDelAll() {
...
+        var plotPaneContent: string;
+        if (has_xmlns_attribute) {
+            // the xmlns attribute has to be present for data:image/svg+xml to work
```
pfitzseb, line 159:
```suggestion
            // the xmlns attribute has to be present for data:image/svg+xml to work (https://stackoverflow.com/questions/18467982)
```

pfitzseb (changes requested):
Great to have this! Left a few suggestions inline, but otherwise this is good to go.

jkrumbiegel:
I also thought the `xmlns` check was rather hacky but I'm not sure what all the different variants can be that work, so I thought this would have a reasonably low false-negative rate of detecting svgs that work in the img tag.

pfitzseb (approved):

PR #1883, opened 2021-01-24
Replace all uses of single arg JSON RPC types

davidanthoff:


davidanthoff:
@julia-vscode/core Review for this after the fact would be great.

PR #1888, opened 2021-01-26
UI for setting compiled funcs/mods

### LINKED ISSUE #1880: Improve debugging performance
https://discourse.julialang.org/t/debugging-in-juno-extremely-slow/53801/4?u=tim.holy suggests a way that might be useful for many users. The suggestion later in the thread to put this under control of a checkbox seems to make sense.

pfitzseb:
Requires https://github.com/julia-vscode/DebugAdapter.jl/pull/31.
Fixes #1880.

Peek 2021-01-26 22-20

There are still a few issues with the UI and I'd like to add an input because the default search is kinda bad (too fuzzy).


pfitzseb:
Improved the UI/UX a bunch:
image

This still only works for the attached debugger, but feels pretty good so far. I've also added some better defaults, but haven't played around with them too much.

davidanthoff:
This is amazing!

Should we remove the UI for compile mode in the breakpoints section then?

pfitzseb:
Hm, we maybe can move the functionality to our own UI, but I'm not as confident in my own sync code compared to the breakpoint handling. Will play around with adding an extra button to toggle compiled mode though.

pfitzseb:
Now supports setting functions as interpreted, so we can force all of Base to run in compiled mode except for some functions (I've set a list of default functions here, but not sure how sensible that one is). Requires https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/468.

pfitzseb:
Alright, this is the workflow for whitelisting certain functions in a compiled module to be interpreted:
Peek 2021-02-25 13-02

I've also added a button for temporarily toggling compiled mode (the red breakpoint looking thing), so we don't abuse the breakpoint UI for that anymore. I think this should be good to go into the next minor in this state.

package.json
```diff
@@ -807,6 +908,62 @@
...
+                        "-Base.redirect_stderr",
+                        "-Base.redirect_stdin",
+                        "-Base.redirect_stdout",
+                        "-Base.reenable_sigint",
+                        "-Base.setindex!",
+                        "-Base.setprecision",
+                        "-Base.setrounding",
+                        "-Base.show",
+                        "-Base.sprint",
+                        "-Base.sum!",
```
oxinabox, line 950: Do we need 
```suggestion
                        "-Base.sum",
                        "-Base.sum!",
```

package.json
```diff
@@ -807,6 +908,62 @@
...
+                        "-Base.timedwait",
+                        "-Base.withenv",
+                        "Core",
+                        "Core.Compiler.",
+                        "Core.IR",
+                        "Core.Intrinsics",
+                        "DelimitedFiles",
+                        "Distributed",
+                        "LinearAlgebra.",
+                        "Serialization",
```
oxinabox, line 961:
```suggestion
                        "Serialization",
                        "Statistics",
                        "-Statistics.mean",
```

package.json
```diff
@@ -807,6 +908,62 @@
...
+                        "-Base.atreplinit",
+                        "-Base.cd",
+                        "-Base.disable_sigint",
+                        "-Base.filter",
+                        "-Base.filter!",
+                        "-Base.findall",
+                        "-Base.findfirst",
+                        "-Base.findlast",
+                        "-Base.findnext",
+                        "-Base.findprev",
```
oxinabox, line 931: For comprehendsion to work we need to do:
```suggestion
                        "-Base.findprev",
                        "-Base.Generator",
```
right?

package.json
```diff
@@ -807,6 +908,62 @@
...
                     "description": "A workspace relative path to a Julia file that contains the tests that should be run for live testing.",
                     "scope": "window"
+                },
+                "julia.debuggerDefaultCompiled": {
+                    "type": "array",
+                    "default": [
+                        "Base.",
+                        "-Base.!",
+                        "-Base.all",
+                        "-Base.all!",
```
oxinabox, line 918: Are these by defining module, or by namespace module?
i.e. will this also compile my overoads of `Base.map(f, ::MyType)`  etc?

pfitzseb: These are for all methods of that function, so `-Base.map` will force interpreted mode for all overloads.

oxinabox: We should look into working out how to do it by definitition location, but that can be a follow up PR?

package.json
```diff
@@ -807,6 +908,62 @@
...
+                        "-Base.cd",
+                        "-Base.disable_sigint",
+                        "-Base.filter",
+                        "-Base.filter!",
+                        "-Base.findall",
+                        "-Base.findfirst",
+                        "-Base.findlast",
+                        "-Base.findnext",
+                        "-Base.findprev",
+                        "-Base.include_string",
```
oxinabox, line 932: Probably we don't want custom iterators to be compiled? or maybe we are ok with that.
```suggestion
                        "-Base.include_string",
                        "-Base.iterate",
```

package.json
pfitzseb: Well, this list could be method based instead of function based, but I'm not sure whether that's particularly useful.

package.json
pfitzseb: I feel like most people probably don't care about `iterate` (and they can just remove it from the list if they do care), but maybe I'm wrong about that.

package.json
pfitzseb: Yes, but unfortunately that's not enough.
`[f(x) for x in x]` ends up calling 
```
(::Type{Base.Generator})(f::F, iter::I) where {I, F} in Base at generator.jl:32
```
which we catch fine with the above definition, but that then calls the parametric constructor
```
(::Type{Base.Generator{I,F}})(f, iter) where {I, F} in Base at generator.jl:32
```
which isn't included in `methods(Base.Generator)`. So I'm not sure where to get that from...

timholy:
I do think there's some reason to think seriously about interpreting `-Statistics.mean` to signify "all methods of `Statistics.mean` that are owned by `Statistics`." That said, perhaps it doesn't much matter: if I've written a specialization in a package, it's unlikely I mix that with calls to methods defined in other modules.

davidanthoff (changes requested):
Fantastic, just needs a merge conflict resolution.

package.json
pfitzseb: Was just missing an `unwrap_unionall` here.

pfitzseb:
Ok, breakpoints now work at least in `foreach`/`map`/generators/loops on Julia 1.5. Also fixed an issue with frame cache invalidation.

davidanthoff (approved):

davidanthoff (changes requested):
Needs merge conflict resolution. Looks to me as if someone updated the submodules on the branch, which is never a good idea :)

pfitzseb:
Yeah, I checked in submodules because this needed a branch of DebugAdapter to work. Should be all good now :)

davidanthoff (approved):

Issue #1030, opened 2020-03-03
Add an option to connect a running REPL instance to the extension


davidanthoff:
This came out of the discussion at https://github.com/julia-vscode/julia-vscode/issues/1027.

The idea would broadly be that one can load some package in a REPL, and then this REPL would be "connected" to VS Code, so that one can send code pieces to that REPL, the plot gallery etc all work.

This is a fairly involved project:
- We would probably generally need the notion of multiple running Julia REPLs in the extension (right now we only have one).
- We would need to figure out how a given REPL knows to which VS Code instance it should connect (we should take a look how the node stuff implements the auto debugger attach functionality)

I think we should hold off from doing anything about this until we have finished both a) the debugger stuff (https://github.com/julia-vscode/julia-vscode/pull/957) and b) the workspace UI (https://github.com/julia-vscode/julia-vscode/pull/1003), otherwise this will get too complicated. But eventually it would be good to support this scenario.

haberdashPI:
Great! I've subscribed to those two issues. If I have the resources to contribute to this when it makes sense to start working on it, I'd love to help.

haberdashPI:
I potentially simpler solution occurred to me for my particular use case. It seems like these capabilities would be sufficient for my needs:

1.) The concept of multiple REPL sessions.
2.) An API to programmatically start a new REPL session with a specified name and a specified command for starting said REPL (e.g. `tmux new-session -A -s 'my-project-workspace-julia-session-1' julia`)

I think that approach could solve some of the issues you raise with your second point. The julia REPL would know which VSCode instance to run, because said instance would be the one it was launched by (and presumably there is some process by which, when starting julia you tell it about vscode??).

davidanthoff:
Oh, so you mean that you wouldn't start the Julia REPL from a shell terminal, but instead the julia extension starts it somehow, and then can pass information to that instance on how to connect back to the running extension version? Yes, that would work, that is essentially how we start REPLs currently.

[...]

PR #1893, opened 2021-01-29
support persistent tmux sessions

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1030. I have only tested this on Linux machines (local and remote), which is probably the most common use case anyways.


davidanthoff:
If VS Code Remote is now supporting persistent remote sessions (something @pfitzseb mentioned [here](https://github.com/julia-vscode/julia-vscode/issues/1030#issuecomment-769781162)), do we really want to add our own home grown system here?

pfitzseb:
Yes, I think we do. AFAIU the upstream implementation only works when VSCode is reloaded, not restarted. In general I'd not trust the extension host to keep processes alive over an arbitrary long period of time.

AntoineAsselin78:
Hey good day @pfitzseb, I have been waiting for this pull request for a long long time. You have no idea how much I wanted it. I use TMUX in VSCode but I have not been able to fully enjoy the features in vscode-julia due to the inability to connect Julia extension to TMUX session. So a BIG thank you for creating this!

I am confused how to start tmux session connected to VSCode Julia extension or the other way round. I have enabled VSCode Settings `"julia.persistentSession.enabled":true,` 

I follow the developer instructions here https://github.com/julia-vscode/julia-vscode/blob/master/CONTRIBUTING.md

```bash
git clone --recurse-submodules [thisPullRequestBranch]
git submodule update --init
npm install
vsce package
```

Install language-julia-1.1.10.vsix in VSCode.

Next, do I run `Julia: Start REPL` ? How do I start tmux session connected to VSCode Julia extension or the other way round?

I did that but the same happens as before as if without `"julia.persistentSession.enabled":true,`

pfitzseb:
That should just work, yes (and does for me).
Do note that your workflow only works for *local* sessions -- the remote installs its own copy of the extension, which of course doesn't have the modifications made by this PR.

AntoineAsselin78:
Finally I managed to get it working for my circumstances! I am so happy thank you again @pfitzseb. Finally I get to see inline results, Julia environment pane and debugger!!

I am using [WSL version 2 (not version 1)](https://docs.microsoft.com/en-us/windows/wsl/install-win10#step-2---check-requirements-for-running-wsl-2) and my Julia VSCODE extension is installed in `/home/<user>/.vscode-server/extensions/julialang.language-julia-1.1.10`

So following your instructions, I set
1. `julia.persistentSession.enabled: true`
2. `Julia: Start REPL`
3. starts a new tmux session named set by `julia.persistentSession.tmuxSessionName`

However, the typical way of using TMUX in VSCODE, we follow this instructions by Senior Software Engineer working on VS Code: [Persistent terminal sessions in VS Code](https://medium.com/@joaomoreno/persistent-terminal-sessions-in-vs-code-8fc469ed6b41)
1. start VSCODE first
2. then automatically launch tmux set by VSCODE settings `terminal.integrated.shell.linux: /home/<user>/path/to/custom_tmux_startup_script` where `custom_tmux_startup_script` starts up session automatically for instance `tmux new-session -d -s XXX.......`

My question is, for instance, my Julia session named `Existing_Julia_Session` already started by `custom_tmux_startup_script`. How do I connect Julia extension from `Existing_Julia_Session` (instead of using `Julia: Start REPL`), for instance, can I send a command such as `connect_this_julia_session_to_vscode_extension()` from my existing Julia REPL?

Hope my question is clear not and confusing.

Using `Julia Version 1.6.0`

pfitzseb:
In theory yes, in practice it's hard to figure out the pipe name yourself (I'm planning on adding a command that shows instructions on how to manually connect to an arbitrary Julia process though).

That said, I'd recommend just letting us handle the initial tmux session creation since we're doing the exact same thing the linked startup script does.

pfitzseb:
Ok, I just pushed a commit that adds a "Connect external REPL" command.

davidanthoff (commented):
So if one configures this, then _all_ VS Code instances will share one Julia REPL, right? Because each instance will connect to the already running tmux session? Is that good? Does that actually properly work? With things like progress reporting, plot panes etc?

I'm a bit hesitant about this whole approach: it just seems a bit too specific to me...

pfitzseb:
Well, the session-name setting is `machine-overridable`, so you can simply change it on a per-instance/per-workspace/per-folder basis. 
When multiple instances try to connect to one tmux session the last one trying will win out, all others are disconnected (although they'll still show up in the terminal pane, of course). So no, none of our integration features will work in multiple VSCode instances.

> I'm a bit hesitant about this whole approach: it just seems a bit too specific to me...

It does solve a very real problem though. I can see where you're coming from, but I suspect that people who value persistent session might be happy to change the session name if they use multiple VSCode instances.

We could also make this a two-step process where we show a confirmation dialog before connecting to the persistent session. Would that make the PR more palatable to you? I'd really like to get this in :)

davidanthoff (approved):
Lets jus try it for now, and I think it is labeled "experimental", right? If the experiment fails, we can still remove it again ;) Or maybe VS Code Remote adds proper support for this at some point.

PR #1896, opened 2021-01-30
Fix for #1895 Benchmark task MethodError

### LINKED ISSUE #1895: Benchmark task MethodError
When trying to run the benchmark task I get the following error:
`ERROR: MethodError: no method matching benchmarkpkg(::String, ::BenchmarkConfig; promptsave=false, promptoverwrite=false)`

It looks like the `promptsave` and `promptoverwrite` keywords were replaced with `overwrite` in JuliaCI/PkgBenchmark.jl#40 and then removed completely for v0.1.0  in JuliaCI/PkgBenchmark.jl#57.  

From [their documentation](https://juliaci.github.io/PkgBenchmark.jl/stable/run_benchmarks/) It looks like it now takes the keyword argument `resultfile` instead:

```
resultfile - If set, saves the output to resultfile
```

I think removing the `promptsave` and `promptoverwrite` arguments in 67 of src/tasks.ts
https://github.com/julia-vscode/julia-vscode/blob/2d38dd17ea4e711cfd2898cabfaab48d350ca373/src/tasks.ts#L67
by changing it to this should fix the error. 
```julia
let benchmarkTask = new vscode.Task({ type: 'julia', command: 'benchmark' }, folder, `Run benchmark`, 'julia', new vscode.ProcessExecution(jlexepath, ['--color=yes', '-e', 'using PkgBenchmark; benchmarkpkg(Base.ARGS[1], resultfile=" ")', package_name]), "");
```
I've left the `resultfile` string empty as I don't know what the previous naming convention was, but once that is filled out it should have the same behavior as before.

Oliver-Leete:
Hopefully this should fix #1895. 

I'm not too sure what to put for the name and directory of the results file, or if there is some way of having the user pass this into the task as an argument. At the moment I've set it to save it in the benchmark subdirectory as that seems to make sense to me, but if anyone knows what the old behaviour was it might be worth changing it to match that.

### DIFF
--- a/src/tasks.ts
+++ b/src/tasks.ts
@@ -71,7 +71,7 @@ class JuliaTaskProvider {
             }
 
             if (await fs.exists(path.join(rootPath, 'benchmark', 'benchmarks.jl'))) {
-                const benchmarkTask = new vscode.Task({ type: 'julia', command: 'benchmark' }, folder, `Run benchmark`, 'julia', new vscode.ProcessExecution(jlexepath, ['--color=yes', `--project=${pkgenvpath}`, '-e', 'using PkgBenchmark; benchmarkpkg(Base.ARGS[1], promptsave=false, promptoverwrite=false)', folder.name]), '')
+                const benchmarkTask = new vscode.Task({ type: 'julia', command: 'benchmark' }, folder, `Run benchmark`, 'julia', new vscode.ProcessExecution(jlexepath, ['--color=yes', `--project=${pkgenvpath}`, '-e', 'using PkgBenchmark; benchmarkpkg(Base.ARGS[1], resultfile="benchmark/results.json")', folder.name]), '')
                 benchmarkTask.presentationOptions = { echo: false, focus: false, panel: vscode.TaskPanelKind.Dedicated, clear: true }
                 result.push(benchmarkTask)
             }


davidanthoff (approved):

davidanthoff:
Awesome, thanks!

I think this broke when https://github.com/JuliaCI/PkgBenchmark.jl/pull/57 was merged. The fact that this happened three years ago suggests to me that you might be the only person using this feature at the moment, so I think we don't have to worry about other users too much ;)

I'm going to merge this as is, because it clearly is an improvement to go from "doesn't work at all" to this.

I could imagine that it might be helpful to use a filename that either has the current date/time in it, or the commit hash or something like that, just so that one can run this multiple times and compare results more easily?

I also think we could do a bunch more things to make this whole experience a lot smoother, but I think that would depend on whether someone wants to tackle that. You maybe? Help would be super welcome! Here are two ideas:
1. we could ship the necessary packages to run benchmarks in the extension, so that things don't depend anymore on having PkgBenchmark.jl installed in the current environment. We're doing that for a lot of things now, and it would make things just "always work". I would probably have to do this, because our method for that isn't really documented :)
2. instead of writing things to a file, the extension could open a webview that displays the results in some nice UI. Or maybe it still does write to disc, but we register a viewer for that file type and if one opens that file it is shown in some nice HTML.
3. it would also be nice to have a UI to compare two result sets, essentially similar to what `judge` does, but integrated into the extension and with a nice UI.

Oliver-Leete:
Yeh, I did think that I might be the only user, part of the reason I looked into exactly when things changed was that I thought it must be me doing something wrong if no-one else had reported it.

I'd definitely be up for having a go at working on this. Although I should warn you that I'm a mechanical engineer and have no prior experience with typescript or any web development, so it might not be the best code. 

2.  I like the idea of a nice UI, although I'm not quite sure what the cleanest way of displaying the data would be. A table would be the obvious choice.
3. I think comparing results would be where dedicated UI would be the most help. I guess the most common use case for this would be comparing the results from two different branches or commits, which PkgBenchmarks' judge function should handle and I'd just have to come up with some pretty visualisation (perhaps a bar chart comparing the averages with error bars to show the ranges. 
A use case that would be of more interest to me (although I have no idea if it would be of use to others) would be to let the user define a stripped down benchmark suite (a benchmark equivalent to the live test file) that could be run often and used for a line plot to visualise the impact of small changes. In this case it would make sense to just store the results temporarily. 

As for me giving this a go, would the VSCode documentation on making an extension be the best place to start?

PR #1898, opened 2021-01-30
Add a cancel token to getModuleAt story

davidanthoff:
My hope is that something like that might solve some of the problems we are seeing around version sync in the `getModuleAt` requests in crash reporting.

Essentially my theory is that between the time when one of the events occurs that triggers an update, and the moment when the actual request is sent to the LS, there are `await` statements in between and during those the document ends up being updated/manipulated, but then those changes aren't sent to the LS yet. So then we end up with a `getModuleAt` request that references a newer doc version than what the LS has. The hope with this would be in such a situation, the request gets actually cancelled before it even is sent to the LS.

I'm not sure it really works, though... Very difficult to test. In particular, it kind of relies on the selection update event also firing when there is a change to the doc content... We might also have to listen to doc change events and then cancel the `getModuleAt` request for those as well... But not sure.

Maybe we ship this in insider for now and see whether it helps?

Medium term we should actually add the ability for these cancel events to traverse over the JSON-RPC barrier as well.


davidanthoff:
I added a retry option now, and also made it not error if the versions are out of sync.

pfitzseb (approved):

PR #1899, opened 2021-01-30
Cancelable debounced getModuleAt with retries

pfitzseb:
Alternative to https://github.com/julia-vscode/julia-vscode/pull/1898.

This makes the `getModuleAt` requests that update the module selector cancelable and debounced (200ms), which should drastically cut down on the number of requests we make. `getModuleAt` in the LS will also never hard-fail; if the documents are out of sync, we'll instead return `'#retry'`, which causes the client to retry up to three times.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -838,6 +838,14 @@
                 "tunnel": "0.0.4",
                 "typed-rest-client": "1.2.0",
                 "underscore": "1.8.3"
+            },
+            "dependencies": {
+                "underscore": {
+                    "version": "1.8.3",
+                    "resolved": "https://registry.npmjs.org/underscore/-/underscore-1.8.3.tgz",
+                    "integrity": "sha1-Tz+1OxBuYJf8+ctBCfKl6b36UCI=",
+                    "dev": true
+                }
             }
         },
         "babel-runtime": {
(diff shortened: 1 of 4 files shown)


pfitzseb:
Now without a sentinel value and a bit more correct.

davidanthoff:
Hm, is this really easier than https://github.com/julia-vscode/julia-vscode/pull/1898?

So it seems to me that using the cancel token story is a) more in line with what is usually done in VS Code, b) I get a bit nervous about the counter stuff being outside of the debouncing story... It seems to me that adding the counter stuff in my PR might be conceptually simpler, if it just happens inside `getModuleForEditor`?

And at least from my reading I got the impression that a cancel token story should generally be more robust than the debouncing approach. Still not perfect, though...

pfitzseb:
Ok, so `getModuleAt` is used in two places -- the module selector and inline evaluation.

The former case is purely visual, basically -- nothing depends on us getting the module right for every cursor position or whatever. The latter case is critical: we need to abort the command if we can't figure out the module. This then gives us the following requirements:

1. Module selector requests should be debounced (so we don't make a request for _every_ cursor move).
2. Module selector requests should be cancelable, so that a newer request stops older ones. This should only happen if the request takes a long time (because of debouncing), or if we make a inline evaluation request (debatable whether that should actually cancel anything though).
3. Inline evaluation requests should never be canceled by newer requests or debounced.

Additionally, all requests should be retried a couple of times as a workaround for the synchronicity issues we were having.

So, that said, I don't particularly care about my implementation here :) I found the `Promise.race` approach slightly nicer than VSCode's `CancellationToken` story, but I'm also not particularly attached to it.

davidanthoff:
So I think my PR sort of does all of that, right? The debounce is kind of implicit: if your selection changes fast enough to cancel the previous request before it goes out to the LS, well, then it gets cancelled :) The requests from the eval parts should never get cancelled because they don't receive a token (I think, should probably double check that). I think right now I'm just thinking it is better to not take a dependency on another library, and instead just do it with the build in, default VS Code patterns?

Should we try to merge my PR and see how it goes? And if we still see problems, we could still switch over to a debouncing story?

PR #1907, opened 2021-02-02
more env improvements

pfitzseb:
This verifies the user supplied env path before returning it in `getAbsEnvPath`. Falls back to the default env path if verification fails. Also properly hides the startup notification if we fail early in `startLanguageServer`.


davidanthoff (approved):
One thing we probably need to revisit in general at some point is the `workspaceFolders[0]` part... Whenever we need a workspace folder, I just did that, but we probably at some point need to systematically go through this and make sure we properly handle multi root workspaces...

Issue #1716, opened 2020-10-24
Surface plot not working in plotly or plotlyjs backend


crlaugh:
When I use the following code to generate surface data in VSCode, I get a plot that doesn't show anything.  This behavior was also referenced in https://github.com/JuliaPlots/Plots.jl/issues/3052.

```
using Plots

xTest = collect(Float16, range(-2,length=100,stop=2));
yTest = collect(Float16, range(sqrt(2),length=100, stop=2));
fEval = zeros(length(xTest), length(yTest))
for (ix, xVal) in enumerate(xTest)
    for (iy, yVal) in enumerate(yTest) 
        fEval[ix, iy] = xVal*yVal-xVal-yVal+1
    end
end

plotly()
surface(fEval)
plot(xTest, yTest, fEval, st=:surface)
```

![image](https://user-images.githubusercontent.com/9616147/97066286-27a52c00-1582-11eb-8d4e-97518450db81.png)

This problematic behavior is only demonstrated in VSCode; if I do the same thing in either Atom or a Julia terminal (all examples in Windows), the plot is correct.  

My VSCode information is: 
Version: 1.50.1 (user setup)
Commit: d2e414d9e4239a252d1ab117bd7067f125afd80a
Date: 2020-10-13T15:06:15.712Z
Electron: 9.2.1
Chrome: 83.0.4103.122
Node.js: 12.14.1
V8: 8.3.110.13-electron.0
OS: Windows_NT x64 10.0.19041

and Julia terminal in VSCode: 
  [c52e3926] Atom v0.12.21
  [aae01518] BandedMatrices v0.15.20
  [336ed68f] CSV v0.7.7
  [5ae59095] Colors v0.12.4
  [8f4d0f93] Conda v1.4.1
  [a93c6f00] DataFrames v0.21.7
  [5789e2e9] FileIO v1.4.3
  [033835bb] JLD2 v0.2.3
  [e5e0dc1b] Juno v0.8.3
  [47be7bcc] ORCA v0.5.0
  [14b8a8f1] PkgTemplates v0.7.10
  [a03496cd] PlotlyBase v0.4.1
  [f0f68f2c] PlotlyJS v0.14.0
  [91a5bcdd] Plots v1.6.4
  [438e738f] PyCall v1.91.4
  [295af30f] Revise v2.7.6 ⚲
  [44cfe95a] Pkg

I am using Julia 1.5.1 in Win10.

pfitzseb:
[...]

PR #1911, opened 2021-02-03
Update plotly

pfitzseb:
and fix auto-resizing for Plotly and VegaLite.

Fixes https://github.com/julia-vscode/julia-vscode/issues/1716.


pfitzseb:
With this you can get automatically resized plots with
```
plot(xTest, yTest, fEval, st=:surface, size=(0,0))
``` 
or
```
@vlplot(
    :point,
    x=:Horsepower,
    y=:Miles_per_Gallon,
    color=:Origin
)
```
whereas 
```
plot(xTest, yTest, fEval, st=:surface)
```
or
```
@vlplot(
    :point,
    x=:Horsepower,
    y=:Miles_per_Gallon,
    color=:Origin,
    width=400,
    height=400
)
```
will show a statically sized plot.

src/interactive/plots.ts
```diff
@@ -212,6 +212,10 @@ export function displayPlot(params: { kind: string, data: string }) {
                         actions: false
                     }
                     var spec = ${payload}
+                    if (!(spec.width || spec.height)) {
+                        spec.width = 'container'
+                        spec.height = 'container'
+                    }
```
davidanthoff, line 218: Hm, I think this is a bit too magical at this point here, I don't think we should modify the spec itself. Or minimally I think this would need to be behind an option that is off by default. I think I could consider making this some sort of default in VegaLite.jl itself, maybe...

davidanthoff (changes requested):

src/interactive/plots.ts
pfitzseb: Hm, if the user doesn't specify a height or width then they clearly don't care about it much and we can offer a better experience than just using VegaLite's defaults, no?

davidanthoff: No, I think if they don't specify it they assume that it will be the default size, the same size that they would get if they save the plot. What I really would want to avoid is that one generates a different plot depending on whether it gets shown in the UI or saved to disc, from the same plot command.

pfitzseb: Fair enough. Removed the auto-resizing for Vega then.

davidanthoff (approved):

PR #1926, opened 2021-02-10
treerender: dispatch on AbstractDict (instead of Dict) --> fix

MarkusSchildhauer:
Since version 1.1.10 the nice tree rendering of special dictionaries (like OrderedDict) in the workspace was gone.
This fix changes the appropriate method dispatch back to AbstractDict.

### DIFF
--- a/scripts/packages/VSCodeServer/src/trees.jl
+++ b/scripts/packages/VSCodeServer/src/trees.jl
@@ -92,7 +92,7 @@ function treerender(x)
     end
 end
 
-function treerender(x::Dict{K,V}) where {K,V}
+function treerender(x::AbstractDict{K,V}) where {K,V}
     treerender(LazyTree(string(nameof(typeof(x)), "{$(K), $(V)} with $(pluralize(length(keys(x)), "element", "elements"))"), wsicon(x), length(keys(x)) == 0, function ()
         if length(keys(x)) > MAX_PARTITION_LENGTH
             partition_by_keys(x, sz=MAX_PARTITION_LENGTH)


pfitzseb:
That was an intentional change in https://github.com/julia-vscode/julia-vscode/pull/1709. I'm not sure if we want to revert this (at least for my specific example only the specialization on `Array` instead of `AbstractArray` is important.

MarkusSchildhauer:
I understand the issue. Being able to browse the object properties is important, but for all sorts of dictionaries a special key-value rendering is definitely desirable. Especially for nested dictionaries. For special Arrays the same consideration may apply.
I propose 3 alternatives to address both requirements:
 1. Introduce settings for this
 2. Make "hackable"  treerender methods that can be specialised externally (e.g. startup.jl) 
 3. Show always both (my favourite). Object properties can be accessible via the top list entry which has a special name like "..", see screenshot below. It's easily implemented, see here: https://github.com/julia-vscode/julia-vscode/compare/master...MarkusSchildhauer:treerender-dict-proposal

If you like this solution I make a pull request.

tree-dict-proposal

pfitzseb:
I like that idea a lot!
Maybe it'd be nicer to use `#props` or something like that instead of `..` and move this to the end, but that's nitpicky. Would love to see a PR for this.

davidanthoff:
@MarkusSchildhauer, yes, I think your 3 is the perfect solution.

MarkusSchildhauer:
I made a pull request for this solution here #1927.

PR #1927, opened 2021-02-11
Treerender proposal for a sorts of dictionaries and arrays

MarkusSchildhauer:
Solves the conflict of interests arising in #1926 and discussed in https://github.com/julia-vscode/julia-vscode/pull/1926#issuecomment-777417836.

All sorts of dictionaries and arrays are rendered primarily displaying their key/index => values. But additionally the properties of the underlying objects are browsable via a top list entry called #properties (using any-icon and the type display style, but that can be changed easily). I recommend leaving the properties entry on top to make it more acceptable for use cases where browsing the properties is more likely (e.g. ODESolutions) and because it could be seen as some sort of meta information.

tree-dict-array-proposal

### DIFF
--- a/scripts/packages/VSCodeServer/src/trees.jl
+++ b/scripts/packages/VSCodeServer/src/trees.jl
@@ -80,26 +80,34 @@ end
 
 getfield_safe(x, f, default=UNDEF) = isdefined(x, f) ? getfield(x, f) : default
 
-function treerender(x)
+struct AnyWrapper
+    x
+end
+treerender(x::AnyWrapper) = invoke(treerender, Tuple{Any,String}, x.x, "")
+
+function treerender(x, typedisplay=typeof(x))
     fields = fieldnames(typeof(x))
 
     if isempty(fields)
-        treerender(Text(string(typeof(x), "()")))
+        treerender(Text(string(typedisplay, "()")))
     else
-        treerender(LazyTree(string(typeof(x)), wsicon(x), function ()
+        treerender(LazyTree(string(typedisplay), wsicon(x), function ()
             collect([SubTree(string(f), wsicon(getfield_safe(x, f)), getfield_safe(x, f)) for f in fields])
         end))
     end
 end
 
 function treerender(x::AbstractDict{K,V}) where {K,V}
-    treerender(LazyTree(string(nameof(typeof(x)), "{$(K), $(V)} with $(pluralize(length(keys(x)), "element", "elements"))"), wsicon(x), length(keys(x)) == 0, function ()
-        if length(keys(x)) > MAX_PARTITION_LENGTH
-            partition_by_keys(x, sz=MAX_PARTITION_LENGTH)
-        else
-            collect([SubTree(repr(k), wsicon(v), v) for (k, v) in x])
-        end
-    end))
+    treerender(LazyTree(string(nameof(typeof(x)), "{$(K), $(V)} with $(pluralize(length(keys(x)), "element", "elements"))"), wsicon(x), length(keys(x)) == 0,
+        () -> pushfirst!(
+            if length(keys(x)) > MAX_PARTITION_LENGTH
+                partition_by_keys(x, sz=MAX_PARTITION_LENGTH)
+            else
+                [SubTree(repr(k), wsicon(v), v) for (k, v) in x]
+            end,
+            SubTree("..", wsicon((;)), AnyWrapper(x)),
+        )
+    ))
 end
 
 function treerender(x::Module)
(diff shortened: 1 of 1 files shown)


scripts/packages/VSCodeServer/src/trees.jl
```diff
@@ -80,28 +80,50 @@ end
...
+
+function treerender(x, typedisplay=typeof(x), icon=wsicon(x))
     fields = fieldnames(typeof(x))
 
     if isempty(fields)
-        treerender(Text(string(typeof(x), "()")))
+        treerender(Text(string(typedisplay, "()")))
     else
-        treerender(LazyTree(string(typeof(x)), wsicon(x), function ()
-            collect([SubTree(string(f), wsicon(getfield_safe(x, f)), getfield_safe(x, f)) for f in fields])
```
MarkusSchildhauer, line 90: I removed collect because it makes a shallow copy of an already allocated array that seems to be of no use here.

pfitzseb: We do actually need the `collect`s  here:
```
julia> x = @SArray [1,2,3]
3-element SArray{Tuple{3},Int64,1,3} with indices SOneTo(3):
 1
 2
 3

julia> [x^2 for (i,x) in zip(keys(x), vec(x))]
3-element SizedArray{Tuple{3},Int64,1,1} with indices SOneTo(3):
 1
 4
 9

julia> collect([x^2 for (i,x) in zip(keys(x), vec(x))])
3-element Array{Int64,1}:
 1
 4
 9
 ```
 I've added a few comments about why it's necessary.

File: scripts/packages/VSCodeServer/src/trees.jl [hunk 2]
```diff
@@ -126,7 +152,8 @@ struct Undef end
 const UNDEF = Undef()
 
 function assign_undefs(xs)
-    xs′ = similar(xs, Any)
+    s = size(xs)
+    xs′ = Array{Any,length(s)}(undef, s...)
```
pfitzseb, line 156: `similar` fails for readonly arrays, so this makes sure we use an  `Array`.

scripts/packages/VSCodeServer/src/trees.jl
```diff
@@ -143,17 +170,7 @@ function assign_undefs(xs)
...
-        if length(x) > MAX_PARTITION_LENGTH
-            partition_by_keys(x, sz=MAX_PARTITION_LENGTH)
-        else
-            collect([SubTree(repr(k), wsicon(v), v) for (k, v) in zip(keys(x), vec(assign_undefs(x)))])
-        end
-    end))
-end
-
-function treerender(err, bt)
+function treerender(err::Exception, bt)
```
pfitzseb, line 173: Needs the `::Exception` because otherwise it overwrites a method definition above.

scripts/packages/VSCodeServer/src/trees.jl
```diff
@@ -80,28 +80,54 @@ end
...
+    treerender(LazyTree(string(typeof(x), " with $(pluralize(size(x), "element", "elements"))"), wsicon(x), length(x) == 0,
+        function ()
+            out = if length(x) > MAX_PARTITION_LENGTH
+                partition_by_keys(x, sz=MAX_PARTITION_LENGTH)
+            else
+                # collect is necessary because the return type of an array comprehension depends on the iterator,
+                # but we only want Arrays here
+                collect([SubTree(repr(k), wsicon(v), v) for (k, v) in zip(keys(x), vec(assign_undefs(x)))])
+            end
+            x isa Array ? out : pushfirst!(out, SubTree("", wsicon(x), PropertyBox(x)))
```
pfitzseb, line 125: No need to show an empty `#properties` item for `Base.Array`s, so let's special case that here.

pfitzseb (approved):
Love this!
I've added two more commits that fix various minor issues (see inline comments). Could also use a CHANGELOG entry, but I can do that later.

scripts/packages/VSCodeServer/src/trees.jl
pfitzseb: Or rather, `similar` of a read-only array produces another read-only array, which then causes a failure on the `setindex!` call below.

MarkusSchildhauer:
Great you approved! Thanks

davidanthoff:
@MarkusSchildhauer Thanks a lot of this PR! Just a note on the process: we are right now in bug-fix mode, so at the moment we are not merging features. But in a few days we'll open up `master` for features again, and then we'll merge this PR right away.

scripts/packages/VSCodeServer/src/trees.jl
MarkusSchildhauer: OMG😯! I was not aware of that occasional iterator type dependency of comprehensions. But makes sense if special indexing is involved.

MarkusSchildhauer:
Nice. But I have it anyway 😉

PR #1929, opened 2021-02-11
Make optional the File Explorer context menu entries

tapeinosyne:
The globally enabled entries in the Explorer right-click menu are a bit intrusive for users who are not working with Julia exclusively or nearly so. This quick patch introduces an option to enable or disable the entries, also in accordance with the newly released [VSCode Extension Guidelines](https://code.visualstudio.com/api/references/extension-guidelines#context-menus).

Currently, the menu entries are disabled by default. An argument can be made that they should remain enabled by default, for ease of use or for the sake of continuity across extension versions.

***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to the Julia extension will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+### Added
+* The global Julia entries in the File Explorer right-click menu can now be toggled on or off.
 
 ## [1.1.10] - 2021-01-28
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -301,15 +301,18 @@
                 },
                 {
                     "command": "language-julia.cdHere",
-                    "group": "julia"
+                    "group": "julia",
+                    "when": "julia.showInExplorerContextMenu && explorerResourceIsFolder"
                 },
                 {
                     "command": "language-julia.activateHere",
-                    "group": "julia"
+                    "group": "julia",
+                    "when": "julia.showInExplorerContextMenu"
                 },
                 {
                     "command": "language-julia.activateFromDir",
-                    "group": "julia"
+                    "group": "julia",
+                    "when": "julia.showInExplorerContextMenu"
                 }
             ],
             "editor/title": [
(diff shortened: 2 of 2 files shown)


davidanthoff:
I'm very much on board that we shouldn't display these all the time! I am wondering, though, whether we could do this without a config setting in a more dynamic way.

For example, the "cd" command could only show up if a Julia REPL is currently open, and the activate commands maybe only if there is actually a `Project.toml` in the workspace, or Julia files, or something like that?

tapeinosyne:
The feasibility of proper contextual behavior depends on what information VS Code exposes to the Explorer context menu. 

I am not familiar with the extension API, but a cursory glance over the documentation suggests `resourceLangId == julia` (already in use elsewhere) and perhaps `debugType == 'julia'` as parameters available to `when` clauses which might provide a reasonable approximation of the behaviors you describe.

I'll see what I can do.

pfitzseb:
https://github.com/julia-vscode/julia-vscode/pull/1933 only shows the Julia specific context menu entries when there's a running REPL.

tapeinosyne:
Thank you, @pfitzseb. I'll rebase my branch on your PR and add some further improvements to this patch, so that it may be merged after yours if need be.

davidanthoff:
So I think the patch from @pfitzseb we just merged should fix the immediate problem. 

BUT, I think it would still be nicer if we could make the activate commands _not_ depend on whether there is a REPL around, but rather on the existence of Project.toml and friend files, right?

pfitzseb:
Activating the current directory to create a Project.toml post-hoc is a valid use case, imho.

davidanthoff:
Also true... Hm, not sure what a good trigger would be... Any Julia find around somewhere, maybe? Because tying it to the REPL also seems not ideal, after all I might want to activate a folder purely for the LS integration, right?

tapeinosyne:
`language-julia.executeFile` can be restricted to Julia files. `language-julia.cdHere` could reasonably appear only if a Julia REPL is active and the target is a folder.

The environment activation commands ("activate this environment" and "activate parent environment") may require some additional logic.

davidanthoff:
I think the best way here would be to not allow activating an empty folder, but rather only allow activating a folder that contains a `Project.toml`. I'm pretty sure that we also don't allow activating empty folders via the "change active env" command, because essentially we aren't handling the case of an empty env well throughout the extension.

tapeinosyne:
>  allow activating a folder that contains a Project.toml

At a glance, I can't find anything in the extension API that would allow us to trivially check the contents of a folder upon activation of the context menu. Activating directly on a `Project.toml` file would be easy, however.

> we also don't allow activating empty folders via the "change active env" command

As far as I can tell, the extension doesn't prevent you from doing so. (I just tried.)

PR #1932, opened 2021-02-12
WIP: Add a plot panel for switching plots

TonyLianLong:
This PR adds a plot panel for switching plots. The thumbnail has not been implemented so now it shows up as text. However, the switching part works.

image

I don't think we need an end-user documentation, but after I finish everything I will add this to the CHANGELOG.


azev77:
Thanks!
Maybe this is for a different PR, but is it possible to add a “save plot as” button?

TonyLianLong:
> Thanks!
> Maybe this is for a different PR, but is it possible to add a “save plot as” button?

Thanks for the idea of adding a new functionality!

This PR aims to add a sidebar for previewing plots and I think your suggestion could be implemented after this PR. Submitting an issue may be a good way to note this down. You could mention me in the issue and I will pay attention to it while implementing this PR to avoid conflicts and probably implement this functionality after this PR.

src/interactive/plots.ts
```diff
@@ -26,6 +28,120 @@ export function activate(context: vscode.ExtensionContext) {
...
+                    console.error(`Unknown message type from WebView: ${msg.type}, value: ${msg.value}`)
+            }
+
+        })
+
+        this.reloadPlotPane();
+    }
+
+    getWebviewHTML(innerHTML: string) {
+        const extensionPath = g_context.extensionPath
```
pfitzseb, line 78: I'd prefer having `g_context` be a property of this class (set in the constructor) instead of referencing the global. That's basically the direction we wanna go everywhere in the extension.

package.json
```diff
@@ -849,6 +849,11 @@
                     "type": "webview",
                     "id": "julia-documentation",
                     "name": "Documentation"
+                },
+                {
+                    "type": "webview",
+                    "id": "julia-plot-navigator",
+                    "name": "Plot Navigator"
```
pfitzseb, line 856: It might make sense to add a `"when": "hasJuliaREPL"` here -- the navigator is basically useless when there's no running Julia session, right?

libs/plotter/plotter.css.map
```diff
@@ -0,0 +1,7 @@
+{
```
pfitzseb, line 1: Probably best to `.gitignore` this.

pfitzseb (commented):
Looks good so far (left a few comments inline). I'd like to get actual thumbnails in before merging this, but not entirely sure how to best go about it. Will probably need some experimenting with.

package.json
davidanthoff: Well, one can open a REPL, plot some things, kill the REPL and still navigate around in the plot panel, right? I.e. plots outlive the REPL.

pfitzseb: True that.

TonyLianLong:
I just added a functionality to generate thumbnails using html2canvas, which should support all types of HTML.

image

TonyLianLong:
As a side-effect of capturing the whole webpage, I get a lot of white space (which could be fixed by capturing specific elements only). However, the framework is done. I'll fix the issues mentioned above. One more thing: when I plot 2 figures in a consecutive way using a Julia function, the first plot is not rendered and the thumbnail not generated. I'll try my best to fix these problems.

src/interactive/plots.ts
TonyLianLong: Fixed.

libs/plotter/plotter.css.map
TonyLianLong: Added.

davidanthoff (changes requested):
Two pretty minor change requests from me.

If @pfitzseb could do the final sign-off on this PR it would be great, I think he has a better understanding at the moment how this is working :)

.gitignore
```diff
@@ -5,3 +5,7 @@
 /dist/
 /out/
 /scripts/languageserver/julia_pkgdir/lib/
+/scripts/languageserver/julia_pkgdir/compiled/
+/scripts/languageserver/julia_pkgdir/logs/
+/libs/plotter/plotter.css.map
+.sass-cache
```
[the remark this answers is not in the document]
TonyLianLong: I agree. I added them because it appears on the untracked stuff. However, they should be otherwhere instead of current PR.

libs/plotter/main_plot_webview.js
```diff
@@ -0,0 +1,25 @@
...
+}
+
+window.addEventListener('load', _ => {
+    html2canvas(document.getElementsByTagName("body")[0], {
+        onrendered: (canvas) => {
+            console.log("Thumbnail is taken")
+            postMessageToHost("thumbnail", canvas.toDataURL("png"))
+        }
+    })
+})
```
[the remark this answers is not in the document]
TonyLianLong: I agree. I'll fix this with other changes in file structure.

.gitignore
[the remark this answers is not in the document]
davidanthoff: So I have to admit I don't fully understand why we need any changes to this file? Ideally this PR would just leave that alone?

davidanthoff (changes requested):
One more thing on file locations: can we have `libs/html2canvas/html2canvas.min.js ` instead of putting it into the `libs/plotter` folder? The general idea is that we have one folder per upstream project we incorporate.

davidanthoff:
@TonyLianLong Could you resolve the merge conflict?

pfitzseb:
I was trying this again yesterday and for am not getting any previews anymore (or rather, only empty white panels). Would be good if you could
1. Update html2canvas to the latest version (it's at 0.4 atm).
2. Limit the maximum height of each preview.
3. Give the actual plot element an id and pass that to html2canvas, so hopefully only the right part is plotted.
4. Scale down the preview (with 2. implemented we know its maximum size) and pass that out of the plot pane.

It also seems like the first plotting command in a session is broken and doesn't display anything, although that might be unrelated.

TonyLianLong:
@pfitzseb Do you mind if I use this one: https://html2canvas.hertzen.com/dist/html2canvas.min.js (the 1.0.0-rc7)? The most updated version, excluding pre-release ones, is 0.4.1, which is what I use now.

pfitzseb:
Yes, that's what I meant (worded that badly, sorry).

TonyLianLong:
With reference to plotgallery (https://github.com/julia-vscode/plotgallery), the current plot is highlighted:
image

davidanthoff (approved):
Alright, I merged `master`, and I undid the edits to `.gitignore`.

I just tried it, and I don't see the visual indicator for the currently active plot in the plot navigator that is shown in the latest screenshot?

I think the question now is whether we merge this as is and then try to fix things like the timing issue if the plot pane gets closed too quickly for the thumbnail to be generated later, or do we try to fix all of this before we merge anything?

My vote is to merge now: I think this is a nice improvement over the status quo. But @pfitzseb should also weigh in, I think.

pfitzseb (approved):
Made a few changes here, most crucially:
- In most (all?) cases the aspect ratio of the preview should be correct now. Also shouldn't include any superfluous parts of the plot pane now anymore.
- The newest preview is now at the top (so the list is reversed compared to previously).

Issue #1940, opened 2021-02-15
Displaying `DataValue`s crashes Julia


lhupe:
When I run this snippet
```julia
using DataValues
DataValue{Any}("foo")
```
in a Julia REPL in VSCode, the session crashes with `signal (4): illegal instruction`. 
```
Unreachable reached at 0x7f20d354eb2d

signal (4): Illegal instruction
in expression starting at none:1
put! at ./channels.jl:311
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2231 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2398
macro expansion at /home/lhupe/.vscode-server/extensions/julialang.language-julia-1.1.17/scripts/packages/VSCodeServer/src/eval.jl:33 [inlined]
#56 at ./task.jl:356
unknown function (ip: 0x7f20d34f781c)
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2231 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2398
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1690 [inlined]
start_task at /buildworker/worker/package_linux64/build/src/task.c:707
unknown function (ip: (nil))
Allocations: 19841063 (Pool: 19834840; Big: 6223); GC: 21
```
This does not happen 
 - in a regular REPL
 - with concrete type parameters (e.g. `DataValue{String}("foo")` is fine)

I have tested this on several Linux machines, one running Julia 1.5.2 and VSCode 1.52.1, the other running Julia  1.5.3 and VS Code 1.53.2. Both use version v1.1.17 of the julia-vscode extension.

pfitzseb:
Technically an upstream issue (https://github.com/queryverse/DataValues.jl/issues/51#issuecomment-779342664), but can also be fixed via https://github.com/julia-vscode/julia-vscode/pull/1943.

PR #1943, opened 2021-02-15
wrap out-channel content

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1940.

Not sure whether we should merge this or not, since technically it's not a bug in our code. AFAICT it won't hurt either though.


davidanthoff (approved):
Lets merge. Even if I fix the upstream issue, folks might use an old version and similar things might also occur in other packages, so I'm always in favor of hardening our code against all sorts of corner situations.

PR #1959, opened 2021-02-18
use repl ast transforms on 1.5 and up

### LINKED ISSUE #1958: ast_transforms
The REPL in julia-vscode appears to not run `Base.active_repl_backend.ast_transforms` functions at the right time, kind of understandable since it's only available in 1.5+.

```
julia> show_ast(ex) = (@show ex.args[2].args[1:3]; ex)
show_ast (generic function with 1 method)

julia> pushfirst!(Base.active_repl_backend.ast_transforms, show_ast)
3-element Vector{Any}:
 show_ast (generic function with 1 method)
 revise_first (generic function with 1 method)
 softscope (generic function with 1 method)

julia> 1 + 2
(ex.args[2]).args[1:3] = Any[VSCodeServer.evalrepl, :Main, "1 + 2"]
3
```

From the above you can see that the functions in `ast_transforms` are getting passed the expression from https://github.com/julia-vscode/julia-vscode/blob/e96e9425dd8e9d120177d5e8635b5048c0d48b9b/scripts/packages/VSCodeServer/src/repl.jl#L72-L74 rather than the actual expression `:(1 + 2)`.

If it's feasible to fix this, and someone can point me in the general direction where things would need changing I'm happy to take a look at fixing this myself.

pfitzseb:
Fixes #1958.

### DIFF
--- a/scripts/packages/VSCodeServer/src/repl.jl
+++ b/scripts/packages/VSCodeServer/src/repl.jl
@@ -66,12 +66,21 @@ function hook_repl(repl)
         repl.interface = REPL.setup_interface(repl)
     end
     main_mode = get_main_mode(repl)
-
-    # TODO: set up REPL module ?
-    main_mode.on_done = REPL.respond(repl, main_mode; pass_empty=false) do line
-        quote
-            $(evalrepl)(Main, $line, $repl, $main_mode)
+    if VERSION > v"1.5-"
+        push!(Base.active_repl_backend.ast_transforms, ast -> transform_backend(ast, repl, main_mode))
+    else
+        # TODO: set up REPL module ?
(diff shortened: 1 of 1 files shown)


MichaelHatherly:
Thanks, appears to solve it on my end.

davidanthoff (approved):

pfitzseb:
Just gotta hope that people don't try to `push!` their own transforms and expect them to work properly :)

MichaelHatherly:
Yeah, I'm `pushfirst!`ing mine to not a problem, and Revise also does that. It's unfortunate that there isn't some form of priority for the transforms to alleviate that issue.

pfitzseb:
Same as the display system, actually (cf https://github.com/JuliaLang/julia/pull/30922).

Issue #1900, opened 2021-01-31
Bug in live testing


davidanthoff:
When I try the live testing feature on CSTParser, I get this:
```
Status `C:\Users\david\AppData\Local\Temp\jl_Bd50j1\Project.toml`      
  [00ebfdb7] CSTParser v3.1.1-DEV `C:\Users\david\.julia\dev\CSTParser`
  [0796e94c] Tokenize v0.5.8
  [8dfed614] Test

CSTParser: Error During Test at C:\Users\david\.julia\dev\CSTParser\test\runtests.jl:4
  Got exception outside of a @test
  
Test Summary: | Error  Total
CSTParser     |     1      1
ERROR: Some tests did not pass: 0 passed, 0 failed, 1 errored, 0 broken.
finish(::Test.DefaultTestSet) at C:\Users\david\AppData\Local\Programs\Julia 1.5.3\share\julia\stdlib\v1.5\Test\src\Test.jl:879
top-level scope at C:\Users\david\AppData\Local\Programs\Julia 1.5.3\share\julia\stdlib\v1.5\Test\src\Test.jl:1125
Revise evaluation error at C:\Users\david\AppData\Local\Programs\Julia 1.5.3\share\julia\stdlib\v1.5\Test\src\Test.jl:1125

Stacktrace:
 [1] methods_by_execution!(::Any, ::VSCodeLiveUnitTesting.Revise.CodeTrackingMethodInfo, ::Dict{Module,Array{Expr,1}}, ::Module, ::Expr; mode::Symbol, disablebp::Bool, always_rethrow::Bool, kwargs::Base.Iterators.Pairs{Symbol,Bool,Tuple{Symbol},NamedTuple{(:skip_include,),Tuple{Bool}}}) at c:\Users\david\source\julia-vscode\scripts\packages\Revise\src\lowered.jl:231
 [2] #eval_with_signatures#90 at c:\Users\david\source\julia-vscode\scripts\packages\Revise\src\packagedef.jl:463 [inlined]
[...]

PR #1964, opened 2021-02-19
Fix live testing

davidanthoff:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1900.

Still todo:
- [ ] somehow add the equivalent of the `touch` thing that @timholy mentioned over in the issue so that we run all `include`s in the first execution story.


timholy:
I'm in the middle of a class so can't check, but can't you call `entr` directly as a way of avoiding the `touch`? Meaning don't call `track`, just call `entr`.

davidanthoff:
Hm, so I removed the `track` stuff and am only using `entr` now, but it doesn't seem to pick up edits in files that are included directly in `runtests.jl`...

timholy:
Oh, I see, you were using `track` to identify dependencies? Of course that makes you vulnerable to the issue in https://github.com/julia-vscode/julia-vscode/issues/1900#issuecomment-778789515. How does `test_file` get determined, could you just list them all?

Just FYI if necessary I'm happy to make Revise changes to help support this if it's necessary, but I'd like to make sure we don't open vulnerabilities to weird corner cases (Revise has found it necessary to take robustness very seriously).

davidanthoff:
Yes, exactly, I essentially called `track` on `test/runtests.jl`. `test_file` is just going to be `test/runtests.jl`, or some other entry file that the user can configure.

I think ideally we would have the ability to call `entr` on `test/runtests.jl` (or some other entry file that the user might configure), maybe with some option so that it behaves a bit more like `track`? Or something like that... The goal really is simply to rerun `test/runtests.jl` everytime anything in the tests or in the package being tested is updated.

davidanthoff:
@timholy Do you have any other suggestion how I might get this to work?

timholy:
Sorry for the delay here. My other big fires seem to be extinguished. I guess one option would be to go back to calling track and pre-load the `task_local_storage` with the right path? `includet` does this kind of thing, see https://github.com/timholy/Revise.jl/blob/ab805afa1fe780f4ea867e67764f131ec5fe46dd/src/packagedef.jl#L979-L1001.

One potential reason for my slowness is a bit of nervousness. Revise is a lot of work to maintain---the challenge is exacerbated by the twin facts that Julia is so flexible and that lowering is not documented to any substantial degree. Nevertheless I think I've squashed "almost" all bugs by being really fastidious. Having been so careful myself, setting myself up for things like

> > how do you avoid endless cycles?

> I'd be perfectly fine to simply say: that doesn't work :)

makes me a bit uneasy because I'm not sure what the error is going to look like or who is going to get the blame for that. I think questions like that merit a bit of thought before we go live with this kind of feature.

ThanoS-max:
Can we add an online judge for vscode?

davidanthoff:
Future of testing in VS Code is https://github.com/julia-vscode/julia-vscode/pull/2350.

PR #1966, opened 2021-02-20
Auto update more Julia submodules

davidanthoff:
This should make it easier to handle more Julia packages within the Julia extension. Essentially our update deps script now also updates Julia packages that we don't ship on master, by always pulling the newest git tag version of the package.

- [X] Changelog mention. If this PR should be mentioned in the CHANGELOG, please edit the CHANGELOG.md file in this PR.


davidanthoff:
I'm just going to merge this because I'm the only person that uses this script in any case :) So this PR is just a FYI for the rest of the team.

PR #1968, opened 2021-02-22
more robust repl hook setup

pfitzseb:
Fixes a bug that got reported on Slack.

Turns out the repl backend isn't guaranteed to be set up in an `atreplinit` block, so let's do the same as Revise to hook it.


davidanthoff (approved):
Uh, that is ugly, but I guess we don't really have a choice...

pfitzseb:
We could actually be slightly smarter about this by having the backup `on_enter` hook check whether the backend is available and switching to the AST-transform implementation if so. But that's a much more complicated implementation that hopefully isn't necessary

Issue #1978, opened 2021-02-28
Execute code command fails after REPL restart


yha:
I find that restarting julia ("Julia: Stop REPL") pretty consistently makes execution commands (alt/shift+enter) fail.
I see one of three behaviors after a restart:
1. (most common) the first execute command does nothing, the developer console says:
```
[[object Object]]listen EADDRINUSE: address already in use \\.\pipe\vsc-julia-repl-8224
$onExtensionRuntimeError @ /C:/Users/sternlab/AppData/Local/Programs/Microsoft VS Code/resources/app/out/vs/workbench/workbench.desktop.main.js:1961
/C:/Users/sternlab/AppData/Local/Programs/Microsoft VS Code/resources/app/out/vs/workbench/workbench.desktop.main.js:1961 
```
```
Error: listen EADDRINUSE: address already in use \\.\pipe\vsc-julia-repl-8224
	at Server.setupListenHandle [as _listen2] (net.js:1296:21)
	at listenInCluster (net.js:1361:12)
	at Server.listen (net.js:1458:5)
	at c:\Users\sternlab\.vscode\extensions\julialang.language-julia-insider-1.1.31\dist\extension.js:1:142996
	at c:\Users\sternlab\.vscode\extensions\julialang.language-julia-insider-1.1.31\dist\extension.js:1:143008
	at Generator.next (<anonymous>)
	at c:\Users\sternlab\.vscode\extensions\julialang.language-julia-insider-1.1.31\dist\extension.js:1:141176
	at new Promise (<anonymous>)
	at r (c:\Users\sternlab\.vscode\extensions\julialang.language-julia-insider-1.1.31\dist\extension.js:1:140921)
	at S (c:\Users\sternlab\.vscode\extensions\julialang.language-julia-insider-1.1.31\dist\extension.js:1:141989)
	at c:\Users\sternlab\.vscode\extensions\julialang.language-julia-insider-1.1.31\dist\extension.js:1:148903
	at Generator.next (<anonymous>)
	at c:\Users\sternlab\.vscode\extensions\julialang.language-julia-insider-1.1.31\dist\extension.js:1:141176
	at new Promise (<anonymous>)
	at r (c:\Users\sternlab\.vscode\extensions\julialang.language-julia-insider-1.1.31\dist\extension.js:1:140921)
	at oe (c:\Users\sternlab\.vscode\extensions\julialang.language-julia-insider-1.1.31\dist\extension.js:1:148727)
[...]

PR #1980, opened 2021-03-01
unique pipe names

pfitzseb:
Might fix https://github.com/julia-vscode/julia-vscode/issues/1978. I don't think this will break anything, so might make sense to get in regardless.

### DIFF
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -3,6 +3,7 @@ import { Subject } from 'await-notify'
 import { assert } from 'console'
 import * as net from 'net'
 import * as path from 'path'
+import { uuid } from 'uuidv4'
 import * as vscode from 'vscode'
 import * as rpc from 'vscode-jsonrpc/node'
 import * as vslc from 'vscode-languageclient/node'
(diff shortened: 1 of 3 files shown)


davidanthoff:
We had some problems at some point with pipe names that were too long, I think that was on Mac. I'm not sure, but we might just be at he limit with the old names? I'm wondering whether we should just take out the `pid`, and just go with name and random id?

pfitzseb:
Hm, I kinda like the disambiguation by PID for debugging, but it's probably pretty unlikely to be necessary for multiple sessions.

Also, afaik Windows has the lowest maximum path length at 256 (although longer paths are possible nowadays).

davidanthoff:
I'm pretty positive the problem was on MacOS. I guess one way we could try this is to make sure to have things in a crash reporting block where this is used, so that we at least would be informed if this is a problem on some user systems?

pfitzseb:
Just pushed a commit that consistently uses `uuid()` instead of the PID everywhere (we were already doing that for the debugger, which seems to have worked just fine).

davidanthoff (approved):

PR #1987, opened 2021-03-03
Execute code block/cell support for .jmd

### LINKED ISSUE #1807: integrated support for Weave.jl is not working
For .jmd files `execute Code Block (Alt+Enter)` and `Execute Code Cell (Shift+Enter)` commands do not work in vs code (in Juno analogous commands work without any problem).

pgunnink:
I got annoyed with the lack of shift+enter/alt+enter support in .jmd files, so I created this. Right now, alt-enter in juliamarkdown mode just sends the entire current block to the REPL. Shift-enter (execute cell) does check if the cell is a Julia-block, starting with ` ```julia / ```{julia` and so does nothing in a markdown section. I could include the same check for the alt-enter command if needed.

Fixes #1807.

A more general note: do you have a direction to go with the .jmd files? Right now the experience is not really polished, and I could make some more fixes. But if you are going in a completely different direction than that is a bit useless (I read somewhere in an issue that you wanted to integrate .jmd and the Notebook interface in VS Code?)

### DIFF
--- a/package.json
+++ b/package.json
@@ -437,6 +437,11 @@
                 "when": "editorTextFocus && editorLangId == julia"
             },
             {
+                "command": "language-julia.executeCodeBlockOrSelectionAndMove",
+                "key": "alt+Enter",
+                "when": "editorTextFocus && editorLangId == juliamarkdown"
+            },
+            {
                 "command": "language-julia.interrupt",
                 "key": "Ctrl+C",
                 "when": "terminalFocus && isJuliaREPL && isJuliaEvaluating"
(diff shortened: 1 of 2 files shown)


pgunnink:
I just saw #1101, which is almost exactly the same PR for alt+enter support. Only difference I can spot right now is that my PR also allows for cell execution. So I'll leave it up to you what you want to do with this PR.

pfitzseb:
This doesn't quite seem to do the right thing for block execution, although I think we need to fix that in the LS. 

In general I kinda feel like it would be best to move the Weave support into a different extension or even get rid of it altogether though...

pgunnink:
I agree about the behavior, this was more intended as a quick fix for a problem I was dealing with. But if you first want to overhaul/remove Weave support that is also fine by me.

matrix-ranger (approved):

bgctw:
Can this be extended to work also with .md files?
This will be useful when writing package documentation with Documenter.jl using ` ```julia ` blocks.

PR #2010, opened 2021-03-09
fix inline eval ans handling

### LINKED ISSUE #2007: alt+enter errors while shift+enter doesn't
I guess this is due to some evaluation mechanism inside the julia-vscode extension, when you use `shift` + `enter` to evaluate the following code line by line

```julia
ci, typ = code_typed(cos, (Float64, ))[1]
ci.code[50]
```

it should give you the result of calling `getindex` on `ci.code`, however, if you use `alt` + `enter`, you will get the following when evaluating `ci.code[50]`

```julia
ERROR: syntax: SSAValue objects should not occur in an AST
Stacktrace:
 [1] top-level scope
   @ none:1
```

pfitzseb:
for SSAValues etc.

Fixes #2007.


pfitzseb:
Upstream fix is at https://github.com/JuliaLang/julia/pull/39965, but we'll still need this workaround for older versions.

davidanthoff (approved):

PR #2029, opened 2021-03-12
Ensure we perform updates to Document sequentially

DonJayamanne:
@davidanthoff 
Not sure I need an issue when submitting changes to the feature branch.
Same with updates to the change log.

Let me know if these are required.

### DIFF
--- a/src/notebook/helpers.ts
+++ b/src/notebook/helpers.ts
@@ -1,0 +1,42 @@
+import { NotebookDocument, workspace, WorkspaceEdit } from 'vscode'
+
+/**
+ * Use this to perform updates on all cells.
+ * We cannot update cells in parallel, this could result in data loss.
+ * E.g. assume we update execution order, while that's going on, assume we update the output (as output comes back from jupyter).
+ * At this point, VSC is still updating the execution order & we then update the output.
+ * Depending on the sequence its possible for some of the updates to get lost.
+ *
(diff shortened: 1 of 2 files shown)


src/notebook/helpers.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Depending on the sequence its possible for some of the updates to get lost.
+ *
+ * Excellent example:
+ * Assume we perform the following updates without awaiting on the promise.
+ * Without awaiting, its very easy to replicate issues where the output is never displayed.
+ * - We update execution count
+ * - We update output
+ * - We update status after completion
+ */
+const pendingCellUpdates = new WeakMap<NotebookDocument, Promise<unknown>>()
```
DonJayamanne, line 17: Hopefully by next week all of these work arounds will be a thing of the past, VSCode is updating their API (changes to how execution is handled and output is updated).

DonJayamanne: This is one of the reasons why outputs might not appear at times,

DonJayamanne: julia

davidanthoff:
> Not sure I need an issue when submitting changes to the feature branch.

No need for issues, we can just move quickly on all of this without much process as all, especially given that we do all work on a branch.

> Same with updates to the change log.

Also: no need for that.

PR #2032, opened 2021-03-12
Fix clearing output

DonJayamanne:
I think i added a bug, during a merge issue.
The output is getting cleared
still trying to test this, things are a little slow to start (starting julia everytime is a little slow)


davidanthoff:
Yeah, Julia startup is still not great...

PR #2036, opened 2021-03-12
Add ability for users to create a Julia Notebook (defaults to Julia langauge)

DonJayamanne:
This is a  feature (extensibility point we added for .NET Interacgive notebooks extension).

See below git (it doesn't make Jupyter a dependency, merely adds a command if the user has it installed).
Not sure you want this, but i figured it could be handy.
julia

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -14,6 +14,7 @@ import { ProfilerResultsProvider } from './interactive/profiler'
 import * as repl from './interactive/repl'
 import * as jlpkgenv from './jlpkgenv'
 import * as juliaexepath from './juliaexepath'
+import { registerWithJupyter } from './notebook/jupyterExtension'
 import { JuliaNotebookFeature } from './notebook/notebookFeature'
 import * as openpackagedirectory from './openpackagedirectory'
 import { JuliaPackageDevFeature } from './packagedevtools'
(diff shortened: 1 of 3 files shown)


src/extension.ts
```diff
@@ -1,302 +1,304 @@
...
+        vscode.languages.setLanguageConfiguration('julia', {
+            indentationRules: {
+                increaseIndentPattern: /^(\s*|.*=\s*|.*@\w*\s*)[\w\s]*(?:["'`][^"'`]*["'`])*[\w\s]*\b(if|while|for|function|macro|(mutable\s+)?struct|abstract\s+type|primitive\s+type|let|quote|try|begin|.*\)\s*do|else|elseif|catch|finally)\b(?!(?:.*\bend\b[^\]]*)|(?:[^\[]*\].*)$).*$/,
+                decreaseIndentPattern: /^\s*(end|else|elseif|catch|finally)\b.*$/
+            }
+        })
+
+        // Active features from other files
+        juliaexepath.activate(context)
+        registerWithJupyter()
```
DonJayamanne, line 63: No need to await on the promise

davidanthoff:
Just a warning: I had for a brief while broken the `notebook` branch. I just pushed a new commit that fixes things. But if you were unlucky and branched off that bad commit from `notebook`, then the kernel would not start. Sorry!

davidanthoff:
Not sure what happened to the diff there? Ideally we wouldn't have a commit in the history that changes every line in the extension.ts file, because that would merging things later on very difficult, I think.

DonJayamanne:
> ot sure what happened to the diff there? Ideally we wouldn't have a commit in the history that changes every line in the extension.ts file, because that would merging things later on very difficult, I think.

I agree, i couldn't figure it out either.
I'll try to fix it.

DonJayamanne:
No joy, reverted it and tried still showing same changes, let me try another branch.

Issue #2045, opened 2021-03-13
Support multiple Julia version kernels


davidanthoff:
For the Jupyter kernel story, we need to support multiple Julia version kernels.

At the moment we just return one kernel [here](https://github.com/julia-vscode/julia-vscode/blob/334a1334189f1dfe28d77c14b879b4ac59cc0e5b/src/notebook/notebookFeature.ts#L12), but really we should return a list of kernels there where each kernel corresponds to a different Julia version that is installed.

I think essentially we need to change things as follows:
- Detect _all_ Julia versions that are currently installed, not just the newest one. Potentially we might need something like an array for manually configured Julia paths, but I think for beginners we could just go with the auto-detected versions.
- In general then return a list of kernels we can offer, one for each Julia version we have.
- For existing notebooks that are opened, where the notebook was created with a Julia version that is not installed, we then need some smart fallback options that we probably still have to figure out...

davidanthoff:
CC @DonJayamanne Should I tag you on all issues that are related to notebooks that I create you? That way you would get notified without having to subscribe to all notifications from this repo (where there are probably _a lot_ that are of zero interest to you). Or do you just want to look up issues with the `area-notebook` label yourself?

DonJayamanne:
> do you just want to look up issues with the area-notebook label yours

Latter is fine. I'll be checking the issues.

davidanthoff:
This is implemented on our end now.

PR #2053, opened 2021-03-13
Search all julia versions and display in the kernel picker

DonJayamanne:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2045


src/notebook/notebookKernel.ts
```diff
@@ -167,11 +175,11 @@ export class JuliaKernel implements vscode.NotebookKernel {
 
             await serverListeningPromise.wait()
 
-            const jlexepath = await getJuliaExePath()
+            const jlexepath = this.executablePath
             const pkgenvpath = await getAbsEnvPath()
 
             this._terminal = vscode.window.createTerminal({
-                name: 'Julia Notebook Kernel',
+                name: `Julia Notebook Kernel ${path.basename(this.document.uri.fsPath)}`,
```
DonJayamanne, line 182: This way if you have multiple notebooks you can tell which one is which.
Btw, what's the purpose of displaying this to the user, is it for diagnostic purposes (if things go wrong, they can see the output in the terminal)

src/notebook/notebookKernel.ts
```diff
@@ -78,6 +85,7 @@ export class JuliaKernel implements vscode.NotebookKernel {
 
     public async start() {
         if (!this._terminal) {
+            this._current_request_id = 0
```
DonJayamanne, line 88: When restarting, or the like, this needs to be reset (at least in Jupyter, thats the behavior)

src/notebook/notebookKernel.ts
```diff
@@ -21,6 +21,9 @@ const notifyTypeRunCell = new NotificationType<{ current_request_id: number, cod
 const notifyTypeRunCellSucceeded = new NotificationType<{ request_id: number }>('runcellsucceeded')
 const notifyTypeRunCellFailed = new NotificationType<{ request_id: number, output: {ename: string, evalue: string, traceback: string}}>('runcellfailed')
 
+function getDisplayPathName(pathValue: string): string {
+    return pathValue.startsWith(homedir()) ? `~${path.relative(homedir(), pathValue)}` : pathValue
```
DonJayamanne, line 25: Display relative paths to home directory (else the paths are just unnecessarily long). & users in general know that `~/sdfwef` means its relative to home

DonJayamanne:
@davidanthoff I'll try to work on the tests next.

DonJayamanne:
@davidanthoff pinging as a reminder

PR #2073, opened 2021-03-23
Updates based on changes to execution API

DonJayamanne:
* Updates based on changes in Proposed API related to execution

@davidanthoff  Please do not merge this, we'll need to wait til tomorrow for the new VS Code Insiders version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "preview": false,
     "publisher": "julialang",
     "engines": {
-        "vscode": "^1.54.0"
+        "vscode": "1.55.0-insider"
     },
     "enableProposedApi": true,
     "license": "SEE LICENSE IN LICENSE",
--- a/src/notebook/notebookKernel.ts
+++ b/src/notebook/notebookKernel.ts
@@ -8,11 +8,12 @@ import { getAbsEnvPath } from '../jlpkgenv'
 import { getJuliaExePath } from '../juliaexepath'
 import { getCrashReportingPipename } from '../telemetry'
 import { generatePipeName } from '../utils'
-import { chainWithPendingUpdates } from './helpers'
 
 interface ExecutionRequest {
     cell: vscode.NotebookCell
-    executionOrder: number
+	executionOrder: number
+	task: vscode.NotebookCellExecutionTask
+	startTime: number;
 }
 
 const notifyTypeDisplay = new NotificationType<{ mimetype: string, current_request_id: number, data: string }>('display')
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
     "preview": false,
     "publisher": "julialang",
     "engines": {
-        "vscode": "^1.54.0"
+        "vscode": "1.55.0-insider"
```
DonJayamanne, line 9: We cannot use stable Version as the API is too different.

src/notebook/notebookKernel.ts
```diff
@@ -30,182 +30,144 @@ export class JuliaKernel implements vscode.NotebookKernel {
     private _msgConnection: MessageConnection;
     private _current_request_id: number = 0;
 
+    public id = 'JuliaKernel'
```
DonJayamanne, line 33: This is no longer optional.

src/notebook/notebookKernel.ts
```diff
@@ -30,182 +28,144 @@ export class JuliaKernel implements vscode.NotebookKernel {
...
+	                if (executionRequest) {
+	                    executionRequest.task.replaceOutput([new vscode.NotebookCellOutput([new vscode.NotebookCellOutputItem(mimetype, data)])])
+	                }
+	            })
+
+	            this._msgConnection.onNotification(notifyTypeStreamoutput, ({ name, current_request_id, data }) => {
+	                if (name === 'stdout') {
+	                    const executionRequest = this.executionRequests.get(current_request_id)
+	                    if (executionRequest) {
+	                        executionRequest.task.replaceOutput([new vscode.NotebookCellOutput([new vscode.NotebookCellOutputItem('application/x.notebook.stdout', data)])])
```
DonJayamanne: I can see one problem with the append approach.
Assume you have a progress bar, the kernel will send the output for each tick,
however we could end up with 100s if not 1000s or more output objects for a single line progress bar.

In Jupyter extension we do not append stdout, as it could be a progress indicator.
We compress the ansi output by handling special characters used by terminals to generate progress bars.

I'm assuming we'd want the same fix here.
FYI - If we don't do that, then users could end up with a very large ipynb file with lots of unnecessary output.

src/notebook/notebookKernel.ts
```diff
@@ -30,182 +28,144 @@ export class JuliaKernel implements vscode.NotebookKernel {
...
+	                    }
+	                }
+	                else if (name === 'stderr') {
+	                    const executionRequest = this.executionRequests.get(current_request_id)
```
DonJayamanne, line 109: Same as earlier, depends on what stream packages write to when generating progress bars.
I've found plenty of pacakges using stdout & stderr in python.

DonJayamanne:
@davidanthoff pinging as a reminder

davidanthoff:
@DonJayamanne presumably this can be closed because your recent PR replaces it, right?

PR #2076, opened 2021-03-25
Add support for Julia 1.6.1

davidanthoff:
Just want to be ready for it :)

- [X] Changelog mention. If this PR should be mentioned in the CHANGELOG, please edit the CHANGELOG.md file in this PR.

### DIFF
--- a/src/juliaexepath.ts
+++ b/src/juliaexepath.ts
@@ -36,6 +36,7 @@ export async function getJuliaExePath() {
             let pathsToSearch = []
             if (process.platform === 'win32') {
                 pathsToSearch = ['julia.exe',
+                    path.join(homedir, 'AppData', 'Local', 'Programs', 'Julia 1.6.1', 'bin', 'julia.exe'),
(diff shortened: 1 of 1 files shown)


pfitzseb (approved):

jwortmann:
I think the paths for 1.6.0 and 1.6.1 don't work. I just installed 1.6.0 on Windows and the correct directory is `Julia-1.6.0` and not `Julia 1.6.0`. See https://github.com/JuliaLang/julia/pull/36693.

So the VS Code extension doesn't work if Julia is not in the PATH and if there's no previous Julia version installed to fall back on. I would suggest to fix the paths for 1.6.0 and 1.6.1 and make a new release.

davidanthoff:
Oh, thanks for catching that!! I just fixed it and am pushing out a new build now.

And also: arghhh, why are they changing the install path so often?!? This is now the fourth distinct pattern they have been using... I don't understand why on earth this had to be changed _at all_ from the format they had up until 1.3... It was just fine and everything worked. Frustrating...

Issue #1924, opened 2021-02-10
Backtrace formatting on 1.6


marius311:
I'm sure this would eventually happen anyway but didn't see an issue so figured I'd open one. The backtrace formatting isn't set up to handle the new (awesome) style of backtraces on 1.6, it looks like:

![image](https://user-images.githubusercontent.com/1510968/107484052-bdb39100-6b36-11eb-86ca-2e29f433e3c3.png)

instead of the working / hyperlinked formatting on 1.5.

PR #2091, opened 2021-03-31
fix stacktrace rendering on 1.6

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/1924. Tested this on 1.0, 1.5, and 1.6 -- seems to work decently well on all of them.

### DIFF
--- a/scripts/packages/VSCodeServer/src/eval.jl
+++ b/scripts/packages/VSCodeServer/src/eval.jl
@@ -279,16 +279,19 @@ end
 const LOCATION_REGEX = r"\[\d+\]\s(?<body>.+)\sat\s(?<path>.+)\:(?<line>\d+)"
 
 function backtrace_string(bt)
-    s = sprintlimited(bt, func = Base.show_backtrace, limit = MAX_RESULT_LENGTH)
-    lines = strip.(split(s, '\n'))
+    io = IOBuffer()
 
-    return join(map(enumerate(lines)) do (i, line)
-        i === 1 && return line # "Stacktrace:"
-        m = match(LOCATION_REGEX, line)
-        m === nothing && return line
-        linktext = string(m[:path], ':', m[:line])
-        linkbody = vscode_cmd_uri("language-julia.openFile"; path = fullpath(m[:path]), line = m[:line])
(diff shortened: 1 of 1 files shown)


scripts/packages/VSCodeServer/src/eval.jl
```diff
@@ -279,16 +279,19 @@ end
 const LOCATION_REGEX = r"\[\d+\]\s(?<body>.+)\sat\s(?<path>.+)\:(?<line>\d+)"
 
 function backtrace_string(bt)
-    s = sprintlimited(bt, func = Base.show_backtrace, limit = MAX_RESULT_LENGTH)
-    lines = strip.(split(s, '\n'))
-
-    return join(map(enumerate(lines)) do (i, line)
-        i === 1 && return line # "Stacktrace:"
-        m = match(LOCATION_REGEX, line)
```
aviatesk, line 287: I think we can remove [the definition of `LOCATION_REGEX`](https://github.com/julia-vscode/julia-vscode/blob/a84b6def506dfad9ce81e2eefaf637597c384a23/scripts/packages/VSCodeServer/src/eval.jl#L278-L279).

scripts/packages/VSCodeServer/src/eval.jl
```diff
@@ -279,16 +279,19 @@ end
...
+    io = IOBuffer()
+
+    println(io, "Stacktrace:")
+    for (i, frame) in enumerate(stacktrace(bt))
+        file = string(frame.file)
+        full_file = fullpath(something(Base.find_source_file(file), file))
+        cmd = vscode_cmd_uri("language-julia.openFile"; path = full_file, line = frame.line)
+
+        print(io, i, ". `")
+        Base.StackTraces.show_spec_linfo(io, frame)
```
aviatesk, line 291: I wonder if this code is compatible with pre-1.6 versions ? Not sure we want to support those versions, though..

scripts/packages/VSCodeServer/src/eval.jl
pfitzseb: Done.

aviatesk (approved):

PR #2098, opened 2021-04-08
Fix the bug that semicolon suppression does not work with inline evaluation 

### LINKED ISSUE #2093: Inline evaluation of a line ending with a semicolon displays output
Inline evaluation of a line ending with a semicolon displays output both inline and on the terminal.
I would expect both outputs to be suppressed by the semicolon.

Julia extension 1.1.38; MacOS; Julia 1.6

Olament:
Fixes #2093.

### DIFF
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -503,7 +503,7 @@ async function evaluateBlockOrSelection(shouldMove: boolean = false) {
 
         if (selection.isEmpty) {
             const currentBlock = await getBlockRange(getVersionedParamsAtPosition(editor.document, startpos))
-            range = new vscode.Range(currentBlock[0].line, currentBlock[0].character, currentBlock[1].line, currentBlock[1].character)
+            range = new vscode.Range(currentBlock[0].line, currentBlock[0].character, currentBlock[1].line, currentBlock[1].character + 1)
             nextBlock = editor.document.validatePosition(new vscode.Position(currentBlock[2].line, currentBlock[2].character))
         } else {
             range = new vscode.Range(selection.start, selection.end)
@@ -539,9 +539,10 @@ async function evaluate(editor: vscode.TextEditor, range: vscode.Range, text: st
     const section = vscode.workspace.getConfiguration('julia')
     const resultType: string = section.get('execution.resultType')
     const codeInREPL: boolean = section.get('execution.codeInREPL')
+    const isSuppressed: boolean = text.endsWith(';')
 
     let r: results.Result = null
-    if (resultType !== 'REPL') {
+    if (resultType !== 'REPL' && !isSuppressed) {
         r = results.addResult(editor, range, ' ⟳ ', '')
     }
 
@@ -554,12 +555,12 @@ async function evaluate(editor: vscode.TextEditor, range: vscode.Range, text: st
             code: text,
             mod: module,
             showCodeInREPL: codeInREPL,
-            showResultInREPL: resultType !== 'inline',
+            showResultInREPL: resultType !== 'inline' && !isSuppressed,
             softscope: true
         }
     )
 
-    if (resultType !== 'REPL') {
+    if (resultType !== 'REPL' && !isSuppressed) {
         if (result.stackframe) {
             results.clearStackTrace()
             results.setStackTrace(r, result.all, result.stackframe)


pfitzseb:
I'm pretty sure this needs to happen in the [Julia code](https://github.com/julia-vscode/julia-vscode/blob/master/scripts/packages/VSCodeServer/src/eval.jl#L94) so we stop e.g. plots from being displayed as well. Do you want to take a shot at that?

Olament:
You're right! Thanks for point that out. I will give it a try tomorrow.

Olament:
@pfitzseb I just pushed a commit that address the issue (plot being displayed when suppressed). Let me explain my code here.  

#### Suppressing REPL output
[repl_runcode_request](https://github.com/julia-vscode/julia-vscode/blob/3d7b5d79509fdcda354ea9853d8955dc262f7f40/scripts/packages/VSCodeServer/src/eval.jl#L94) actually implemented it correctly. It only display the result if the line does not ends with semicolon (code [here](https://github.com/julia-vscode/julia-vscode/blob/3d7b5d79509fdcda354ea9853d8955dc262f7f40/scripts/packages/VSCodeServer/src/eval.jl#L162)). The reason that we still get REPL output when a line ends with semicolon is because the julia-vscode strips the semicolon before passing it to the Julia code. julia-vscode determine codes that need to be evaluated when pressing ```Alt + Enter``` by calling [getBlockRange](https://github.com/julia-vscode/julia-vscode/blob/c747d812df0026e3e3ede1c802810d02aaad48cc/src/interactive/repl.ts#L361) which then call LSP. However, the range that return by julia LSP will not include the trailing semicolon. For following code block,

```julia
string(1+3);
```

the LSP will return a range that include ```string(1+3)``` instead of ```string(1+3);```. I am not sure if this is a bug or intended design. But a temporary workaround is to extend the range by 1 such that we always include semicolon (if there is any). Since we validate the range before pass it to following functions it is always safe to do so. 


#### Suppressing  Inline evaluation and plot pop-up

Since displaying the inline evaluation is happened at the TypeScript level, I suppressed any inline evaluation UI if the code ends with semicolon. Also, the original code of [repl_runcode_request](https://github.com/julia-vscode/julia-vscode/blob/3d7b5d79509fdcda354ea9853d8955dc262f7f40/scripts/packages/VSCodeServer/src/eval.jl#L94) does not suppress plot pop-up when ```ResultType === 'inline'```. So I added few line of codes addressing that as well.

pfitzseb:
Something like
```diff
diff --git a/scripts/packages/VSCodeServer/src/eval.jl b/scripts/packages/VSCodeServer/src/eval.jl
index 3546f4d..3e5b054 100644
--- a/scripts/packages/VSCodeServer/src/eval.jl
+++ b/scripts/packages/VSCodeServer/src/eval.jl
@@ -166,7 +166,7 @@ function repl_runcode_request(conn, params::ReplRunCodeRequestParams)
                             Base.display_error(stderr, err, catch_backtrace())
                         end
                     end
-                else
+                elseif res !== nothing && !ends_with_semicolon(source_code)
                     try
                         Base.invokelatest(display, InlineDisplay(), res)
                     catch err
@@ -177,6 +177,10 @@ function repl_runcode_request(conn, params::ReplRunCodeRequestParams)
                     end
                 end
 
+                if !(res isa EvalError) && ends_with_semicolon(source_code)
+                    res = nothing
+                end
+
                 rendered_result = safe_render(res)
             end
         end
diff --git a/src/interactive/repl.ts b/src/interactive/repl.ts
index 00b8fd5..84ded04 100644
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -497,19 +497,24 @@ async function evaluateBlockOrSelection(shouldMove: boolean = false) {
 
     for (const selection of selections) {
         let range: vscode.Range = null
+        let fullrange: vscode.Range = null
         let nextBlock: vscode.Position = null
         const startpos: vscode.Position = editor.document.validatePosition(new vscode.Position(selection.start.line, selection.start.character))
         const module: string = await modules.getModuleForEditor(editor.document, startpos)
 
         if (selection.isEmpty) {
             const currentBlock = await getBlockRange(getVersionedParamsAtPosition(editor.document, startpos))
-            range = new vscode.Range(currentBlock[0].line, currentBlock[0].character, currentBlock[1].line, currentBlock[1].character)
+            const blockStartPos = editor.document.validatePosition(new vscode.Position(currentBlock[0].line, currentBlock[0].character))
+            const blockEndPos = editor.document.validatePosition(new vscode.Position(currentBlock[1].line, currentBlock[1].character))
+            const lineEndPos = editor.document.validatePosition(new vscode.Position(currentBlock[1].line, Infinity))
+            range = new vscode.Range(blockStartPos, blockEndPos)
+            fullrange = new vscode.Range(blockStartPos, lineEndPos)
             nextBlock = editor.document.validatePosition(new vscode.Position(currentBlock[2].line, currentBlock[2].character))
         } else {
             range = new vscode.Range(selection.start, selection.end)
         }
 
-        const text = editor.document.getText(range)
+        const text = editor.document.getText(fullrange)
 
         if (shouldMove && nextBlock && selection.isEmpty && editor.selections.length === 1 && editor.selection === selection) {
             validateMoveAndReveal(editor, nextBlock, nextBlock)
```
is slightly nicer imho, because the changes are much better localiczed. The new definition of `fullrange` makes sure that the whole last line is passed to Julia, including trailing whitespace/semi-colons etc, which should be more robust than just checking if the next char is a semi-colon.

Olament:
This looks much better indeed. However, when using ```resultType === 'inline'```, all of the inline evaluation result shows ```⟳``` (except code block and line with only side effects). Under the debug mode, it seems like the evaluation is stuck at [requestTypeReplRunCode](https://github.com/julia-vscode/julia-vscode/blob/c747d812df0026e3e3ede1c802810d02aaad48cc/src/interactive/repl.ts#L548) indefinitely. Any ideas what happened?


This is the testing code I used.
```julia
a = 4  #⟳
b = 10; #⟳
print(b) #✓

for j = 1:5
    string(2 * 3)
    print(10)
end #✓

## code cell
string(1+3); #⟳
string(1 * 5) #⟳
##

using Plots #✓
x = 1:10; y = rand(10); #⟳
plot(x, y) #⟳
plot(x, y); #⟳
```

PR #2110, opened 2021-04-18
Update README.md

claforte:
"If you face any issue..." points to the **python** vscode issues... is that on purpose? ;-)


pfitzseb (approved):

pfitzseb:
It most definitely isn't. Good catch and thanks for the PR.

davidanthoff:
Haha, I have to say, though, that was a pretty elegant way to make us think we don't have bugs ;)

Issue #2173, opened 2021-05-27
Tmux pane killed


ericphanson:
I started a Julia repl with the experimental tmux feature, then left it there (not running anything) and the next day I reconnected to my ec2 instance and started the REPL, and I did get a tmux session that I could scroll up in etc, but with this message:
```
(env) pkg> ┌ Warning: Some Julia code in the VS Code extension crashed with
│   e = InvalidStateException("Channel is closed.", :closed)
└ @ VSCodeServer ~/.vscode-server/extensions/julialang.language-julia-insider-1.2.0/scripts/error_handler.jl:5
Pane is dead (status 1, Wed May 26 19:50:38 2021)
[dir]0:[tmux][dead]*  
```

pfitzseb:
Turns out this regressed before release. The issue is that the Julia session dies if the remote (aka client) is closed. Will fix.

PR #2177, opened 2021-05-27
fix tmux issues

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2173.

I *think* our connection handling is fairly correct with this.


davidanthoff (approved):

davidanthoff:
I merged before tests pass because all those tests will run again for the tag, that way we can save some time.

PR #2217, opened 2021-06-14
WIP - Support Notebooks using Stable API

DonJayamanne:
Adds support for Notebooks using the Stable VSC API:
* Remove proposed API
* Users no longer need Insiders (this will work in VS Code Stable 🥳 🍾 )
* User will require a minimum of VS Code 1.57 (to use Notebooks in Stable)
* Removed *.jlnb (this was not implemented & I'm assuming support for ipynb from Jupyter extension is sufficient)
	* If not, please provide details of *.jlnb & I can create a notebook (integration) in VS Code for that file format
	* Note: Support for *.ipynb currently requires users to install Jupyter extension
	* If this is not desired, we can discuss this on Jupyter repo
	* If we'd prefer to use *.jlnb, please provide details of this file format (is this JSON, etc)
* Renderers:
	* If we're relying on user to have Jupyter extension installed to open *.ipynb, then rendering of output is also handled by that extnsion
	* We can optionally use the code in this extension to render the output as well (I'll need to re-create all of the renderers we have in this extnsion as a renderer)

**Suggestions**
* Stage 1
	* To get this PR working/approved, we can assume we rely on Jupyter extension for opening *.ipynbn & rendering of the output
	* Note: Julia extension is still responsible for execution
	* We can update the Jupyter extension to recommend users to install Julia extension for running notebooks when opening existing Julia notebooks (we've done this or others)
* Stage 2
	* Consider adding support for *.jlnb (this was not implemented in the original notebook branch, hence I removed it, so we can have a working `Stage 1`)
	* Consider creating renderers for the output (this is assuming we use *.jlnb file format)
* Stage 3
	* Support for Interrupts (currently not supported in the extension)

**Pending**
* Ensure outputs have the right bytes (convert JSON to bytes, image to Bytes, etc)


davidanthoff:
Fantastic! I have some time over the next weeks to help get this over the finishing line :)

I'm just going to merge this PR as is, so that we can continue finishing everything up on the notebook branch, ok?

davidanthoff:
Oh, and yes: lets forget about the entire `jlnb` file format, that was really a relict from before it was clear that the Jupyter extension would be what it now is.

Issue #2244, opened 2021-06-25
Handle Julia environments in notebooks


davidanthoff:
Main question is: do we follow the IJulia semantics to run in the env of the parent folder of the notebook, or do we run in the currently active environment?

PR #2252, opened 2021-06-29
Detect notebook Julia environments

davidanthoff:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2244.

This now uses a Julia env if it is in the folder of the notebook, or a parent if that is still in the workspace.

### DIFF
--- a/src/notebook/notebookKernel.ts
+++ b/src/notebook/notebookKernel.ts
@@ -1,3 +1,4 @@
+import { exists } from 'async-file'
 import { Subject } from 'await-notify'
 import { ChildProcess, spawn } from 'child_process'
 import * as net from 'net'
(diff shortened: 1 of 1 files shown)


src/notebook/notebookKernel.ts
```diff
@@ -112,6 +113,40 @@ export class JuliaKernel {
         }
     }
 
+    private async containsJuliaEnv(folder: string) {
+        return (await exists(path.join(folder, 'Project.toml')) && await exists(path.join(folder, 'Manifest.toml'))) ||
+            (await exists(path.join(folder, 'JuliaProject.toml')) && await exists(path.join(folder, 'JuliaManifest.toml')))
```
pfitzseb, line 118: Probably a pointless microoptimization, but `readdir` seems like the better choice.

pfitzseb (approved):

src/notebook/notebookKernel.ts
davidanthoff: I'm not even sure? I guess kind of depends how `exists` is implemented, right? But we'd have to handle casing manually, folders with a very large number of files etc. Maybe also all easy, but for now this works :)

PR #2253, opened 2021-06-29
Use IJuliaCore.jl

davidanthoff:
I moved all the code that I had originally just blindly copied from IJulia.jl into IJuliaCore.jl. The plan is that IJulia.jl will then also eventually take a dependency on IJuliaCore.jl, so that we can actually share that code and don't have it diverge.


davidanthoff:
I'm just going to merge this without a review for now. @pfitzseb if you would look over it after the fact it would still be great, and then we can fix any problems you might see in follow up PRs. Just want to get an insider build out right now.

PR #2257, opened 2021-06-29
Reuse kernel even if notebook was closed

### LINKED ISSUE #2238: Reuse running notebook kernel if notebook is closed
Say a user opens a notebook, runs some code, closes the notebook. The kernel will still be running. Now the user re-opens the notebook, and we should make sure that we re-use the already running kernel, like JupyterLab does.

davidanthoff:
Fixes #2238.

@DonJayamanne, does this look ok to you? It feels a bit ad-hoc, but I guess this is how we would do this kind of thing?

### DIFF
--- a/src/notebook/notebookFeature.ts
+++ b/src/notebook/notebookFeature.ts
@@ -57,15 +57,31 @@ export class JuliaNotebookFeature {
     private async executeCells(cells: vscode.NotebookCell[], notebook: vscode.NotebookDocument, controller: vscode.NotebookController): Promise<void> {
         // First check whether we already have a kernel running for the current notebook document
         if (!this.kernels.has(notebook)) {
-            const kernel = new JuliaKernel(this.context.extensionPath, controller, notebook, this._juliaVersions.get(controller.id), this._outputChannel, this)
-            await this.workspaceFeature.addNotebookKernel(kernel)
-            this.kernels.set(notebook, kernel)
 
-            kernel.onStopped(e => {
-                if (this.kernels.get(notebook) === kernel) {
-                    this.kernels.delete(notebook)
+            // Check whether there is still a running kernel for a closed notebook with the same Uri
+            let foundExistingKernel = false
+            for (const [k, v] of this.kernels) {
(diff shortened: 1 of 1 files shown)


src/notebook/notebookFeature.ts
```diff
@@ -57,15 +57,31 @@ export class JuliaNotebookFeature {
...
-                if (this.kernels.get(notebook) === kernel) {
-                    this.kernels.delete(notebook)
+            // Check whether there is still a running kernel for a closed notebook with the same Uri
+            let foundExistingKernel = false
+            for (const [k, v] of this.kernels) {
+                if (k.isClosed && (k.uri.toString() === notebook.uri.toString())) {
+                    foundExistingKernel = true
+                    v.notebook = notebook
+
+                    this.kernels.delete(k)
```
DonJayamanne, line 68: Does this do anything, the key in the map is `notebook`, hence this line is a noop.

src/notebook/notebookFeature.ts
```diff
@@ -57,15 +57,31 @@ export class JuliaNotebookFeature {
...
-            await this.workspaceFeature.addNotebookKernel(kernel)
-            this.kernels.set(notebook, kernel)
 
-            kernel.onStopped(e => {
-                if (this.kernels.get(notebook) === kernel) {
-                    this.kernels.delete(notebook)
+            // Check whether there is still a running kernel for a closed notebook with the same Uri
+            let foundExistingKernel = false
+            for (const [k, v] of this.kernels) {
+                if (k.isClosed && (k.uri.toString() === notebook.uri.toString())) {
```
DonJayamanne, line 64: Shouldn't this be `if (!k.isClosed`

src/notebook/notebookFeature.ts
DonJayamanne: Rest seem ok.

DonJayamanne (approved):

src/notebook/notebookFeature.ts
davidanthoff: The naming here is confusing, the `k` is the key, so it is the previous (closed) notebook, so this line removes that.

PR #2260, opened 2021-06-30
Rework notebook display

davidanthoff:
Just avoid some code duplication and make it easier to add more MIME types. Also use the limit function to generate messages.

### DIFF
--- a/src/notebook/notebookKernel.ts
+++ b/src/notebook/notebookKernel.ts
@@ -125,7 +125,7 @@ export class JuliaKernel {
         }
         else {
             // First, figure out whether the notebook is in the workspace
-            if (this.notebook.uri.scheme==='file' && vscode.workspace.getWorkspaceFolder(this.notebook.uri) !== undefined) {
+            if (this.notebook.uri.scheme === 'file' && vscode.workspace.getWorkspaceFolder(this.notebook.uri) !== undefined) {
                 let currentFolder = path.dirname(this.notebook.uri.fsPath)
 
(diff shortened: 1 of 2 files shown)


src/notebook/notebookKernel.ts
```diff
@@ -163,9 +163,12 @@ export class JuliaKernel {
                 const execution = this._currentExecutionRequest
                 if (execution) {
                     execution.appendOutput(new vscode.NotebookCellOutput(items.map(item => {
-                        if (item.mimetype == 'image/png' || item.mimetype == 'image/jpeg') {
+                        if (item.mimetype === 'image/png' || item.mimetype === 'image/jpeg') {
                             return new vscode.NotebookCellOutputItem(Buffer.from(item.data, 'base64'), item.mimetype)
                         }
+                        else if (item.mimetype.endsWith('+json')) {
+                            return vscode.NotebookCellOutputItem.json(item.data, item.mimetype)
+                        }
```
davidanthoff, line 171: @DonJayamanne if I look at the saved version of a notebook file, this seems to properly set JSON style output. But for the vega-lite MIME types for example it doesn't render at all. But I think that might be a problem in the Jupyter extension at the moment, right?

DonJayamanne: Let me check.

davidanthoff: I merged this for now, and then will test with the vega fixes in Jupyter.

PR #2267, opened 2021-07-03
Save/Copy Plots

### LINKED ISSUE #172: Add copy to clipboard option to plot pane
There should be an option to copy the content of the plot pane to the clipboard.

### LINKED ISSUE #424: add 'save plot' button

AbdulrhmnGhanem:
Add `Save Plot` button, fixes #424.
Add `Copy Plot` button, fixes #172.

- [x] End-user documentation check.
- [x] Changelog mention.

##
- Relying on the graphical backends turned out to a bit problematic because it will need to keep a reference to the code which produced the plot.
- Even if the complexity of storing the code which produced the plot is acceptable, it won't work for calls to `plot` with mutable data e.g., `plot(1:10, rand(10))`.
- Instead, save the SVG displayed in the plot pane.


pfitzseb:
Nice! I'll take a more detailed look tomorrow, but could you please make sure to configure eslint so that it conforms to the code style we have going so far (no semi-colons etc). Really hard to read the diff otherwise (and it would be good to keep the codebase consistent).

src/utils.ts
```diff
@@ -88,3 +93,78 @@ function parseEnvVariables(p: string) {
         return process.env[variable] || ''
     })
 }
+
+type FileLike = string | Buffer;
+export class ClipboardManager {
```
pfitzseb, line 98: Can we just use the [`ClipboardItem`](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem/ClipboardItem) API instead? Browser support isn't great, but for the majority of users that isn't relevant.

Generally I'm very uncomfortable with shipping binaries here.

scripts/plots/main_plot_webview.js
```diff
@@ -1,42 +1,109 @@
...
+            ? src.replace(/data:image\/gif;base64,/, '')
+            : null
+
+        postMessageToHost(reqType, { svg, png, gif, index })
+    }
 }
 
 window.addEventListener('load', getImage)
+window.addEventListener('load', () => {
+    // Remove Plotly builtin export button; it's nonfunctional in VSCode and can confuse users.
```
pfitzseb, line 90: 👍

pfitzseb (changes requested):
Ok, the code here looks good (although there still are a bunch of weird eslint-looking changes around -- maybe we should just take the time to configure eslint properly and run it on the whole codebase).

That said, I don't think OS specific binaries are the way to go. We really want something that 
1. Runs in the renderer process (so that it uses the correct clipboard when developing on a remote).
2. Doesn't require us to ship binaries.

src/utils.ts
AbdulrhmnGhanem: > Generally I'm very uncomfortable with shipping binaries here.

So do I, but I tried several other ways and nothing worked.

The browser support isn't the problem, Clipboard Item [doesn't support SVGs ](https://stackoverflow.com/questions/62390200/domexception-sanitized-mime-type-image-svgxml-not-supported-on-write). Also, Electron has an experimental clipboard manager but the Electron API isn't accessible from the extension.

pfitzseb: Mhh. IMHO either converting to png or just using the source string are better than shipping binaries, even though  both of those options kinda suck too.

AbdulrhmnGhanem: To get the clipboard to work reliably with remote environments the [VSCode clipboard API](https://code.visualstudio.com/api/advanced-topics/remote-extensions#using-the-clipboard) must be used which only supports text via `readText`/`writeText` which makes it impossible to copy an SVG or a PNG.

pfitzseb: Are you sure about that? JS running in an in-editor iframe should have access to the clipboard, given that these iframes are created with `allow="clipboard-read; clipboard-write;"`. You just need to use the VSCode clipboard API when you want to copy something out of the extension host process.

AbdulrhmnGhanem: >Historically, extension authors have used Node.js modules such as clipboardy to interact with the clipboard. Unfortunately, if you use these modules in a Workspace Extension, they will use the remote clipboard instead of the user's local one. The VS Code clipboard API solves this problem. It is always run locally, regardless of the type of extension that calls it.

I might understand this wrong.
So should I go with png then?

pfitzseb: Yeah, the important difference here is where the extension code runs.
"Normal" extension code (e.g. all the TS code in this extension) will run on the remote machine (in an extension host process), so the clipboard that's available from there is not the one the user has immediate access to. JS in the iframe on the other hand always runs on your local machine/browser, so it should have access to your clipboard as well.

So imho it's worth investigating that solution, yes.

AbdulrhmnGhanem: Now, it uses the Clipboard API.

src/interactive/plots.ts
```diff
@@ -626,12 +829,83 @@ export function displayPlot(params: { kind: string, data: string }) {
...
+        _writePlotFile(fileName, plot.svg)
+    } else if (plot.png !== null) {
+        const fileName = `${plotName}.png`
+        const buffer = Buffer.from(plot.png, 'base64')
+        _writePlotFile(fileName, buffer)
+    } else if (plot.gif !== null) {
+        const fileName = `${plotName}.gif`
+        const buffer = Buffer.from(plot.gif, 'base64')
+        _writePlotFile(fileName, buffer)
+    }
```
pfitzseb, line 884: Would be good to add a warning message if this fails, just so the user isn't super confused as to what's going on.

src/interactive/plots.ts
```diff
@@ -626,12 +829,83 @@ export function displayPlot(params: { kind: string, data: string }) {
...
+
+/**
+ * Write the plot file to disk.
+ * @param fileName
+ * @param data
+ * @param encoding
+ */
+function _writePlotFile(fileName: string, data: FileLike) {
+    const rootPath = vscode.workspace.workspaceFolders[0].uri.fsPath
+    const plotsDir: string = vscode.workspace
```
pfitzseb, line 896: Hm, I think we should be using `window.showSaveDialog` here to prompt for a location; the default path and file name handling here are fine, of course.

pfitzseb (changes requested):
Looking good and works great! I've left two comments that I'd like to see addressed before merging this though.

Also, there still are a bunch of code style changes in here. Please revert them for now; we can figure out a proper eslint setup later on.

src/interactive/plots.ts
AbdulrhmnGhanem: Resolved in 824acb4.

src/interactive/plots.ts
AbdulrhmnGhanem: Resolved in f710ca1, and cc5586b.

pfitzseb (approved):

davidanthoff (changes requested):
@AbdulrhmnGhanem same thing as with the other one: would be great if you could resolve the merge conflict, and then we can include it in the next minor release.

davidanthoff (approved):

pfitzseb:
This is included in https://github.com/julia-vscode/julia-vscode/pull/2273, so closing.

Issue #2256, opened 2021-06-29
Kernel choice is not saved in notebook file


davidanthoff:
It seems that the kernelspec/metadata fields in notebooks are not populated properly when saving a Jupyter notebook.

@DonJayamanne this is also something that needs to be fixed in the Jupyter extension, right?

DonJayamanne:
Yes, sorry, was on holidays, today. Will work on this.

davidanthoff:
@DonJayamanne so this will essentially go https://github.com/julia-vscode/julia-vscode/blob/2bbf02bfec80f84992a3bf61b3a60d3ae2357fad/src/notebook/notebookFeature.ts#L161, right?

Do you know when the relevant upstream code will ship, i.e. when we can implement that? Is there an issue/PR upstream that I can follow?

davidanthoff:
@DonJayamanne and @claudiaregio, so I think this issue is the one I'm most worried about at the moment. My understanding from an email from Don is that we can't do anything about this, because some more APIs in VS Code itself would be needed?

I'm really almost torn whether we should release the whole notebook functionality before this is resolved at all (that is why this was one of our "must be done before release" items). The problem here is that users will actually save in some way faulty files if this is not resolved, and such faulty files can have a long lifetime... Say someone uses this to write some replication code for a paper and then deposits such files in zenodo.org or something like that (this happens all the time with scientific code). If they create that notebook file with VS Code without this being fixed, those files will miss really crucial information.

Is there a timeline for this? I.e. will be able to fix this with the July VS Code release? I am much less worried about rough edges around user experience, but when it comes to file formats, I think we really need to get it right from the start because those files will stay around even after we've updated the extension to a new version that might fix this...

mjbvz:
I have a PR out on the VS Code side that would allow setting the kernelspec on ipynb files:

https://github.com/microsoft/vscode/pull/131219

You would use this api through the built-in `ipynb` extension

Please let me know if this PR adds what you need for Julia

PR #2271, opened 2021-07-07
Support for preferred kernels

### LINKED ISSUE #2255: Automatic kernel selection for existing notebook doesn't work
When I open a notebook that was created in JupyterLab and had a specific Julia kernel set there, and then open the same notebook in VS Code, it does not automatically select our kernel to run things.

It is not clear to me whether we need to add functionality in say `onDidOpenNotebookDocument` to handle that (by calling `updateNotebookAffinity`) or whether the Jupyter extension will handle that?

@DonJayamanne do you know how this is supposed to work?

DonJayamanne:
Fixes #2255 
Part of #2256

### DIFF
--- a/src/notebook/notebookFeature.ts
+++ b/src/notebook/notebookFeature.ts
@@ -5,14 +5,35 @@ import { WorkspaceFeature } from '../interactive/workspace';
 import { getJuliaExePaths, JuliaExecutable } from '../juliaexepath';
 import { JuliaKernel } from './notebookKernel';
 
+const JupyterNotebookViewType = 'jupyter-notebook';
+type JupyterNotebookMetadata = Partial<{
+    custom: {
+        metadata: {
+            kernelspec: {
+                display_name: string;
+                language: string;
+                name: string;
+          },
+            language_info: {
+                name: string;
+                version: string;
+                mimetype: string;
+                file_extension: string;
+            }
+        }
+    }
+}>
+
 export class JuliaNotebookFeature {
-    private readonly _controllers: vscode.NotebookController[] = []
+    private readonly _controllers = new Map<vscode.NotebookController, { version: string}>();
     private readonly _juliaVersions = new Map<string, JuliaExecutable>()
     private readonly kernels: Map<vscode.NotebookDocument, JuliaKernel> = new Map<vscode.NotebookDocument, JuliaKernel>()
     private _outputChannel: vscode.OutputChannel
+    private readonly disposables: vscode.Disposable[] = [];
 
     constructor(private context: vscode.ExtensionContext, private workspaceFeature: WorkspaceFeature) {
         this.init()
+        vscode.workspace.onDidOpenNotebookDocument(this.onDidOpenNotebookDocument, this, this.disposables)
     }
 
     private async init() {
@@ -23,7 +44,6 @@ export class JuliaNotebookFeature {
         // Find the highest installed version per minor version
         for (const i of juliaVersions) {
             const ver = i.getVersion()
-
             const kernelId = `julia-${semver.major(ver)}.${semver.minor(ver)}`
 
             if (this._juliaVersions.has(kernelId)) {
@@ -39,8 +59,8 @@ export class JuliaNotebookFeature {
         // Add one controller per Julia minor version that we found
         for (const [kernelId, juliaVersion] of this._juliaVersions) {
             const ver = juliaVersion.getVersion()
-
-            const controller = vscode.notebooks.createNotebookController(kernelId, 'jupyter-notebook', `Julia ${ver} Kernel`)
+            const displayName = `Julia ${ver} Kernel`;
+            const controller = vscode.notebooks.createNotebookController(kernelId, JupyterNotebookViewType, displayName)
             controller.supportedLanguages = ['julia']
             controller.supportsExecutionOrder = true
             controller.onDidChangeSelectedNotebooks((e) => {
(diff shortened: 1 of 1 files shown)


src/notebook/notebookFeature.ts
```diff
@@ -39,8 +59,8 @@ export class JuliaNotebookFeature {
         // Add one controller per Julia minor version that we found
         for (const [kernelId, juliaVersion] of this._juliaVersions) {
             const ver = juliaVersion.getVersion()
-
-            const controller = vscode.notebooks.createNotebookController(kernelId, 'jupyter-notebook', `Julia ${ver} Kernel`)
+            const displayName = `Julia ${ver} Kernel`;
```
DonJayamanne, line 62: Do we need the suffix `Kernel`, without that, its identical to the name in `Jupyer`

src/notebook/notebookFeature.ts
```diff
@@ -74,6 +122,38 @@ export class JuliaNotebookFeature {
...
+            language: 'julia',
+            name: name
+        }
+        nbmetadata.custom.metadata.language_info = {
+            name: name,
+            version: version,
+            mimetype: 'text/julia',
+            file_extension: '.jl'
+        }
+        // TODO: Update the notebook metadata (when its stable).
```
DonJayamanne, line 145: The API used to udpate the notebook metadata is still proposed API.
Without this, any notebook created will not have the julia kernel information in the metadata.
We can make use of the proposed API, however it will not work in stable.

DonJayamanne: Let me check with the team

src/notebook/notebookFeature.ts
```diff
@@ -5,14 +5,35 @@ import { WorkspaceFeature } from '../interactive/workspace';
 import { getJuliaExePaths, JuliaExecutable } from '../juliaexepath';
 import { JuliaKernel } from './notebookKernel';
 
+const JupyterNotebookViewType = 'jupyter-notebook';
+type JupyterNotebookMetadata = Partial<{
```
DonJayamanne, line 9: Currently there are no type definitions for this, 
structure of the metadata. Same approach used by .NEt interactive extension as well.
I'm working on this (trying to get a typed approach to accessing this metadata)

DonJayamanne:
@davidanthoff here's the Jupyter PR to 
* promote Julia extension https://github.com/microsoft/vscode-jupyter/pull/6561
* & ensure ijulia kernels are not promoted as preferred controllers by the jupyter extension

src/notebook/notebookFeature.ts
```diff
@@ -49,11 +69,39 @@ export class JuliaNotebookFeature {
...
+        }
+        // Get metadata from notebook (to get an hint of what version of julia is used)
+        const { name, version } = this.getKernelSpecNameAndVersion(e);
+        let preferredControllerFound = false;
+        this._controllers.forEach((info, controller) => {
+            // If we find a controller that matches the vesion in the notebook metadata, then set
+            // that controller as the preferred controller.
+            if (name.includes(info.version) || version.includes(info.version)) {
+                preferredControllerFound = true;
+                controller.updateNotebookAffinity(e, vscode.NotebookControllerAffinity.Preferred)
```
DonJayamanne, line 96: Magic sauce, this is waht puts Julia on the top of the list.
To test, please open an existing Jualia notebook (created in Jupyter), & ensure you delete (temporarily move) the ijulia kernelspec .json file.

Or you can wait for the CI to complete & download the Jupyter VSIX from here https://github.com/microsoft/vscode-jupyter/actions/runs/1006379241 & test it all together.

src/notebook/notebookFeature.ts
davidanthoff: Agreed, we can remove and use the same name as in Jupyter.

src/notebook/notebookFeature.ts
```diff
@@ -49,11 +69,39 @@ export class JuliaNotebookFeature {
...
+        if (!this.isJuliaNotebook(e) || this._controllers.size === 0) {
+            return;
+        }
+        // Get metadata from notebook (to get an hint of what version of julia is used)
+        const { name, version } = this.getKernelSpecNameAndVersion(e);
+        let preferredControllerFound = false;
+        this._controllers.forEach((info, controller) => {
+            // If we find a controller that matches the vesion in the notebook metadata, then set
+            // that controller as the preferred controller.
+            if (name.includes(info.version) || version.includes(info.version)) {
```
davidanthoff, line 94: I think we probably want to make the first criterion for being a preferred controller whether the `kernelspec.name` matches the `kernelId` of the controller. That string will have the form `julia-1.6`, i.e. it will include the major and minor version part, but not the patch version. If we don't find a match on that, I think it is not clear to me whether we should set a preferred kernel at all. I think JupyterLab doesn't, for example...

The existing code will always pick the latest patch release in a given `major.minor` combo as the controller, if multiple Julia versions in a given `major.minor` series are installed, which matches the Jupyter experience. I _think_ that makes sense, but it does mean that say a notebook was saved with a Julia 1.6.0 kernel, and then is opened on a system that has Julia 1.6.0 and 1.6.1 installed, a kernel with 1.6.1 is going to be used. I'd say we can try that for now, I think it is what folks expect based on the existing Jupyter implementation.

src/notebook/notebookFeature.ts
```diff
@@ -49,11 +69,39 @@ export class JuliaNotebookFeature {
...
+                preferredControllerFound = true;
+                controller.updateNotebookAffinity(e, vscode.NotebookControllerAffinity.Preferred)
+            }
+        })
+        if (!preferredControllerFound) {
+            // We know its a Julia notebook, hence give preference to one of our controllers.
+            const preferredController = Array.from(this._controllers.keys())[0];
+            preferredController.updateNotebookAffinity(e, vscode.NotebookControllerAffinity.Preferred)
+        }
```
davidanthoff, line 103: I'm pretty sure that JupyterLab for example prompts the user to select a kernel in that case, maybe we should mimic that as well? Or we could come up with something more clever. Say if there is only one Julia version installed, we use that. If there are multiple, we use the newest (unless that is older than the version in the notebook file). Or something like that... But I think maybe for starters I would probably just not set anything preferred and trigger the UI for kernel selection in this case.

src/notebook/notebookFeature.ts
davidanthoff: Does "not work" mean it will crash, or it will just not save things properly? I think we should keep `master` in a state now where we can tag insider builds and ship them to our normal user base without breaking them.

davidanthoff (approved):
Fantastic!

I have some comments around the preferred kernel selection logic. Let me know if you want to update the PR, otherwise we can also merge as is and I can create another PR on top that handles that aspect.

src/notebook/notebookFeature.ts
davidanthoff: Oh, and this is something that has to ship in VS Code itself, right? So we are tied with this to the VS Code release cycle, not the Jupyter extension cycle?

davidanthoff:
I'm merging now, and will do some follow up PRs for some of the points I raised.

PR #2273, opened 2021-07-07
Interactive (zoomable/pannable) Plots

### LINKED ISSUE #1227: zoomable and panable plot pane as in Juno
context: https://github.com/JunoLab/Atom.jl/pull/327 and https://discourse.julialang.org/t/ann-juno-0-12-5/39744

### LINKED ISSUE #1655: Interactive plots?
Having the ability to plot within the IDE is awesome to begin with but.....
The surface(3-D) plots can't be interacted with, as in the case of qt and x11 interface. By interaction, I mean stuff like rotating using the mouse, or zooming. Any chance we can get a better version of the Plots pane, which is friendlier to use?

### LINKED ISSUE #1878: Allow plot pane to be zoomed independently of other panes
By default, plots often end up bigger than the plot pane.
It would be useful to be able to zoom in and out of plots in the plot pane.

AbdulrhmnGhanem:
Fixes #1655, fixes #1227, and fixes #1878.

- [x] Changelog mention.


scripts/plots/main_plot_webview.js
```diff
@@ -1,42 +1,233 @@
...
+function removePlotlyBuiltinExport() {
+    if (isPlotly()) {
+        document.querySelector(
+            '[data-title="Download plot as a png"]'
+        ).style.display = 'none'
+    }
+}
+
+function initSvgPanZoom() {
+    const svgTag = document.querySelector('svg')
```
pfitzseb, line 187: This doesn't work for non-svg plots. I haven't really looked at svg-pan-zoom, but would expect it to work fine with other `img`s as well, so maybe 
```suggestion
    const svgTag = getPlotElement()
```
would work fine? If not, then this a `isSVG` guard.

AbdulrhmnGhanem: * Sometimes the `plot_element` is a `div` wrapping the actual plot. 
* All `Plots` backends are displayed as SVG, `Plotly` is ignored because it has its own controllers.
* `Vega` plots were displayed as `canvas` but I had reconfigured it to be displayed as `svg`.

> but would expect it to work fine with other imgs as well, so maybe

It only supports `<svg>` tags, even `img` with svg `src` isn't supported, but it isn't a problem as explained above.

davidanthoff (changes requested):
@AbdulrhmnGhanem we are about to open our merge window for the next minor release, could you resolve the merge conflict here so that we can include this PR? Thanks!

davidanthoff (approved):
Fantastic, thanks!

PR #2276, opened 2021-07-09
Select notebook kernel just on name match

davidanthoff:
@DonJayamanne I think this is the logic we should go with, i.e. purely select on name/id match. If the user doesn't have the Julia version installed that matches on the x.y part of the version, ideally the user would be prompted whether they want to use some other installed Julia or not. We should probably actually even show some sort of message in that case "You opened a notebook that uses Julia 1.5, but you don't have that installed, please pick some other Julia version".

### DIFF
--- a/src/notebook/notebookFeature.ts
+++ b/src/notebook/notebookFeature.ts
@@ -13,7 +13,7 @@ type JupyterNotebookMetadata = Partial<{
                 display_name: string;
                 language: string;
                 name: string;
-          },
+            },
             language_info: {
                 name: string;
                 version: string;
@@ -25,7 +25,7 @@ type JupyterNotebookMetadata = Partial<{
 }>
 
 export class JuliaNotebookFeature {
-    private readonly _controllers = new Map<vscode.NotebookController, { version: string}>();
+    private readonly _controllers: vscode.NotebookController[] = []
     private readonly _juliaVersions = new Map<string, JuliaExecutable>()
     private readonly kernels: Map<vscode.NotebookDocument, JuliaKernel> = new Map<vscode.NotebookDocument, JuliaKernel>()
     private _outputChannel: vscode.OutputChannel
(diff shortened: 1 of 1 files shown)


DonJayamanne:
> "You opened a notebook that uses Julia 1.5, but you don't have that installed, please pick some other Julia version".

FYI - regardless of whether we find the exact kernel or not, the user will always have to select an item from the list. Setting the affinity merely pushes the preferred item to the top (i.e. the preferred is on top).

> but you don't have that installed, please pick some other Julia version".

Not sure this is necessary & will make much difference. Again, the user will have to select a kernel, and they'll see that the kernel they desire isn't on the list.
Will we need a prompt, not sure, then again, I don't know about the julia users.
When running Python notebooks we don't display such a message.

src/notebook/notebookFeature.ts
```diff
@@ -122,15 +115,14 @@ export class JuliaNotebookFeature {
...
         const metadata = (notebook.metadata as JupyterNotebookMetadata)?.custom.metadata;
         const kernelspecName = metadata?.kernelspec?.name || '';
-        const version = metadata?.language_info?.version || '';
-        return this.isJuliaNotebook(notebook) ? { name: kernelspecName, version } : {name:'', version:''}
+        return this.isJuliaNotebook(notebook) ? kernelspecName : ''
     }
     private updateNotebookWithSelectedKernel(notebook: vscode.NotebookDocument, name: string, version: string) {
         // Dont edit in place, create a copy of the metadata.
-        const nbmetadata: JupyterNotebookMetadata = JSON.parse(JSON.stringify((notebook.metadata || { custom: { metadata: {}}})));
+        const nbmetadata: JupyterNotebookMetadata = JSON.parse(JSON.stringify((notebook.metadata || { custom: { metadata: {} } })));
```
DonJayamanne, line 125: What formatter do you use, i use is prettier and it seems to format the whole document & don't know whether my code is inline with the formatting/style that you follow in this repo.

DonJayamanne (approved):

pfitzseb (approved):

davidanthoff:
> FYI - regardless of whether we find the exact kernel or not, the user will always have to select an item from the list.

@DonJayamanne why is that? My understanding was that we would try to avoid showing any UI to select a kernel if the right one is present. For example, if a user opens an existing notebook that has a kernel named `julia-1.6` in the kernel metadata in the file, and our extension finds an installed version of Julia 1.6.2 and registers a notebook controller with name `julia-1.6`, then really I think we should not show any UI for picking a kernel, but instead that notebook controller should just be used automatically. That is how all other Jupyter notebook clients work. What would be the point of asking a user to select a kernel in that case? That seems just unnecessary extra work for them?

I have to admit that I don't understand the current design in VS Code with these preferred notebook controllers, and what that concept actually means. It seems to me that VS Code should behave in the following way on notebook file open:
1. first look if there is a matching kernel on id/name, if there is, pick that, don't show any UI, be done
2. if step 1 didn't lead to a decision, then show a UI picker where _all_ kernels are displayed that support the Julia language, if that was the language that is stored in the metadata of the file.

I don't understand where the concept of a preferred notebook controller factors into any of this... 

Also CCing @claudiaregio.

davidanthoff:
I have a new PR https://github.com/julia-vscode/julia-vscode/pull/2315 that I think is better than this PR here, so probably treat this here as outdated and take a look at the other PR instead.

DonJayamanne:
> understanding was that we would try to avoid showing any UI to select a kernel if the

Tried to explain this in a seprate comment in another PR in this repo.

PR #2296, opened 2021-07-19
Overhaul keybindings

pfitzseb:
Supersedes https://github.com/julia-vscode/julia-vscode/pull/1785.

This mostly follows the keybindings for notebooks, so `Meta+Enter` evaluates the current block, `Shift+Enter` also jumps to the next one. (Shift+)Alt+Enter executes the current "cell" (and jumps to the next one).

Furthermore:
- `Escape` clears the current inline result if there is one, because the notebook interface uses it as well (to exit the current cell's editor). 
- `Ctrl+Shift+C` interrupts the Julia REPL from anywhere
- `Alt+J Alt+P` shows the plot pane
- `Alt+J Alt+W` shows the workspace
- There is no longer a default keybinding for sending code to the REPL.


davidanthoff:
Should we queue this for 1.4, i.e. the release after the release for Juliacon? I think we're too close to release at the moment, i.e. ideally I think we should really just be bug fixing right now.

pfitzseb:
This is kinda necessary for the notebook UI to work properly, but since that's off by default we can indeed wait until after JuliaCon.

davidanthoff:
So the one thing that I'm a bit worried about is that `Alt+Enter` kind of deviates from the Jupyter story, _and_ with this essentially both of the "main" keybindings we had previously for executing the "current" code and move to the next line no longer do that... I'm not really sure whether there is a way out, but this is really very disruptive...

One random other thought: we could not have different keybindings for the cell and block story, but instead just have one set of bindings work on cells if there are cell delimiters in the file, and on blocks otherwise. Not sure whether that really makes sense, but it doesn't strike me as crazy?

pfitzseb:
We could introduce a `julia.smartExecute` that intelligently switches between all three behaviours (eval selection, eval block, eval cell) and bind that by default. Or just remove all default keybindings for cells...

davidanthoff:
> We could introduce a `julia.smartExecute` that intelligently switches between all three behaviours (eval selection, eval block, eval cell) and bind that by default.

davidanthoff:
> We could introduce a `julia.smartExecute` that intelligently switches between all three behaviours (eval selection, eval block, eval cell) and bind that by default.

Yeah, I _think_ I like that best. Not 100% sure, though, as I haven't really thought very much about it... Maybe we should ask on the slack channel for some feedback what folks would think about that?

> Or just remove all default keybindings for cells

Haha, I'd be game, I never liked it. But we might have a revolt at hand if we do that...

PR #2306, opened 2021-07-23
Clear cell output when cell is queued for exec

davidanthoff:
@DonJayamanne, this errors because apparently one can only delete cell output during execution of a cell. I think we would really want to delete cell output when the cell gets queued, though, like in other Jupyter clients. Is there some way to do that?

### DIFF
--- a/src/notebook/notebookKernel.ts
+++ b/src/notebook/notebookKernel.ts
@@ -67,10 +67,7 @@ export class JuliaKernel {
 
         // Now create execution object that actually will run the code
         const execution = this.controller.createNotebookCellExecution(cell)
-        execution.executionOrder = executionOrder
-
-        await execution.clearOutput()
-
+        execution.token.onCancellationRequested(e=>execution.end(undefined))
         this._scheduledExecutionRequests.push(execution)
 
         this._processExecutionRequests.notify()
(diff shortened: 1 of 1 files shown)


src/notebook/notebookKernel.ts
```diff
@@ -88,9 +87,6 @@ export class JuliaKernel {
                     const runStartTime = Date.now()
                     this._currentExecutionRequest.start(runStartTime)
 
-                    // TODO Ideally we would clear output at scheduling already, but for now do it here
-                    await this._currentExecutionRequest.clearOutput()
```
DonJayamanne, line 92: Clearing of output via the Execution task can only be done after you start the execution task.
So if we want to clear output when we queue, then we'll need to create an execution task as follows:
* Create execution task
* Call start on the execution task
* Clear the output
* End the execution task 

I.e. we must call start before calling clear.
In this case its best to create a whole new execution task prior to queueing of the cells.

src/notebook/notebookKernel.ts
```diff
@@ -64,8 +64,7 @@ export class JuliaKernel {
         const execution = this.controller.createNotebookCellExecution(cell)
         execution.executionOrder = executionOrder
 
-        // TODO For some reason this doesn't work here
-        // await execution.clearOutput()
+        await execution.clearOutput()
```
DonJayamanne, line 67: Clearing of output via the Execution task can only be done after you start the execution task.
In this case its best to create a whole new execution task prior to queueing of the cells.

I.e. i'd change as follows:
* Create an execution task, start it, clear the output & then end that execution task (we're now done with this execution task).
* Then create another new execution task & push into the `scheduledExecutionTasks`.

davidanthoff: Ah, yes, that works, I just pushed that.

It is not ideal, though :) Now when one queues a large notebook, for all cells the stopwatch "running" status icon briefly appears, then the output disappears, and then the status changes to the queued state. Maybe the underlying code could eventually allow calls to `clearOutput` _before_ one even starts execution? I'll merge this for now, though.

I also changed when an execution order gets assigned, namely when execution of a cell starts, not when it is queued. That brings things more in line with other clients.

And I changed cancel token handling a bit, so that queued cells that get cancelled immediately leave the queued state.

PR #2313, opened 2021-07-24
Add a debug launch for the LSP

davidanthoff:
With this the LSP process starts in debug mode when we debug the extension. Breakpoints in request handlers? Check!

Probably still a bit iffy, and the fact that the debugger doesn't handle `@async` will probably be a problem, but crudly it works and is probably already helpful.

@ZacLN just pinging you here to make sure you see this, I think this might make working on the LS a lot easier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /dist/
 /out/
 /scripts/languageserver/julia_pkgdir/lib/
+/globalstorage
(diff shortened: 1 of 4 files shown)


davidanthoff:
Alright, I think this is actually ready to be merged.

pfitzseb (approved):

PR #2329, opened 2021-07-31
Extensible diagnostics display

pfitzseb:
Ok, so this allows types to display arbitrary in-editor traces with something like
```julia
struct Frame
    path::String
    line::Int
    msg::String # optional
end
struct JETReport
    trace::Vector{Frame}
end

Base.showable(::MIME"application/vnd.julia-vscode.trace", ::JETReport) = true
Base.show(::IO, ::MIME"application/vnd.julia-vscode.trace", t::JETReport) = t.trace

jr = JETReport(
    [
        Frame(@__FILE__, 3, "blah"),
        Frame(@__FILE__, 5, "blub")
    ]
)
```

User code needs to define `show`/`showable` for the `application/vnd.julia-vscode.trace`; the `show` method is kinda scuffed in the sense that it needs to return a vector that `JSON.json`s to something like
```
[
  {
    "path": "foo",
    "msg": "blah",
    "line": 2
  },
  ...
]  
```
All of that should be completely harmless in other contexts though.

Output will look like this:
image

I'm planning to add another optional field that allows specifying a variant (`danger` for errors, `info`, `warning` etc).

cc @aviatesk


src/interactive/repl.ts
```diff
@@ -376,6 +376,19 @@ function clearProgress() {
...
+    if (params.kind === 'application/vnd.julia-vscode.trace') {
+        displayTrace(params)
+    } else {
+        plots.displayPlot(params)
+    }
+}
+
+function displayTrace(params: { kind: string, data: any }) {
+    results.clearStackTrace()
+    results.setStackFrameHighlight('', params.data)
```
aviatesk, line 389: I want to have this like:
```julia
    results.setStackFrameHighlight(params.data.all, params.data.frames)
```
so that we can have messages for all the frames, but then it won't be clear how to handle multiple traces.

Maybe we can change the interface of `TRACE_MIME` so that it can take an array of traces ?

aviatesk: Never mind, I think this will complicate the interface, and we can just attach message for each frame.

davidanthoff:
I have to admit I don’t even understand roughly what this might be for? Could you give some context? What kind of feature is this?

pfitzseb:
So the idea here is to allow packages like JET.jl to provide a `show` method that VSCode can translate into diagnostics like so:
image

It (ab)uses the display system to do so, and as such doesn't require any package loading hooks; everything is handled via MIME types.

The JET side of the integration lives at https://github.com/aviatesk/JET.jl/pull/227.

scripts/packages/VSCodeServer/src/display.jl
```diff
@@ -82,6 +82,33 @@ const DISPLAYABLE_MIMES = [
...
 
+"""
+    DIAGNOSTIC_MIME = "application/vnd.julia-vscode.diagnostics"
+
+Trace format is:
+```
+[
+    Dict(
+        "msg" => ""::AbstractString,
+        "path" => ""::AbstractString,
```
davidanthoff, line 93: Should we use URIs here from the get go? Identifying notebook locations for example always requires URIs, so just settling on that might be easier?

src/interactive/repl.ts
```diff
@@ -376,6 +376,63 @@ function clearProgress() {
...
+
+    if (g_trace_diagnostics.has(source)) {
+        g_trace_diagnostics.get(source).clear()
+    } else {
+        g_trace_diagnostics.set(source, vscode.languages.createDiagnosticCollection('Julia Diagnostics: ' + source))
+    }
+
+    const diagnostics = params.data.items.map((frame): [vscode.Uri, vscode.Diagnostic[]] => {
+        const diagnostic = new vscode.Diagnostic(
+            new vscode.Range(frame.line - 1, 0, frame.line - 1, 99999),
```
davidanthoff, line 406: Maybe the start column should just be the first char on that line that isn't a space or tab? I think all our other diagnostics looks "good" in that sense now?

src/interactive/repl.ts
```diff
@@ -376,6 +376,63 @@ function clearProgress() {
...
+    relatedInformation: {msg: string, path: string, line: number}[]
+}
+const g_trace_diagnostics: Map<string, vscode.DiagnosticCollection> = new Map()
+function displayDiagnostics(params: { kind: string, data: { source: string, items: diagnosticData[] } }) {
+    const source = params.data.source
+
+    if (g_trace_diagnostics.has(source)) {
+        g_trace_diagnostics.get(source).clear()
+    } else {
+        g_trace_diagnostics.set(source, vscode.languages.createDiagnosticCollection('Julia Diagnostics: ' + source))
```
davidanthoff, line 401: I would prefix this with something like `User provided Julia Diagnostics: ` or something like that, just to make clear that whatever message shows up here does not originate with us.

davidanthoff (commented):
I think this should be hidden behind a config flag with a "experimental" label and maybe ideally off by default? I understand that this could be very useful to have while JET.jl matures, but at the same time this generally doesn't strike me as a great UX model that arbitrary running code can spit out diagnostic messages. I'm also generally a bit hesitant to put things into the main extension that most likely really only will be used by a very small fraction of our user base (either because they are not of general enough interest or because they are too difficult to use), but I think maybe in this case a config flag and an "experimental" label would strike on OK balance?

Is the eventual goal to integrate this with the LS and make this available like regular diagnostic messages?

aviatesk:
> I think this should be hidden behind a config flag with a "experimental" label and maybe ideally off by default?
> I'm also generally a bit hesitant to put things into the main extension that most likely really only will be used by a very small fraction of our user base

Hm, I still don't think we need to hide it by default ? The custom diagnostic is displayed only when user explicitly uses a package that provides diagnostics. It never appears as far as user doesn't use a provider package and it's already "opt-in" feature in that sense, and so I think we really don't need another layer of configuration.

> Otherwise we commit to a larger API surface and just could just mean more work down the road, with no real benefit now because we can just pick the one format that works well for JET.jl, right?

Well, other code analysis tools can also benefit from this extensible diagnostic interface. For example, @timholy may be interested in this because we can use this interface to nicely render the result of [SnoopCompile.jl's `suggest`](https://timholy.github.io/SnoopCompile.jl/stable/reference/#SnoopCompile.suggest), which can help developers improve latency problems more easily.

package.json
```diff
@@ -268,6 +268,10 @@
                 "command": "language-julia.activateFromDir",
                 "title": "Julia: Activate Parent Environment"
             },
+            {
+                "command": "language-julia.clearDiagnostics",
```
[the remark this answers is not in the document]
aviatesk: The "user" terminology can be ambiguous I think, because it's not users who provide diagnostics but it's a provider package. Maybe we can rename it as `clearCustomDiagnostics` and it will make it more explicit that it's different from the diagnostics provided by LS ?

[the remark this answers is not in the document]
pfitzseb: Right, the command really should be named `language-julia.clearRuntimeDiagnostics`.

scripts/packages/VSCodeServer/src/display.jl
pfitzseb: Julia code generally doesn't have access to URIs, so I don't see how that would work out. Unrelated, but I think we need to change the file path passed to notebooks so that it includes cell info or something like that.

scripts/packages/VSCodeServer/src/display.jl
```diff
@@ -82,6 +82,33 @@ const DISPLAYABLE_MIMES = [
...
+        "line" => 0::Int
```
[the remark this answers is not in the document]
pfitzseb: They are. This was just a random value to show the format.

scripts/packages/VSCodeServer/src/display.jl
```diff
@@ -82,6 +82,33 @@ const DISPLAYABLE_MIMES = [
...
+    ),
+    ...
+]
+```
+or everything JSON-compatible with that (so e.g. a struct with those fields).
```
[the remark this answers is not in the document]
pfitzseb: I mean, we can change the docs so that only named tuples or dicts or whatever work; the implementation basically prescribes "if JSON and only JSON is loaded, any object that JSON-serializes to something of that format works". I don't see any benefit to checking this anywhere.

src/interactive/repl.ts
pfitzseb: As above, `Julia Runtime Diagnostics: ...` is probably most accurate, but this also doesn't show up anywhere.

src/interactive/repl.ts
pfitzseb: Yeah, fair.

pfitzseb: I've added an additional `range` argument; doing this automatically when only `line` is provided is hard(ish) and brittle.

scripts/packages/VSCodeServer/src/display.jl
```diff
@@ -82,6 +87,41 @@ const DISPLAYABLE_MIMES = [
...
+            range = [[1, 2], [1, 4]] # or [[start_line, start_char], [end_line, end_char]],
+            severity = 1, # optional; 0: Error, 1: Warning, 2: Information, 3: Hint
+            relatedInformation = [
+                (
+                    msg = "foobar",
+                    path = "/some/other/absolute/path.jl",
+                    line = 1,
+                    range = [[1, 2], [1, 4]] # or [[start_line, start_char], [end_line, end_char]],
+                )
+            ] # optional
```
aviatesk, line 111: I think this isn't actually optional ?
https://github.com/julia-vscode/julia-vscode/blob/fa7ab65ae1c94ef50334cd3cf1bcd9983ff2bc2f/src/interactive/repl.ts#L424

scripts/packages/VSCodeServer/src/display.jl
```diff
@@ -82,6 +87,41 @@ const DISPLAYABLE_MIMES = [
...
+        ),
+        ...
+    ]
+)
+```
+One of `line` or `range` needs to be specified for each item and relatedInformation.
```
aviatesk, line 117:
```suggestion
One of `line` or `range` needs to be specified for each item and `relatedInformation`.
```

scripts/packages/VSCodeServer/src/display.jl
pfitzseb: Ah. It should be though :P

aviatesk (approved):
I think this PR is in good shape.

aviatesk:
Can we move this forward, @pfitzseb @davidanthoff ?

timholy:
Thanks for pinging me on this. @davidanthoff, by way of explanation: one good use for this kind of feature might be in-source type annotation. Instead of saying "check `@code_warntype`, which requires users to learn how to read type-inferred code that looks nothing like "regular" julia code, it might be nice to annotate lines or even expressions with the results of type-inference. To really do this kind of thing well I think we'll need https://github.com/JuliaLang/julia/issues/31162, but even getting the line number is a step forward.

aviatesk:
Let me merge this by today -- it shouldn't break any of existing features.

pfitzseb:
Might be worth doing one last patch release and then moving on to all the new features.

aviatesk:
So we basically want to wait for #2347 to be merged, right ?

src/interactive/repl.ts
```diff
@@ -376,6 +376,97 @@ function clearProgress() {
...
+    } else {
+        plots.displayPlot(params)
+    }
+}
+
+interface diagnosticData {
+    msg: string,
+    path: string,
+    line?: number,
+    range?: number[][],
```
davidanthoff, line 391: That seems strange to me, to have both `line` and `range`. I think we should just mimic the diagnostic API in VS Code here, so have one field `range` that has fields `start` and `end`, and these each have a field `line` and `character`.

src/interactive/repl.ts
```diff
@@ -376,6 +376,97 @@ function clearProgress() {
...
+    if (params.kind === 'application/vnd.julia-vscode.diagnostics') {
+        displayDiagnostics(params)
+    } else {
+        plots.displayPlot(params)
+    }
+}
+
+interface diagnosticData {
+    msg: string,
+    path: string,
```
davidanthoff, line 389: Should this maybe be a `uri`? The entire VS Code ecosystem seems to use Uris everywhere now, and that might make it easier down the road to for example designate locations in notebooks?

scripts/packages/VSCodeServer/src/display.jl
```diff
@@ -82,6 +87,41 @@ const DISPLAYABLE_MIMES = [
     "image/gif"
 ]
 
+"""
+    DIAGNOSTIC_MIME = "application/vnd.julia-vscode.diagnostics"
+
+User type needs to implement a `show` method that returns a named tuple or dictionary like the following
+```
+Base.show(io::IO, ::MIME"application/vnd.julia-vscode.diagnostics", t::YourType) = (
+    source = ""
```
davidanthoff, line 96:
```suggestion
    source = "Name of my diagnostic tool"
```

src/interactive/repl.ts
```diff
@@ -376,6 +376,97 @@ function clearProgress() {
...
+    severity: number,
+    relatedInformation?: {
+        msg: string,
+        path: string,
+        line?: number,
+        range?: number[][]
+    }[]
+}
+const g_trace_diagnostics: Map<string, vscode.DiagnosticCollection> = new Map()
+function displayDiagnostics(params: { kind: string, data: { source: string, items: diagnosticData[] } }) {
```
davidanthoff, line 401: Wouldn't it make more sense to just pass the `data` part of `params` to this function?

davidanthoff (changes requested):

davidanthoff:
> Instead of saying "check @code_warntype, which requires users to learn how to read type-inferred code that looks nothing like "regular" julia code, it might be nice to annotate lines or even expressions with the results of type-inference.

Yeah, that would be useful, but at the same time I think a good user experience for a feature like that would be to have it as part of the LS diagnostics, i.e. without the need for the user to run some code in the REPL, but just as part of the usual static analysis that is happening in the background as users type. I'm OK that we merge this now, I just would generally like to keep a focus in the extension on exposing features in a simple, more or less automatic way to users :)

> So we basically want to wait for #2347 to be merged, right ?

The way we've done this in the last year is that we open a merge window for PRs that are marked as minor, and then we merge them all at the same time. We typically open that window when we feel that there are no more urgent bugs on `master` that we want to address first in patch releases. I think at the moment it would be great if we could patch a couple more things from crash reporting before we start to merge new features.

src/interactive/repl.ts
pfitzseb: Either is possible. `range` takes precedence. Both `[[], []]` and the `Range`/`Position` formats are used in VSCode. This just mimics the VSCode API we're using later on. Generally I'd like to have as few types as possible here.

src/interactive/repl.ts
pfitzseb: We can't, unless Julia is also URI aware (which it isn't). So yes, possible that this is something we need to tackle in the future, but definitely not here.

src/interactive/repl.ts
pfitzseb: Ye.

pfitzseb:
> Yeah, that would be useful, but at the same time I think a good user experience for a feature like that would be to have it as part of the LS diagnostics, i.e. without the need for the user to run some code in the REPL, but just as part of the usual static analysis that is happening in the background as users type. I'm OK that we merge this now, I just would generally like to keep a focus in the extension on exposing features in a simple, more or less automatic way to users :)

We will definitely not get anywhere near as good diagnostics in the LS anytime soon. The whole abstract interpreter framework kinda paves the way for that, but it's still a long road (especially until the tooling is fast enough to continuously run).

Generally we shouldn't just cater to "users", but also "devs", imo. For the latter on-demand analysis is important, and we should think about making the UX nice for them.

In general we should think about getting more runtime info exposed in the frontend. We'll need to think carefully on how to do that, but it's important for a substantial part of our users. Some of the functionality might go into the LS at some point (if we ever get to a Roslyn-like compiler-integrated LS), but, as I said above, that's not gonna happen in the near future.

src/interactive/repl.ts
davidanthoff: Still want to do this?

davidanthoff (approved):
One small thing that could be done, but doesn't have to.

src/interactive/repl.ts
pfitzseb: Yes :)

timholy:
I'm sure it's obvious to everyone, but just in case: the biggest problem with hoping to do inferrability at the LS level is that the same method might be fully-inferrable for some argument types and not for others. Julia's compiler specialization makes this a bit of a hard problem.

An interesting question: what if a user runs a test suite and now had 37 different `MethodInstance`s of the same method. Is there a sense in which inference results across all of them should be integrated?

PR #2332, opened 2021-07-31
wip: inline evaluation: handle jump to unsaved editors

aviatesk:
Just found VSCode API doesn't support this.


aviatesk:
VSCode doesn't provide an API to traverse all editors (we have `vscode.window.visibleTextEditors`, but it's not enough to handle any editor that isn't necessarily visible). So AFAIU there is no good way to handle it. Let me close it.

PR #2334, opened 2021-08-02
Use JuliaFormatter instead of DocumentFormat

pfitzseb:
Requires https://github.com/julia-vscode/LanguageServer.jl/pull/972.

Also fixes our `.gitmodules`.


.gitmodules
```diff
@@ -1,34 +1,31 @@
-[submodule "scripts/languageserver/packages/JSON"]
+[submodule "scripts/packages/JSON"]
 	path = scripts/packages/JSON
 	url = https://github.com/JuliaIO/JSON.jl.git
-[submodule "scripts/languageserver/packages/LanguageServer"]
+[submodule "scripts/packages/LanguageServer"]
 	path = scripts/packages/LanguageServer
 	url = https://github.com/julia-vscode/LanguageServer.jl.git
-[submodule "scripts/languageserver/packages/URIParser"]
+[submodule "scripts/packages/URIParser"]
```
davidanthoff, line 7: Not for this PR, but we should eventually replace all uses of this with URIs.

.gitmodules
```diff
@@ -58,3 +55,24 @@
 [submodule "scripts/packages/IJuliaCore"]
 	path = scripts/packages/IJuliaCore
 	url = https://github.com/JuliaLang/IJuliaCore.jl.git
+[submodule "scripts/packages/Crayons"]
+	path = scripts/packages/Crayons
+	url = https://github.com/KristofferC/Crayons.jl.git
+[submodule "scripts/packages/DataAPI"]
```
davidanthoff, line 61: I don't think we need that? At least I wasn't able to figure out which package might be pulling this in?

package.json
```diff
@@ -832,72 +832,13 @@
...
-                    "default": true,
-                    "description": "Format inline documentation.",
-                    "scope": "window"
-                },
-                "julia.format.keywords": {
-                    "type": "bool",
-                    "default": true,
-                    "description": "Ensure single spacing following keywords.",
-                    "scope": "window"
+                    "scope": "application"
```
davidanthoff, line 835: I'm wondering whether we should try to just get rid of all formatting config values in VS Code? If we are already going for file system config files, shouldn't we just have a user-level JuliaFormatter specific config file for settings as well, that is not tied to VS Code? I kind of find the mix of user level config in VS Code but then workspace level in a file confusing.

package.json
```diff
@@ -832,72 +832,13 @@
...
-                        "none",
-                        "single",
-                        "off"
-                    ],
-                    "scope": "window"
+                "julia.format.margin": {
+                    "type": "integer",
+                    "default": 100,
+                    "description": "Maximum line length for formatting.",
+                    "scope": "application"
```
davidanthoff, line 841: I think we had said that the default should be to never do line-breaks, right? So that would suggest we either need a different default here, although I would much prefer to just move all config out of VS Code for formatting.

scripts/environments/languageserver/Manifest.toml
```diff
@@ -9,19 +15,49 @@ path = "../../packages/CSTParser"
...
+
+[[Crayons]]
+path = "../../packages/Crayons"
+uuid = "a8cc5b0e-0ffa-5ad4-8c14-923d3ee1735f"
+version = "4.0.4"
+
+[[DataAPI]]
+path = "../../packages/DataAPI"
+uuid = "9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a"
+version = "1.7.0"
```
davidanthoff, line 38: As above, I don't think we need this.

scripts/environments/languageserver/Manifest.toml
```diff
@@ -45,38 +81,76 @@ path = "../../packages/JSONRPC"
...
 uuid = "2b0e0bc5-e4fd-59b4-8912-456d1b03d8d7"
 version = "4.1.1-DEV"
 
+[[LibCURL]]
+deps = ["LibCURL_jll", "MozillaCACerts_jll"]
+uuid = "b27032c2-a3e7-50c8-80cd-2d36dbcbfd21"
+
+[[LibCURL_jll]]
+deps = ["Artifacts", "LibSSH2_jll", "Libdl", "MbedTLS_jll", "Zlib_jll", "nghttp2_jll"]
+uuid = "deac9b47-8bc7-5906-a0fe-35ac56dc84c0"
```
davidanthoff, line 102: I think this manifest file was created with a recent Julia version, right? I don't think this will run on Julia 1.0, or for that matter on any pre artifact version of Julia. The auto-update script for these environments always uses Julia 1.3.1 so that these files are in a backwards compatible format, which we probably need to do here as well.

davidanthoff (changes requested):
Excellent!

I think we also need https://github.com/domluna/JuliaFormatter.jl/pull/256 upstream.

.gitmodules
pfitzseb: JSON, but we have an older version that doesn't need it.

davidanthoff: Yes, we are on an older version of JSON because it doesn't pull in anything. But even `JSON#master` doesn't seem to pull this in? I think it pulls in Parsers.jl? But in any case, I guess we can remove, right?

pfitzseb: Right, it's a transitive dependency. I was mostly just looking at https://juliahub.com/ui/Packages/JuliaFormatter/OWZnm/0.15.2?t=1 and mentally filtering out what I knew were dependencies of JSON, but missed this one.

davidanthoff:
I'm putting this on the backlog for now, as we are about to open the merge window for the next minor release, but this is not yet ready because we first need https://github.com/domluna/JuliaFormatter.jl/pull/256.

pfitzseb:
Putting this back into "Next Minor" because I really want this sooner rather than later :)

Issue #2294, opened 2021-07-19
Stop asking Julia to resolve Julia path


pfitzseb:
https://github.com/julia-vscode/julia-vscode/blob/5c3d51d4ace5b022c5dd716030eadeb1f2438a43/src/juliaexepath.ts#L142
can be really annoying when the user wants to use e.g. a shell script to start Julia (to set env variables or do other arbitrary things). We should probably be using `which` or `where` directly here.

davidanthoff:
I would actually really like to get to a point where we _don't_ have to resolve the path at all, but can just start everything with a command, so that things like say `julia +1.5` (from `juliaup`) also work. I wanted to look into that soon as well, when I start with the proper `juliaup` integration. Which will make things a lot easier because we won't have to search paths or anything like that anymore, at least on Windows.

PR #2340, opened 2021-08-03
Simplify executable path lookup

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2294.

### DIFF
--- a/src/juliaexepath.ts
+++ b/src/juliaexepath.ts
@@ -139,20 +139,14 @@ export async function getJuliaExePath() {
         if (configPath === null) {
             for (const p of getSearchPaths()) {
                 try {
-                    const res = await exec(`"${p}" --startup-file=no --history-file=no -e "println(Sys.BINDIR)"`)
-                    if (p === 'julia' || p === 'julia.exe') {
-                        // use full path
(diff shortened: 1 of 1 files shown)


davidanthoff:
This doesn't work with juliaup, `which` won't find the necessary binary. I had spent a fair bit of time trying to get things to work properly with `which` at some point on Windows, but that never really worked reliably... I think we should just try to get the entire extension to work with a command line string rather than requiring an absolute path to the Julia executable.

davidanthoff:
I'll take a crack at getting this to work with just command line stuff.

pfitzseb:
But why would juliaup run into this at all? Is it the `julia.exe` case? If so, we could just completely get rid of any path resolution here, no?

davidanthoff:
Yes, that is what I'm trying to do right now on my machine: just don't try to resolve any paths at all. I'm about halfway through, I'd say :)

pfitzseb:
One consideration is that the terminal link resolver uses the result of this function to find files with relative paths (e.g. from stacktraces). Shouldn't hold up any of this though; we can find other solutions for that.

pfitzseb:
Closing in favor of #2341.

PR #2341, opened 2021-08-03
Rewrite Julia executable search code

### LINKED ISSUE #2294: Stop asking Julia to resolve Julia path

davidanthoff:
Fixes #2294. Alternative to #2340.

This got a bit out of control ;) But I think it is now in pretty good shape, hopefully. Couple comments:
- It interprets the config value as a command now, not a path. So things like `julia +1.2`, full paths, just `julia` etc. should all work.
- We no longer try to resolve the path (except for the link stuff) at all. Instead, we now use the command throughout to start everything. We probably need to test every feature that spawns Julia on every platform to be sure it all works... Or have this on insider sufficiently long :)
- Startup should be much faster because we now only run `julia -v` essentially, instead of running actual Julia code.
- Here is my logic for whitespaces in the configuration value: first, I just test if the entire value in the config points to a file. If yes, I take the entire content of the config field as `file` for `execFile`. If that is _not_ the case, I split the config value into args and the first one becomes the `file` part. I think that gets both the case of a value of say `C:\my folder\julia.exe` _and_ the case of `julia +1.2` right. What won't work is `C:\my folder\julia.exe +1.2`, for that case one would have to write `"C:\my folder\julia.exe" +1.2`. I think that makes sense?
- We were up to three different node packages that promisify `child_process`, I consolidated that to one that seemed best.
- I turned the whole Julia executable code into a class in the same way we've done it with other parts of the codebase as well when we made some significant changes, to get rid of global vars etc.

The entire thing is also much in preparation for an upcoming proper integration with `juliaup`, which should make the part where we search for notebook kernels even simpler.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -16,7 +16,7 @@ import { ProfilerResultsProvider } from './interactive/profiler'
 import * as repl from './interactive/repl'
 import { WorkspaceFeature } from './interactive/workspace'
 import * as jlpkgenv from './jlpkgenv'
-import * as juliaexepath from './juliaexepath'
+import { JuliaExecutablesFeature } from './juliaexepath'
 import { JuliaNotebookFeature } from './notebook/notebookFeature'
 import * as openpackagedirectory from './openpackagedirectory'
 import { JuliaPackageDevFeature } from './packagedevtools'
@@ -31,6 +31,7 @@ let g_languageClient: LanguageClient = null
 let g_context: vscode.ExtensionContext = null
 let g_watchedEnvironmentFile: string = null
 let g_startupNotification: vscode.StatusBarItem = null
+let g_juliaExecutablesFeature: JuliaExecutablesFeature = null
 
 export async function activate(context: vscode.ExtensionContext) {
     if (vscode.extensions.getExtension('julialang.language-julia') && vscode.extensions.getExtension('julialang.language-julia-insider')) {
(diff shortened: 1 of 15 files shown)


davidanthoff:
Oh, one other question is whether we should rename the configuration to `juliaCommand` or something like that... The current settings name has quite often actually led to confusion from users, so a clearer name might in general be good? Not sure, though... Maybe we only do this if we can somehow migrate config values over automatically, or something like that?

Another problem about the config setting is that on Linux we actually have no way at the moment to surface multiple different Julia versions as notebook kernels. So we could also rename the configuration to `juliaCommands`, and make it an array of strings, so that one can specify multiple Julia versions. By convention, the first one could then be the default one, or something like that... On the other hand, at that point it might be easier to just port `juliaup` to Linux and properly integrate that ;)

src/extension.ts
```diff
@@ -114,12 +116,13 @@ export async function activate(context: vscode.ExtensionContext) {
...
-            version: 2,
+            version: 3,
             async getEnvironment() {
                 return await jlpkgenv.getAbsEnvPath()
             },
-            async getJuliaPath() {
-                return await juliaexepath.getJuliaExePath()
+            // TODO This is breaking, not sure how to handle that?
+            async getJuliaExecutable() {
+                return await g_juliaExecutablesFeature.getActiveJuliaExecutableAsync()
```
davidanthoff, line 125: @pfitzseb not sure this makes sense? I increased the version number, and then this will return a `JuliaExecutable` now...

pfitzseb: Yeah, this is fine. And I don't think anyone but me is using this API anyways :P

pfitzseb (approved):
Looks great!

We can definitely implement config migration from `executablePath` to `startupCommands`. I can take a stab at that in a follow-up PR.

davidanthoff:
Uuups, shouldn't have merged this just now, as this is really for a minor bump. Sorry. On the other hand this would have been painful in terms of merge conflicts.

PR #2342, opened 2021-08-04
Fix some strict typing bugs

davidanthoff:
I'm going to tackle the entire strict typing mode for TypeScript, i.e. try to switch that on. For now this fixes a small number of problems discovered by that mode.

Once the new exe file search PR is merged, I'll try to convert the entire code base over to strict typing.

### DIFF
--- a/src/interactive/results.ts
+++ b/src/interactive/results.ts
@@ -427,7 +427,7 @@ function isResultInLineRange(editor: vscode.TextEditor, result: Result, range: v
 // goto frame utilties
 
 export async function openFile(path: string, line: number | undefined = undefined) {
-    const newLine = line ?? 1
+    const newLine = line || 1
     const start = new vscode.Position(newLine - 1, 0)
     const end = new vscode.Position(newLine - 1, 0)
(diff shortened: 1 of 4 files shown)


src/interactive/results.ts
```diff
@@ -422,10 +426,10 @@ function isResultInLineRange(editor: vscode.TextEditor, result: Result, range: v
 
 // goto frame utilties
 
-export async function openFile(path: string, line: number = undefined) {
-    line = line || 1
-    const start = new vscode.Position(line - 1, 0)
-    const end = new vscode.Position(line - 1, 0)
+export async function openFile(path: string, line: number | undefined = undefined) {
+    const newLine = line ?? 1
```
pfitzseb, line 430: Pretty sure `||` is more correct here, no? The result should always be at least 1, but `0 ?? 1` would return 0, right?

pfitzseb (approved):

src/interactive/results.ts
davidanthoff: Yep, didn't think of that!

PR #2346, opened 2021-08-05
Commands/button to toggle plot pane status

pfitzseb:
also show all plot pane related commands in the command palette

Fixes https://github.com/julia-vscode/julia-vscode/issues/2330.


package.json
```diff
@@ -417,32 +437,6 @@
...
-                },
-                {
-                    "when": "false",
-                    "command": "language-julia.plotpane-last"
-                },
-                {
-                    "when": "false",
-                    "command": "language-julia.plotpane-previous"
-                }
-            ],
```
davidanthoff, line 445: Obviously not part of this PR, but what happened to all of these commands? Or are they still here and this is just removing a condition?

davidanthoff (approved):

package.json
pfitzseb: Yeah, this just unhides them in the command pane.

PR #2350, opened 2021-08-07
Add test feature

davidanthoff:
~~Now somewhat functional.~~ This works quite well now, you can watch a demo at https://youtu.be/Okn_HKihWn8?t=1267 to get a sense how to use this.

The general idea is that one can add

```julia
@testitem "Name of test item" begin
end
```

anywhere in the Julia code, and then the test runner UI provides a way to execute each of these test items individually inside a new module. https://github.com/julia-vscode/TestItemRunner.jl has some documentation on how to write tests that can be picked up by this feature in the VS Code extension and how to run such new tests via the Base test infrastructure.

https://github.com/julia-vscode/TestItems.jl can be used to write inline tests in say a package.

Please do give this branch a try and report back whether it works!

Lots of things still to do:
- [x] Integrate with Revise
- [x] Provide some testset integration to collect test results
- [x] Handle process crashes
- [x] Create a simple package Testitems that contains just the `@testitem` macro
- [x] Figure out how to attach output to test runs
- [x] Think about tree structure for tests
- [x] Put the LS extension behind some capability detection story
- [x] Handle test environments (easy solution currently blocked by https://github.com/JuliaTesting/TestEnv.jl/issues/27)
- [x] Create a new package that can run these kind of tests outside of VS Code (see https://github.com/julia-vscode/TestItemRunner.jl and https://github.com/julia-vscode/TestItems.jl)


JackKenney:
Thanks for working on this @davidanthoff this is a much needed feature! Would be a huge improvement to my package-development workflow.

davidanthoff:
@pfitzseb this is ready for review.

I think what we should do is merge this relatively quickly but not announce it widely and then have the feature ship for a while "quietly" with just maybe the folks on the Slack channel knowing about it and trying it. We could also just say the whole thing is still in preview, but in the shipping extension. And then once things have settled, we can announce that it is out of preview.

But in general, I think this primarily needs to be tried out by folks now.

PR #2362, opened 2021-08-13
Fix image-like size to always fit pane

### LINKED ISSUE #2288: Plot scaling regression
If I'm not remembering it wrong, in this PR https://github.com/julia-vscode/julia-vscode/pull/1882 I changed the scaling of plots in the plot pane, so that they always fit into the available space, either vertically or horizontally. I don't know when this changed, but it appears that in the current version, the vertical scaling is broken, so that an image that is too high is clipped at the bottom, instead of scaled down.

grafik

jkrumbiegel:
Fixes #2288.

### DIFF
--- a/src/interactive/plots.ts
+++ b/src/interactive/plots.ts
@@ -288,9 +288,7 @@ function wrapImagelike(srcstring: string) {
     const html = `
     <html style="padding:0;margin:0;">
         <body style="padding:0;margin:0;">
-            <div style="max-width: 100%; max-height: 100vh;">
-                <img id="plot-element" style="max-height: 100%; max-width: 100%; object-fit: scale-down; object-position: 0 0;" src="${srcstring}">
-            </div>
+            <img id="plot-element" style="max-height: 100vh; max-width: 100vw; display:block; margin: auto;" src="${srcstring}">
         </body>
     </html>`
     return html


TonyLianLong:
Thanks for your contribution! It seems that only the code for plotting raw images are changed. Do you have any plans to change the code for other plotting libraries, too? I have personally tested this on my Mac and it works, and I believe it could be merged if there are no other things to modify at the same time for other types of images.

image
image

jkrumbiegel:
What do you mean with other plotting packages? This will work for any png, svg, jpg. For something like Vega plots, I don't think it's good to scale them as they are interactive documents

TonyLianLong:
> What do you mean with other plotting packages? This will work for any png, svg, jpg. For something like Vega plots, I don't think it's good to scale them as they are interactive documents

Sure. Makes sense.

TonyLianLong (approved):

davidanthoff:
Yeah, that sounds reasonable: vega-lite and vega plots normally have a fixed size and we should not scale them, presumably. Beyond that, I'll let @pfitzseb decide :)

jkrumbiegel:
@pfitzseb friendly bump :)

pfitzseb (approved):

Issue #2363, opened 2021-08-14
language server fails to start when using "~" alias for home directory


gszep:
~I'm using the [Remote SSH extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-ssh) to open a workspace on an ubuntu 18 server~. I've added julia to the remote environment
Edit: this problem has nothing to do with SSH
```bash
export PATH=$PATH:$HOME/.local/lib/julia/bin
```
and the `settings.json` reads
```json
{
    "julia.executablePath": "~/.local/lib/julia/bin/julia"
}
```
When I type `julia` into the remote terminal it opens a julia session with no problems. The language server however is stuck on the message `Starting Julia Language Server...` and whenever I try to execute an extension command --- for example Shift+Enter in a code block --- the log outputs the following javascript related error
```log
[2021-08-14 13:40:51.048] [exthost] [error] TypeError: Cannot read property 'file' of undefined
	at /home/gszep/.vscode-server/extensions/julialang.language-julia-1.3.28/dist/extension.js:2:679850
	at Generator.next (<anonymous>)
	at s (/home/gszep/.vscode-server/extensions/julialang.language-julia-1.3.28/dist/extension.js:2:677077) language-julia.executeCodeBlockOrSelectionAndMove
```
I have used these two extensions together before and it worked without issues. ~I suspect there is either some package version missing on this server, or that a recent update broke something.~ Would be great to track this down 🙏🏼

gszep:
Solved it. You are not allowed to use `~` as an alias for the home directory. Should this be made more clear in the docs?

pfitzseb:
We did resolve that before. Must've fallen under the table in a recent refactor.

PR #2364, opened 2021-08-14
Fix Julia exe search

davidanthoff:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2363.

I can't test this right now because I _also _broke my `juliaup` at the same time, so nothing is working here... If someone else could check this branch, please do!


pfitzseb (approved):

pfitzseb:
Ok, pushed a commit that incorporates https://github.com/julia-vscode/julia-vscode/pull/2365. All options I tried (command, abs path, path with `~`) work fine.

PR #2365, opened 2021-08-14
Use PATH for LS start

davidanthoff:
This might fix the other error we got reported on Slack.

The logic here would be that I think at the moment we might try to start the LS without a `PATH` being set at all. I don't fully understand why anything would have ever worked if that is the case, though...

Again, I can't test this right now because my juliaup is broken and I need to fix that first...


pfitzseb:
Closing in favor of #2364.

PR #2371, opened 2021-08-15
Use a hack to fix a tasks problem with juliaup

davidanthoff:
Fixes https://github.com/JuliaLang/juliaup/issues/125.

This is a short term hack and I'll try to find a better way to solve this. The problem is this: in general, it seems, that tasks in VS Code don't work with anything that is MSIX installed on Windows and has an execution alias. I tested for example `winget`, and trying to run that also doesn't work. So this is almost certainly an upstream bug in VS Code itself. For now this hack works around that. It breaks the scenario where our config setting is something like `julia +1.2`, i.e. where the config setting is not just pointing to one executable, but rather a combined command. But given that we didn't support that in the past and didn't announce yet that we (thought)  supported that now, I think this hack is OK for now.

### DIFF
--- a/src/juliaexepath.ts
+++ b/src/juliaexepath.ts
@@ -12,6 +12,7 @@ import { resolvePath } from './utils'
 
 export class JuliaExecutable {
     private _baseRootFolderPath: string | undefined
+    private _fullPath: string | undefined
 
     constructor(public version: string, public file: string, public args: string[], public arch: string | undefined, public channel: string | undefined, public officialChannel: boolean) {
     }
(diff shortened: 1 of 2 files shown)


pfitzseb:
>  It breaks the scenario where our config setting is something like julia +1.2, i.e. where the config setting is not just pointing to one executable, but rather a combined command.

Can we split on `' '` and put the `+1.2` into `args`? Should be an easy workaround for that case.

pfitzseb (approved):

davidanthoff:
> Can we split on `' '`

No, because of paths with spaces in it... I think the bug is fixed upstream, so we can revert this soon, hopefully.

PR #2374, opened 2021-08-16
Add juliaup integration

davidanthoff:
Needs https://github.com/JuliaLang/juliaup/pull/126. This will allow us to bypass this entire messy Julia finding code :)


davidanthoff:
The `juliaup` prereq shipped now, so this is ready for review and rollout. At the moment `juliaup` is broadly available on Windows and Archlinux, but we're working on all other platforms as well. We could for now disable this on other platforms, but given that there are already (somewhat complicated, admittedly) ways to run `juliaup` on all other platforms as well, I think it makes sense to just ship for all platforms at this point.

package.json
```diff
@@ -729,6 +729,11 @@
                     "description": "Points to the julia executable.",
                     "scope": "machine-overridable"
                 },
+                "julia.juliaupCommand": {
+                    "type": "string",
+                    "default": "juliaup",
+                    "scope": "machine"
```
pfitzseb, line 735: Can you add a description here? In particular, does this setting allow for version specifiers like `julia +1.6.4`?

davidanthoff: No, this was the path to the `juliaup.exe` binary. I actually removed it now, thinking more about it that is actually not something that should be configurable.

To use a different Julia, it should (in theory) be enough to just put `julia +1.6.4` into the standard exe path setting.

pfitzseb: Ah, makes sense.

pfitzseb:
Needs a rebase, I think.

davidanthoff:
@pfitzseb I think this is ready, if you approve I can merge?

pfitzseb (approved):

rapus95:
What's the current state of this; Open or merged? 🤔😁

davidanthoff:
And also shipping.

PR #2376, opened 2021-08-16
update plotly version to 2.3.1

### LINKED ISSUE #2375: Update plotly.js version
Hello!

The PlotlyBase.jl and PlotlyJS.jl packages have updated to plotly versions 2.X

From what I can tell the version bundled with the vs code extension (and referenced here: https://github.com/julia-vscode/julia-vscode/blob/1558812aa411c0b3a80832cec8f88677e55dce32/src/interactive/plots.ts#L549) is still the 1.X series.

This makes it so that many figures using new features introduced in the 2.X series do not work. 

What would be the best way for us to get that plotly version updated?

sglyon:
Fixes #2375


pfitzseb (approved):

pfitzseb:
Thanks!

davidanthoff:
@sglyon you could also integrate an automatic update [here](https://github.com/julia-vscode/julia-vscode/blob/master/src/scripts/updateDeps.ts#L27). That is what we do for vega-lite and vega and it makes sure we always ship an up-to-date version of those packages.

sglyon:
nice!

For now the cdn for plotly-latest.min.js points to the most recent 1.X release (not sure why??), so it isn't quite time. 

hopefully next time we need an update we'll be able to go the auto route

sglyon:
This is my first time contributing here... so I'm a bit unfamiliar with the release process.

Do you have any idea for when we could expect this to land in the released version of the extension?

davidanthoff:
We collect all PRs that have the "Next Minor" milestone until we open a merge window for the next minor release, and then we merge them all in one go. That way we can keep `master` open for now for bug fix patch releases.

I think I might have misclassified this, though, I'd be fine to ship this out in the next patch release. @pfitzseb agreed?

PR #2394, opened 2021-08-23
Revert #2371

davidanthoff:
The upstream bug is fixed. We just need to wait for a new VS Code version to ship, on insider it works already.


davidanthoff:
@pfitzseb This should be ready for review now. Would also be nice to ship this in a patch before we start the new minor merge window.

pfitzseb (approved):

PR #2372, opened 2021-08-15
split getDefaultEnvPath execFile args

novitk:
Fixes the bug in getting default env. The original code doesn't work on win or linux.

PR #2395, opened 2021-08-23
Fix arg passing to Julia

davidanthoff:
This continues the work started in https://github.com/julia-vscode/julia-vscode/pull/2372.

At least on my system things are _really_ broken without that. I don't understand why that is coming up now... Maybe it has something to do with `juliaup` and that that requires properly escaped arguments, not sure... In any case, the theory here is that `execFile` handles all subtleties of quoting and escaping for us. I hope that is right :)


davidanthoff:
@pfitzseb would be great if you could review after the fact, I now merged because I want to get an insider build with this included out ASAP.

pfitzseb:
LGTM.

Issue #2392, opened 2021-08-22
errors not shown in Julia REPL when PortAudioStream is open


petres:
I am quite new to Julia and VS Code and therefore I am not sure if this is a bug or belongs here in this bug tracker.

The Code:
```javascript
using PortAudio
undefined_var
stream = PortAudioStream("HDA Intel PCH: ALC3246 Analog (hw:0,0)", 2, 2)
undefined_var
close(stream)
undefined_var
```
Evaluation of this code yields in the pure Julia REPL:
```
julia> using PortAudio

julia> undefined_var
ERROR: UndefVarError: undefined_var not defined

julia> stream = PortAudioStream("HDA Intel PCH: ALC3246 Analog (hw:0,0)", 2, 2)
PortAudioStream{Float32}
  Samplerate: 44100.0Hz
  2 channel sink: "HDA Intel PCH: ALC3246 Analog (hw:0,0)"
  2 channel source: "HDA Intel PCH: ALC3246 Analog (hw:0,0)"

julia> undefined_var
ERROR: UndefVarError: undefined_var not defined

julia> close(stream)
0

julia> undefined_var
ERROR: UndefVarError: undefined_var not defined
```

In the Julia REPL of Julia VSC extension I get:
```
julia> using PortAudio

julia> undefined_var
ERROR: UndefVarError: undefined_var not defined

julia> stream = PortAudioStream("HDA Intel PCH: ALC3246 Analog (hw:0,0)", 2, 2)
PortAudioStream{Float32}
  Samplerate: 44100.0Hz
  2 channel sink: "HDA Intel PCH: ALC3246 Analog (hw:0,0)"
  2 channel source: "HDA Intel PCH: ALC3246 Analog (hw:0,0)"

julia> undefined_var

julia> close(stream)
0

julia> undefined_var
ERROR: UndefVarError: undefined_var not defined
```

As you can see, it seems that the `UndefVarError` is not printed in the VSC session, which makes it quite hard to work with it. `UndefVarError` is only an example, other errors are also not shown. 

Thank you for any hint to get rid of this behaviour.

Environment:

Julia v1.3.30

VSC
Version: 1.59.1
Commit: 3866c3553be8b268c8a7f8c0482c0c0177aa8bfa
Date: 2021-08-19T11:54:37.513Z
Electron: 13.1.7
Chrome: 91.0.4472.124
Node.js: 14.16.0
V8: 9.1.269.36-electron.0
[...]

PR #2396, opened 2021-08-23
print eval errors to stdout instead of stderr

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2392.

I'm not 100% this is more correct though.


davidanthoff:
Hm, I don't understand why that would fix it? And agreed, that generally does not seem ideal, it really should go to `stderr`, right?

davidanthoff:
Ah, now just read the discussion in the underlying issue :) I think that is a bug in the package and I think we should not do anything about it. In my mind error messages should only go to `stderr`, really.

PR #2415, opened 2021-08-30
Overhaul table viewer

pfitzseb:
- Updates ag-grid version to 26.0
- Switches to infinite scroll mode for >10k elements
- Uses Tables.jl if loaded but also works without it


pfitzseb:
image

vjd:
Thanks for this @pfitzseb . A few important nice features to have based on what is available in Rstudio

1. https://support.rstudio.com/hc/en-us/articles/205175388-Using-the-Data-Viewer-in-the-RStudio-IDE
2. Ability to pop out the data viewer to a full screen size to improve viewing experience for large datasets
3. Ability to click on a the dataframe object in the workspace to open up the data viewer.

pfitzseb:
Alright, this is the current look:
image

Features:
- You can open tables from the workspace or with `vscodedisplay()`. Small tables (with fewer than 100k elements) are displayed synchronously, bigger ones async. Filtering and sorting works for all small tables; filtering should work for all big tables, but support for sorting is spotty and therefore disabled at the moment.
- The table viewer supports `AbstractVecOrMat`s, `Tables.table`s (e.g. `DataFrame`s), and `IterableTables`.
- Selected rows can be copied into the clipboard as tab separated data (should be compatible with Excel etc).
- Header and row-column are pinned.
- Big tables are fast (filtering on a table with 1e8 elements is reasonably snappy at 1s for filter changes).
- The viewer holds on to a ref to the original table while the table tab is open. Julia can GC the table once the tab is closed.

vjd:
Wow. This is cool

davidanthoff:
@pfitzseb I started to look at the tables PR, very nice, that will be super useful!

There is one relatively large thing that I think we should do differently: I think we should make a copy of the table at the point when the initial display request comes in, store only a link to that in the table cache and then serve all subsequent filtered/sorted/slice requests from that cached value. The primary reason for that is that otherwise we can run into major correctness problems with mutable state: someone displays a `DataFrame` with a large number of elements, so we jump into the paged implementation. Now someone leaves that table tab open in the background for a long time, and starts to manipulate the `DataFrame` instance in the REPL. At that point the table viewer would be thoroughly corrupted and no longer work. I think that is such a common (and valid) scenario that we should work around it by just having very clear semantics of: the table viewer shows a snapshot of the data at the point in time when `vscodedisplay` is called. This has other benefits as well: we shouldn't iterate rows from the start every time someone requests a new slice of the table, as that can be a very expensive operation. That is in particular true for Query.jl results, as this can imply that a very complex query is run again and again every time someone scrolls in the table viewer.

I think we could add optimizations to that down the road: say if the source is a read-only table (like queryverse/DataTables.jl) we would obviously not need to make a copy. Or maybe at some point there will be some notion of observable/change notification for tables, then one could (maybe) consider changing the table viewer into something that provides a "live" view of a given table instance.

We also can't rely on `Tables.istable` as the indicator whether we can display something. That function can return false negatives, so a `false` value there doesn't mean that something is not a table.

Should I maybe open a PR on top of this here to address some of these parts? Some of this touches things that I'm very familiar with and concerns various queryverse internals, so maybe easier for me to do that?

davidanthoff:
Stupid question: does this change the schema of the JSON we are sending as `application/vnd.dataresource+json`? Right now it looks to me as if it does? That will break the integration with notebooks etc, that MIME type is not our own format but a standard.

And I think this also removes the support we currently have for tables that return `TableTraits.isiterabletable(...) === missing`, the code that handles those scenarios also all seems to be gone?

This PR in general feels a bit like a rewrite, instead of adding paging to the existing structure? There was a lot of (subtle, admittedly) logic to how things worked in the old version to make sure it works with all corner cases of Tables.jl and Queryverse, and that just seems to be gone now?

davidanthoff:
Just FYI, I'm working on a PR on top of this one here now.

pfitzseb:
I am very much against caching stuff for the async case, fwiw. We can instead increase the threshold for sync display.

> Stupid question: does this change the schema of the JSON we are sending as application/vnd.dataresource+json? Right now it looks to me as if it does? That will break the integration with notebooks etc, that MIME type is not our own format but a standard.

Ah, didn't know that.

> And I think this also removes the support we currently have for tables that return TableTraits.isiterabletable(...) === missing, the code that handles those scenarios also all seems to be gone?

Hm, probably. I thought that `Tables.istable` would be the more general way to check whether something is table-like and ideally we wouldn't need to worry about IterableTables and so on. But reverting that part is easy enough ;) Generally it's a bit annoying that we have these two subtly different notions of tables...

> Just FYI, I'm working on a PR on top of this one here now.

Feel free to work on this branch directly.

davidanthoff:
> I am very much against caching stuff for the async case, fwiw.

Why? I don't really see how we can get a) get a consistent user experience without that and b) not end up with really weird bugs for mutable tables without that. I think we need to have very simple and clear semantics: either viewing a table is a snapshot or it is a live view. What I think we should not have is that a table is a snapshot if it is small, but a sort-of-live view when it is large. That would just be too inconsistent, I think.

I also should say that doing a live view that actually works is a _really_ hard problem, IMHO. One where values update when one scrolls seems again like a very weird user experience.

There is another aspect of this: serving these data slices from scrolling is really a prime candidate to be done in a background thread, once we have those in Julia. The table viewer should not stop working in an ideal world just because I run some code in the REPL. Doing that on a thread in the background is almost trivial if we have a copy of the data at hand that we know no one else has a reference to, but almost impossible if all we have is a reference to some mutable state.

> We can instead increase the threshold for sync display.

That seems like an orthogonal issue?

> I thought that `Tables.istable` would be the more general way to check whether something is table-like

No, I don't think it has any real information value for our scenario, unfortunately... Because it has false negatives, it essentially doesn't tell us anything we could usefully use, I think.

> But reverting that part is easy enough ;)

Yes :) The bigger change is that we can't iterate from the beginning every time we get a new slice request, that will never work for sources that are more expensive to compute.

I think the big question in terms of design is this question whether we should create a copy in the async case of the data at the point when the user wants to display the data, doing that (or not doing it) really would change the design here.

pfitzseb:
Ok, there are three different domains we need to talk about:
1. Small tables, let's say with fewer than 1e6 elements (so 8MB if those elements are Int64)
2. Medium tables with 1e6 < #elements < 1e9
3. Big tables with >1e9 elements.

Small tables are displayed synchronously. Big tables need to be displayed asynchronously because they'll usually be too big to fit into RAM multiple times.
That leaves medium tables, for which caching does make sense, as you say. I don't think inconsistency is much of an argument though -- it's up to the user not to modify state that they are viewing or deal with the consequences if they do.

> I think the big question in terms of design is this question whether we should create a copy in the async case of the data at the point when the user wants to display the data, doing that (or not doing it) really would change the design here.

It wouldn't change much at all, would it? Currently we hold on to a reference to the table; we just need to replace that with a copy  (`collect` into a `Vector{NamedTuple}`? Not sure if it's even possible to generically collect iterable tables).

davidanthoff:
What about this: small tables the situation is clear, we leave as is now. We then treat everything above a threshold as a medium table (which makes a copy of the data for us), and then we add an `unsafe` option to `vscodedisplay` that doesn't make a copy and essentially means that the user is guaranteeing that the table won't be mutated, i.e. essentially we make your "big table" option opt-in explicitly by the user.

And then down the road we could add features that for example detect if a source is a read-only table and then jump into the big table implementation right away without the need for the user to specify an `unsafe` option, or say add support for something that is observable if that exists at some point etc.

I think in general we would then still need to limit the big table option to cases where the data is actually in memory already, i.e. where we can access individual rows quickly. I think for anything where the row iteration does actual work (say run a potentially complex query, or read a file during iteration, or read stuff from a DB as you iterate) the big table option really won't work.

Down the road I'll have to think a bit more how we could do the sorting, filtering and paging in a more Query.jl integrated way. In principle it should for example by quite straightforward to use the MySQL Query.jl backend to do those operations in the DB itself while we page, without any MySQL specific code in our extension. At least that was very much a design goal for Query.jl, so this would be a good test ;) But, all of that is future talk, obviously.

> Not sure if it's even possible to generically collect iterable tables

Very easy, `collect` will work on any iterable table and create a vector of named tuples with the rows (a TableTraits.jl table is just an iterator of named tuples). It would probably make more sense, though, to store the data in column format. In principle that code is in https://github.com/queryverse/TableTraitsUtils.jl, and it is also structured in such a way that it can be easily vendored, except there are some corner cases that make it a bit more tricky. But I'm working through these right now, don't think anything there will be a real hurdle.

pfitzseb:
Ok, this is much closer to the original implementation now and doesn't do anything Tables.jl-related at all. I did completely rewrite the JSON writing, but that should be fine. 

There are two things I'm a bit unsure about:
- `AbstractMatrix`: I'm storing an iterator over all rows (basically `eachrow`, but available on 1.0)
- Thresholds:
  - Small tables have fewer than 1e5 elements (so 800KiB if the table elements are 8 byte each)
  - Medium tables are cached (and we do in-place sorting of the cache; there's also a cache for the last filter operation) and between 1e5 and 1e8 elements (0.8MiB to 800MiB)
  - Big tables are everything bigger than that (and also everything we don't know the size of); neither filtering nor sorting are enabled for those.

vjd:
🎉

PR #2424, opened 2021-09-02
Save metadata in jupyter notebooks

davidanthoff:
I had hoped that the 1.60 release of VS Code would ship the necessary APIs to enable us to write correct Jupyter Notebook files, but I think that is still not the case... 

With this PR we can now save the correct `kernelspec` in the notebook file (by using https://github.com/microsoft/vscode/pull/131219), but I don't see how we can save the `metadata.language_info` in the file, which is also required for a spec conform notebook file.

So I think we are still at square one, i.e. that VS Code does not provide us with the necessary APIs to write spec conform Jupyter notebook files, which also means that we are _still_ blocked from enabling the notebook controller by default... I think at some point we probably just have to make some public announcement a la "sorry, yes, we said back in July this is ready, but for now we just have to pull this feature" or something like that, we are getting a lot of questions now why the feature is not on by default...

CCing the folks on Microsoft's side that had worked on this: @DonJayamanne, @claudiaregio, @mjbvz and @roblourens. Can you think of a way to solve this, or am I maybe just missing something obvious here?


pfitzseb:
We can also set `language_info` in the metadata directly, apparently. So I think this should be good to go as-is (unless modifying metadata like this is a Bad Idea).

src/notebook/notebookFeature.ts
```diff
@@ -202,19 +202,17 @@ export class JuliaNotebookFeature {
...
-        nbmetadata.custom.metadata.language_info = {
-            name: 'julia',
-            version: `${version}`,
-            mimetype: 'application/julia',
-            file_extension: '.jl'
+        if (notebook.metadata?.custom?.metadata?.language_info?.name) {
+            notebook.metadata.custom.metadata.language_info.name = 'julia'
+        }
+        if (this.vscodeIpynbApi) {
+            this.vscodeIpynbApi.setKernelSpec(notebook.uri, kernelspec)
```
DonJayamanne, line 214: @davidanthoff Yes you are right, this should have been `setMetadata` or `setNotebookMetadata`.
Apologies for the oversight.

davidanthoff:
> We can also set language_info in the metadata directly, apparently. So I think this should be good to go as-is (unless modifying metadata like this is a Bad Idea).

I think it is a Bad Idea :) I don't fully understand why, but that is how I understood @DonJayamanne so far. Metadata needs to be edited by creating a `WorkspaceEdit`, and then using a call to `replaceNotebookMetadata` (like in https://github.com/microsoft/vscode/pull/131219/files#diff-51bf718c615a1b91db875986ac7d65f1dde07d5056787529f002680e6639ef4fR32), but that latter function is only a proposed API so we don't have access to it. The idea for 1.60 had been that the inbox jupyter extension in VS Code, which does have access to proposed APIs, would export an API for us to use until `replaceNotebookMetadata` is released.

src/notebook/notebookFeature.ts
davidanthoff: @DonJayamanne is there any workaround that we could use, or do we have to wait for VS Code 1.61 for this to work?

pfitzseb:
> I think it is a Bad Idea

Ok, then the remaining question is whether it is a BAD IDEA (i.e. whether it can break things instead of just not working) :)

`notebook.metadata` **is** public afterall. Generally it of course makes sense to always use `TextEdit`s.

davidanthoff:
So I expanded @pfitzseb's commit a bit to save the full `language_info` we want, and it seems to work... So I think unless @DonJayamanne has some very strong reason why we shouldn't do this, I'm also tempted to merge this and ship. In particular if the alternative is that we have to wait another month for this to be resolved.

src/notebook/notebookFeature.ts
mjbvz: We wanted to create a limited api targeted just for this use case. My impression from the discussion in microsoft/vscode#131219 was setting the `kernelSpec` was all that was required to unblock you. If we added the ability to set language info too, would that fully address your use case then?

Do you want to try updating the api that VS Code exposes to take a `languageInfo` parameter and submit a PR if it looks good. The addition should be fairly straightforward. If you confirm that fixes the problem, we could ship the PR as part of a 1.60.1 build within a week or so

davidanthoff: > setting the kernelSpec was all that was required to unblock you

Well, anything that allows us to produce valid notebook files :) I should have looked more carefully at the PR to spot that it was missing the language info stuff.

> If we added the ability to set language info too, would that fully address your use case then?

Yes, I'm opening a PR now.

davidanthoff: @mjbvz, see https://github.com/microsoft/vscode/pull/132298.

DonJayamanne (approved):

pfitzseb (approved):

PR #2431, opened 2021-09-06
executeActiveFile runs editor content

pfitzseb:
and doesn't read from FS
Fixes https://github.com/julia-vscode/julia-vscode/pull/1728#issuecomment-913544737.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
         "onCommand:language-julia.stopREPL",
         "onCommand:language-julia.executeCodeBlockOrSelection",
         "onCommand:language-julia.executeFile",
+        "onCommand:language-julia.executeActiveFile",
         "onCommand:language-julia.toggleLinter",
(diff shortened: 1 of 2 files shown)


davidanthoff:
Hm, I don't understand this PR :) As far as I can tell it just introduces a new command, swaps the that in for the command that is currently in the drop down, and then the new command does exactly the same thing as the old one? That changes nothing, right? or am I missing something here?

pfitzseb:
It does. The "play" button in the editor gives a URI argument, so we always use the file on the FS instead of the current editor content. Basically
```
registerCommand('language-julia.executeFile', executeFile)
```
is
```
registerCommand('language-julia.executeFile', uri => executeFile(uri))
```
where `uri` is always truthy.

davidanthoff (approved):
Ah, got it!

PR #2434, opened 2021-09-06
Use tokens for timeout in dynamic completions

davidanthoff:
This rewrites the timeout functionality in terms of cancel tokens, which I think is a) more in line with the general VS Code patterns and b) will allow us to cancel even more aggressively once we support cancel tokens across the JSONRPC boundary.

The code for the two functions is copied from https://github.com/microsoft/vscode/blob/be7046f09199da197a2b566f90f51f7e2653fed7/src/vs/platform/remote/common/remoteAgentConnection.ts#L89, I don't fully understand why that isn't just part of the json-rpc npm package, but whatever :)

The one thing that is not entirely clear to me is whether we should just call `return` if cancelled, or return an empty result? The old code seemed to do both things.

### DIFF
--- a/src/interactive/completions.ts
+++ b/src/interactive/completions.ts
@@ -3,6 +3,7 @@
 import * as vscode from 'vscode'
 import * as rpc from 'vscode-jsonrpc'
 import { Disposable, MessageConnection } from 'vscode-jsonrpc'
+import { combineCancellationTokens, createTimeoutCancellation } from '../canceltokenutils'
 import { getModuleForEditor } from './modules'
 import { onExit, onInit } from './repl'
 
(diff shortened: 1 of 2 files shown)


src/interactive/completions.ts
```diff
@@ -36,42 +37,34 @@ const requestTypeGetCompletionItems = new rpc.RequestType<
...
+            const line = document.getText(lineRange)
+            const mod: string = await getModuleForEditor(document, position) // TODO pass newToken once we support tokens in JSONRPC.jl
+
+            if (newToken.isCancellationRequested) { return }
 
-            return Promise.race([
-                completionPromise,
-                cancelPromise
-            ])
+            const items = await conn.sendRequest(requestTypeGetCompletionItems, { line, mod }) // TODO pass newToken once we support tokens in JSONRPC.jl
```
pfitzseb, line 60: Doesn't this mean that we always wait for this to go through before returning, unless cancellation got requested *before* we dispatch this request?

davidanthoff: Ah, yes, you're right. I don't know actually what the semantics are if we were to pass the token to the `sendRequest` call: would this `await` here finish right away, irrespective of what is happening on the server? In that case we could just pass the token here, even if the server doesn't yet implement cancel notifications. Or would it send the cancel notification message to the server and then wait until the request returns?

pfitzseb: The latter, I think.

davidanthoff: Yep, seems more likely...

Issue #2474, opened 2021-09-27
Plotting pane slow at rendering SVG (the default MIME-type)


ma-laforge:
It appears that the SVG rendering facilities for VSCode's integrated plot pane is noticeably slow --- even when comparing to the inline plots rendered from the Jupyter extension.

The difference is obvious when using Plots.jl (for example) to plot over 1k points using markers (ex: scatterplot).

More details & sample code on discourse:
--> https://discourse.julialang.org/t/why-is-julias-graphics-system-so-slow/68750

axsk:
Same as https://github.com/julia-vscode/julia-vscode/issues/2070

Issue #2070, opened 2021-03-18
performance issue at Plots panel ?


yeruoforever:
I use my laptop to connect to a remote server. When I draw a chart, my laptop becomes slow to respond.
I tried to find the reason, and found that when using the julia plots panel, the cpu occupancy rate increased significantly, and the memory approached 4GB. It seems that this puts a burden on my laptop that shouldn't be borne, especially when I switch charts.
I am interested in Plots planel in julia-vscode. Is the chart rendered by the client's vscode? Does it re-render every time I switch the chart? Is there a better solution to make the user experience smoother?
```julia
julia> using Plots
julia> scatter(rand(100),markershape=:xcross)
julia> scatter(rand(1000),markershape=:xcross)
julia> scatter(rand(10000),markershape=:xcross)
julia> scatter(rand(100000),markershape=:xcross)
julia> scatter(rand(1000000),markershape=:xcross)
```

pfitzseb:
The issue here is that we default to outputting SVGs, which can indeed slow down or crash whatever application they're shown in. We could consider switching to PNG as the default.
In theory
```
scatter(rand(100),markershape=:xcross,html_output_format=:png)
``` 
should do the trick, but that code path isn't actually being hit here. In the meantime
```
julia> p = scatter(rand(100),markershape=:xcross);

julia> display("image/png", p)
```
should always work, even though it's a bit annoying.

yeruoforever:
> The issue here is that we default to outputting SVGs, which can indeed slow down or crash whatever application they're shown in. We could consider switching to PNG as the default.
> In theory
> 
> ```
> scatter(rand(100),markershape=:xcross,html_output_format=:png)
> ```
> 
> should do the trick, but that code path isn't actually being hit here. In the meantime
> 
> ```
> julia> p = scatter(rand(100),markershape=:xcross);
> 
> julia> display("image/png", p)
> ```
> 
> should always work, even though it's a bit annoying.

Thanks for your reply and following your suggestions, I can work happily on my laptop.

jcbritobr:
Im having a problem with svg plots in monte carlo simulations. Vscode plugins do not support that much large sizes and it will crash taking all memory of your system. I agree making png plots defaults, as we no need high resolutions for the most majority of plots. Or, we need to enhance plot panel performance in case to stay with svg.

![image](https://user-images.githubusercontent.com/6579862/137467797-2aaa521d-2157-4780-b92e-8f38418b1e96.png)

PR #2475, opened 2021-09-27
Prefer png plots over svg

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2474


axsk:
This change seems to have been reverted in https://github.com/julia-vscode/julia-vscode/commit/673d17eabe3df5e1548cfccd28e1253723cd7eb2.
Was this on purpose? I am now back to having the issue https://github.com/julia-vscode/julia-vscode/issues/2070

PR #2546, opened 2021-11-08
bugfix: correctly escape `"` characters when starting julia

haberdashPI:
Ran into an error when trying to run the newly released insiders. Turns out the current version doesn't properly replace `"` characters. (I was worried it was something to do with my changes in #2534, but it seems unrelated).


pfitzseb (approved):
Good catch, thanks!

PR #2570, opened 2021-11-18
Use pixelated rendering for images in the plot pane

cdsousa:
This will be closer to how VSCode renders images.

See the difference in the rendering of the current behavior:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to the Julia extension will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+### Changed
+* Use pixelated rendering mode for images in the plot pane, in the same way VS Code renders images. ([#2570](https://github.com/julia-vscode/julia-vscode/pull/2570))
 
 ## [1.5.5] - 2021-11-16
 ### Fixed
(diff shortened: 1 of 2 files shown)


pfitzseb:
> This will be closer to how VSCode renders images.

Closer or the same? Can you post a screenshot?

cdsousa:
> Closer or the same? Can you post a screenshot?

Glad you asked. 

In fact I discovered some differences.
For a 100x100 random pic, at the default zoom, they look the same:
image
Also for zoom in:
image
For zoom out, the rendering of the image by VS Code seems to do some filtering (probably to improve subsampling), whereas plot pane does not:
image

However, for larger images, while the rendering is pixelated the plot pane seems to be doing some pre-filtering no matter the zoom level. E.g., 2000x2000 random image zoom-in:
image
1080x1920 image zoom-in:
image

cdsousa:
That prefiltering is something that will hinder the usefulness of pixelization. So I think this PR is not ready as-is.

pfitzseb:
I'm pretty sure this is due to the fact that we're using `transform: matrix(...)` for pan/zoom. It's gonna be tricky to do without, so I'll just merge this as-is, because that's the best we can do for now.

pfitzseb (approved):

cdsousa:
This happens only for large images, and the sub-sampling does not seem to change with the zoom level. Do you know what/where is such subsampling happening?

pfitzseb:
Nope. As I said above, my best guess is the zoom implementation, which was the only difference to the image viewer I could spot.

cdsousa:
Just a note, the subsampling is happening when any of the vertical or horizontal sizes are greater than 1000.

Issue #1533, opened 2020-07-31
Expose the option to restart REPL


bkamins:
Now you have to `exit()` Julia and "Start REPL". It would be nice to have it as a single command in certain workflows.

@aviatesk suggested on Slack that this functionality already exists, but is just not exposed in UI.

davidanthoff:
Agreed that we should have that.

What I use right now is Ctrl+D to kill the REPL, or I click the little delete button above the REPL to kill it. Not great, but maybe faster than `exit()`.

xgdgsc:
I' m implementing this at https://github.com/xgdgsc/julia-vscode/pull/1/files . But it gives "terminal is already disposed" on start. Any clues?

pfitzseb:
Pretty sure you need to change `killREPL` to something like
```
    if (isConnected()) {
        g_connection.end()
        g_connection = undefined
    }
    if (g_terminal) {
        g_terminal.dispose()
        g_terminal = null
    }
```
But please just open a PR against this repo to discuss things like that.

PR #2720, opened 2022-02-07
remove time wasting dialog and restart REPL

xgdgsc:
remove time wasting dialog


pfitzseb:
Not a fan of this. Killing a tmux session that you may not even have open in VS Code and which may or may not contain an arbitrary number of other running processes seems like a recipe for disaster.

xgdgsc:
Maybe you should poll somewhere to find how people use this. For me the only process I would want in this tmux session is julia. I can' t imagine anyone would want to run other process inside the same session in the tiny horizontal terminal inside vscode rather than starting a terminal app and another tmux session elsewhere. I think the only difference between a persistent and non-persistent is the "persistent", other command behavior should be identical. Or you could make this only appear at first time to inform users and never appear again.

pfitzseb:
That's fair. Maybe just add an additional `Yes, always` choice?

xgdgsc:
Thanks. added https://github.com/julia-vscode/julia-vscode/issues/1533 also in this PR.

package.json
```diff
@@ -997,6 +1006,12 @@
                     "scope": "machine-overridable",
                     "markdownDescription": "Name of the `tmux` session."
                 },
+                "julia.persistentSession.stopWarning": {
+                    "type": "boolean",
+                    "default": true,
+                    "scope": "machine-overridable",
+                    "description": "Warning on closing of persistent session. false to disable warning"
+                },
```
pfitzseb, line 1014:
```suggestion
                "julia.persistentSession.warnOnKill": {
                    "type": "boolean",
                    "default": true,
                    "scope": "machine-overridable",
                    "description": "Warn when stopping a persistent session."
                },
```

src/interactive/repl.ts
```diff
@@ -38,6 +39,65 @@ function startREPLCommand() {
 
     startREPL(false, true)
 }
+async function checkKillPersistentSession() {
```
pfitzseb, line 42:
```suggestion
async function confirmKill() {
```

src/interactive/repl.ts
```diff
@@ -38,6 +39,65 @@ function startREPLCommand() {
...
+    if (persistentStopWarning || vscode.workspace.getConfiguration('julia').get<boolean>('persistentSession.stopWarning') === false) {
```
pfitzseb, line 43: Would be good to share the result of `vscode.workspace.getConfiguration('julia')` with the code below.

src/interactive/repl.ts
```diff
@@ -38,6 +39,65 @@ function startREPLCommand() {
...
+                    throw new Error(`tmux kill-session Process failed with exit code ${code}`)
+                }
+            }
+
+        } catch (err) {
+            vscode.window.showErrorMessage('Failed to close tmux session.')
+        }
+    }
+}
+async function stopREPL() {
```
pfitzseb, line 86: This should just be `killREPL` below, no? I don't see the need for an additional method.

src/interactive/repl.ts
```diff
@@ -38,6 +39,65 @@ function startREPLCommand() {
...
+            if (choice === agree || choice === agreeAlways) {
+                return true
+            }
+        }
+        finally {
+            persistentStopWarning = false
+        }
+    }
+}
+async function stopREPLInPersistent() {
```
pfitzseb, line 68: I think this can live in `killREPL` directly. If not, let's call it `maybeKillPersistentSession`; that name seems a bit clearer.

pfitzseb (changes requested):
Left a few comments inline, but this looks good overall.

src/interactive/repl.ts
```diff
@@ -28,6 +28,7 @@ let g_languageClient: vslc.LanguageClient = null
 let g_compiledProvider = null
 
 let g_terminal: vscode.Terminal = null
+let persistentStopWarning: boolean = false
```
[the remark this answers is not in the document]
xgdgsc: https://github.com/julia-vscode/julia-vscode/pull/2720/commits/0b1afa7ba078c19ac0979d5be7f787ffd5d7c5ab

pfitzseb (approved):
Looking good! Thanks for bearing with me :)

PR #2727, opened 2022-02-08
save and format before execute

xgdgsc:
As now when I save and format the file after getting results. The results at line end would disappear. What about save before execute?


pfitzseb (approved):

xgdgsc:
When would 1.6.0 release outside of insider?

Issue #2706, opened 2022-01-29
Add Option to Debug from Active REPL


jandaa:
Currently the debugger runs in a separate terminal, requiring the source code to be re-evaluated. This slows down the time between clicking debug and running your code. I'm seeing about a 10 to 15 second delay, which is manageable but would be great to bring this time down and allow faster development. It would also be great if we could leverage data which has already been loaded (i.e. large matrices or dataframes) instead of re-loading every time.

### Idea
Add option to run debugger from the active REPL. 

This is already somewhat possible by loading the Debugger.jl package at the top of the file, hitting `Julia: Execute active File in REPL`, and using @run your_function() in the REPL terminal window. Any breakpoints that are set in vs-code will be hit and you can run through your code as a normal vs-code debug session. The advantage is that your breakpoints are hit instantaneously instead of waiting for a new terminal window to load.

I'm not sure if this is possible, but seeing as there is already this work around it should be in theory?

jandaa:
Here's an example. Using the following sample Julia script: 

```julia
using Revise
using Debugger

function main()
    a = 1
    b = 2

    # Put breakpoint here
    c = a + b
end

main()
```
Then:

1. Put a breakpoint at `c` under the comment
2. In the top right run icon, click `Julia: Execute active File in REPL` 
3. Run a normal debug session and wait to hit your breakpoint (should take about 10 seconds)
4. Then, in the active REPL, run the command `@run main()`. You should hit your breakpoint instantly

pfitzseb:
Just don't load Debugger.jl, set a normal breakpoint and use `@run main()` (or `@enter main()`) and you're good to go.

PR #2733, opened 2022-02-10
Feat/debug run cell

xgdgsc:
I find the workflow at https://github.com/JuliaDebug/Debugger.jl#breakpoint-on-error pretty neat as you can use `bt` to show all values at stack. Might be useful for https://github.com/julia-vscode/julia-vscode/issues/2706


src/interactive/repl.ts
```diff
@@ -755,6 +755,34 @@ async function executeCell(shouldMove: boolean = false) {
...
+    const code = doc.getText(cellrange)
+
+    await startREPL(true, false)
+
+    if (shouldMove && ed.selection === selection) {
+        const isJmd = doc.languageId === 'juliamarkdown'
+        const nextpos = new vscode.Position(nextCellBorder(doc, cellrange.end.line + 1, true, isJmd) + 1, 0)
+        validateMoveAndReveal(ed, nextpos, nextpos)
+    }
+    const debugCode = 'using Debugger\nbreak_on(:error)\nDebugger.@run begin\n'+code+'\nend'
```
pfitzseb, line 782: I'm confused. Why would we load Debugger instead of using the built-in Debugger here?

xgdgsc: Don' t know how to do that. calling `    debuggerRun({code: code,filename: ed.document.fileName})` gives me errors

PR #2791, opened 2022-03-20
Add source map support for exceptions

davidanthoff:
I think this should add proper stack traces for exceptions in the extension. It is quite hacky, in particular it will override (potentially not just for our extension) the default VS Code exception stack trace reformatting from https://github.com/microsoft/vscode/blob/32b031eeefc4fd27a21659d35070967bfe965bcc/src/vs/workbench/api/common/extensionHostMain.ts#L92...

Really, MS should ship an implementation of `prepareStackTrace` that handles source maps for extensions properly.

But maybe worth trying this? Given that we have a fair number of unhandled exceptions in the extension, with essentially no way to know what they are...


davidanthoff:
Just FYI: https://github.com/microsoft/vscode/issues/145473

pfitzseb:
> it will override (potentially not just for our extension) the default VS Code exception stack trace reformatting

sounds kinda sketchy. although it seems unlikely to be any worse than the non-source-mapped default, right?

davidanthoff:
> although it seems unlikely to be any worse than the non-source-mapped default, right?

Yeah, that is what I think. I'm also not even sure whether the extension host is shared between different extensions? I think we should try it, and even if we just have it shipping for a few days for everyone, we will collect a ton of super useful data. If it does cause problems, we can still revert?

pfitzseb (approved):

Issue #2484, opened 2021-09-29
No output when running JET in REPL


jakobnissen:
As discussed on Slack, running `@report_call foo()` in the REPL in VSCode will make the JET errors show up in VSCode's linter. Very cool!

But it was also confusing for me, as I'm used to seeing the JET in the REPL. To be clear, I prefer having it in the editor, but it just confused me.

Maybe it would be nice to print a message in the REPL whenever the user does `@report_call`, or another of the JET macros integrated with VSCode.

pfitzseb:
cc @aviatesk

aviatesk:
woa, we already have users of this hidden integration !

You can use [`vscode_console_output`](https://aviatesk.github.io/JET.jl/dev/config/#JET.VSCode.VSCodeConfig) configuration to get REPL printing also, like:
```julia
julia> @report_opt vscode_console_output=stdout sum("julia")
```

And now I wonder `vscode_console_output` might be better to specify `stdout` by default ?

pfitzseb:
We could also fix this here and call `display(REPLDisplay()/InlineDisplay(), result)` when the diagnostics mime type is hit.

aviatesk:
Yeah, it would be simpler if we can fix this on VSCode side and always enable REPL printing.

pfitzseb:
Fixed by https://github.com/julia-vscode/julia-vscode/pull/2536.

jakobnissen:
@pfitzseb is this actually resolved? I just ran JET, with no output in the REPL

claforte:
FWIW I also spent a couple of hours debugging this before figuring out this was a vscode+julialang issue... it would be great if this were fixed...

tpgillam:
> woa, we already have users of this hidden integration !
> 
> You can use [`vscode_console_output`](https://aviatesk.github.io/JET.jl/dev/config/#JET.VSCode.VSCodeConfig) configuration to get REPL printing also, like:
> 
> ```julia
> julia> @report_opt vscode_console_output=stdout sum("julia")
> ```
> 
> And now I wonder `vscode_console_output` might be better to specify `stdout` by default ?

I'm not sure if this is a regression from https://github.com/julia-vscode/julia-vscode/pull/2797 , or is the "Julia: Execute Code in REPL" action intended to have different behaviour?

[...]

PR #2793, opened 2022-03-21
Display diagnostic MIME type in REPL

BenPH:
JET output shows in the REPL with this.

Is this the right idea? I'm not quite clear on when `display` or `show` and family is appropriate.

Fixes https://github.com/julia-vscode/julia-vscode/issues/2484.


pfitzseb:
I think the approach in https://github.com/julia-vscode/julia-vscode/pull/2797 is slightly better in that it will not display the diagnostics twice, as would be the case with this PR in some configurations. Should also be `display(MIME"text/plain"(), diagnostics)`.

pfitzseb:
I think the approach in https://github.com/julia-vscode/julia-vscode/pull/2797 is slightly better in that it will not display the diagnostics twice, as would be the case with this PR in some configurations. Should also be `display(MIME"text/plain"(), diagnostics)`.

BenPH:
Ok thanks I thought there might be more to it. I'll close this one.

PR #2796, opened 2022-03-21
Improve juliaup integration with LS

davidanthoff:
This makes things work when a user configures something like `julia +lts` as their exe. Right now that is broken because we start the LS with a special depot, and then juliaup tries to install julia versions into that special folder, which we don't want to happen.

### DIFF
--- a/src/juliaexepath.ts
+++ b/src/juliaexepath.ts
@@ -282,7 +282,7 @@ export class JuliaExecutablesFeature {
         return this.actualJuliaExePath
     }
 
-    public getUsingJuliaup() {
+    public isJuliaup() {
         return this.usingJuliaup
     }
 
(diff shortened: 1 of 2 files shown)


src/juliaexepath.ts
```diff
@@ -277,6 +282,10 @@ export class JuliaExecutablesFeature {
         return this.actualJuliaExePath
     }
 
+    public getUsingJuliaup() {
```
pfitzseb, line 285:
```suggestion
    public isJuliaup() {
```
reads a bit nicer, but also doesn't matter ;)

davidanthoff: BUT, when one has a really reliable rename functionality, it is also a very easy change, so I renamed ;)

Issue #2447, opened 2021-09-13
FR: Integrate project / package development


mkborregaard:
The functionality I miss the most in Julia is to have creation of new projects or packages be easier - you often have to straddle multiple platforms, use github directly, import PkgTemplates etc etc before you can start coding. We are currently holding a uni course in Julia and it is clear that this is a stumbling block for people coming from R.

The dream would be an ability to create a project directly from a context menu in VSCode, that would initialize a package (using e.g. PkgTemplates) and even set up git etc, automatically. Here is a small recording of the current functionality in RStudio: 

https://user-images.githubusercontent.com/8429802/133039518-5b155923-bbff-4cc5-af63-fa6740ccfeb4.mov

This is a port of a similar issue I opened on Juno 3 years ago https://github.com/JunoLab/Juno.jl/issues/175

PR #2799, opened 2022-03-23
New project creation command

BenPH:
This adds a `Julia: Create New Project` command that takes you through a series of prompts to get the fields needed to create a new project using PkgTemplates. This is a similar experience to calling `PkgTemplates.Template(interactive=true)`.

Any suggestions are appreciated!

Ref https://github.com/julia-vscode/julia-vscode/issues/2447


davidanthoff:
Very nice!

We handle Julia env/packages in two ways. For some things (that we consider really core) we ship all required packages as submodules, and then use a Julia environment where all non stdlib packages are deved. For those env, we never need to instantiate and there is never a need to download anything. For other things, we use more standard env, and then just instantiate them at the beginning of the Julia script, https://github.com/julia-vscode/julia-vscode/blob/master/scripts/packagedev/tagnewpackageversion.jl is an example of that.

I think this PR here introduces a mixed case, where PkgTemplates.jl is a submodule, but the env that is used still has things that we don't ship and therefore it needs to be instantiated before use. I think this should fall into either category, though. Given how many deps this whole thing has, I'm probably in favor of not doing this via the submodule route, but instead just put everything into the manifest like a normal package, and then make sure that at the beginning of the Julia script the project gets instantiated. Does that make sense?

One other thing is that we already have some limited support for the package butler stuff in the extension, and it would be nice if we would have a new project story that is coherent with that... I'm not entirely sure at the moment what the best way to do that would be... One option is probably to just add support for package butler controlled packages to PkgTemplates.jl, but I'll need to think about it a bit more.

src/createproject.ts
```diff
@@ -0,0 +1,157 @@
...
+
+        let host = await vscode.window.showQuickPick(['github.com', 'gitlab.com', 'bitbucket.org', 'Other'], { placeHolder: 'The URL to the code hosting service where the project will reside.' })
+        if (host === 'Other') {
+            host = await vscode.window.showInputBox({ prompt: 'Please enter the URL to the code hosting service.' })
+        }
+        if (host === undefined)
+            return
+
+        // TODO: add gihub.user as placeholder
+        const user = await vscode.window.showInputBox({
```
davidanthoff, line 37: Can we also get that from the built-in VS Code authentication stuff?

src/createproject.ts
```diff
@@ -0,0 +1,157 @@
...
+            validateInput: (input) => {
+                if (input === "")
+                    return 'Username cannot be empty.'
+                return undefined
+            }
+        })
+        if (user === undefined)
+            return
+
+        let juliaVersion = await vscode.window.showQuickPick(
```
davidanthoff, line 48: Can we have a good default here?

src/createproject.ts
```diff
@@ -0,0 +1,157 @@
...
+                { label: 'ColPracBadge', picked: false },
+                { label: 'Coveralls', picked: false },
+                // { label: 'Develop', picked: false },
+                { label: 'Documenter', picked: false },
+                { label: 'DroneCI', picked: false },
+                { label: 'GitHubActions', picked: false },
+                { label: 'GitLabCI', picked: false },
+                { label: 'PkgEvalBadge', picked: false },
+                { label: 'RegisterAction', picked: false },
+                { label: 'TravisCI', picked: false }
```
davidanthoff, line 82: Is there a way to provide more descriptions? This is probably pretty hard to understand for a newcomer.

davidanthoff:
One other thought: might it make sense to have a "short" and simple default UI path here? Say it first asks for the name of the project, and then just whether one wants to configure advanced options. If a user picks the advanced stuff, all the other questions come up, but otherwise it just goes with good defaults?

src/createproject.ts
BenPH: I found that you can access the global git config through vscode's git api but it requires a repository to exist already which doesn't make a lot of sense here.
```ts
const extension = vscode.extensions.getExtension<GitExtension>('vscode.git')
if (extension == null && extension.isActive) {
    const gitExtension = extension.exports
    const api = gitExtension?.getAPI(1)
    const user = await api.repositories[0].getGlobalConfig('github.user')
}
```
There's [simple-git](https://www.npmjs.com/package/simple-git) that does the job. Does it make sense to add a dependency?

BenPH:
> Given how many deps this whole thing has, I'm probably in favor of not doing this via the submodule route, but instead just put everything into the manifest like a normal package, and then make sure that at the beginning of the Julia script the project gets instantiated. Does that make sense?

That makes sense. Thanks for clearing that up.

> One other thought: might it make sense to have a "short" and simple default UI path here? Say it first asks for the name of the project, and then just whether one wants to configure advanced options. If a user picks the advanced stuff, all the other questions come up, but otherwise it just goes with good defaults?

Good idea! If you call `Template()` with no kwargs it gives a pretty suitable default. I'll put together some changes.

src/createproject.ts
BenPH: 1.6? Or the user's current version? If it's left blank PkgTemplates will use whatever version that runs the script. That could be the default.

BenPH:
@davidanthoff Could you have another look when you get the chance? I've cleaned it up with your suggestions and am pretty happy with it now.

PR #2801, opened 2022-03-25
Fix various REPL bugs

davidanthoff:
I think this should all help with various crash reports we have.

### DIFF
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -284,6 +284,7 @@ function disconnectREPL() {
     } else {
         if (isConnected()) {
             g_connection.end()
+            g_connection = undefined
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/interactive/repl.ts
```diff
@@ -900,9 +915,11 @@ async function evaluate(editor: vscode.TextEditor, range: vscode.Range, text: st
             }
             r.setContent(results.resultContent(' ' + result.inline + ' ', result.all, Boolean(result.stackframe)))
         }
+
+        return true
     } catch (err) {
         r.remove(true)
-        telemetry.handleNewCrashReportFromException(err, 'Extension')
+        throw(err)
```
davidanthoff, line 922: I think this function is only being called from functions that already have the error handler set up, we should in general probably only ever send crash reports from the most outer exception handler.

PR #2805, opened 2022-03-26
Add more startupt output to notebook server

davidanthoff:
Hopefully this will allow us to figure out what is going on with https://discourse.julialang.org/t/notebook-cells-hung-up-in-vs-code/78492.


davidanthoff:
@pfitzseb I'm merging as I want to get this out and I think this should be entirely harmless.

Issue #2806, opened 2022-03-27
persistent session isn' t persistent on remote after local suspend


xgdgsc:
In previous versions after a night' s sleep of my computer and me, the tmux persistent session is restored on calling Start REPL. But now it always starts a new session and the previous tmux process is gone even after a simple window reload. I guess it' s because the top 3 commits here: https://github.com/julia-vscode/julia-vscode/commits/master/src/interactive/repl.ts

xgdgsc:
https://github.com/julia-vscode/julia-vscode/commit/6dba59f3bad2e84877b5145fc695540b33744b90 most suspicious.

xgdgsc:
I started seeing this again with 
v1.6.24 after suspend

has anything changed?

PR #2807, opened 2022-03-27
Fix/persistent killed on reload

xgdgsc:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2806


src/interactive/repl.ts
```diff
@@ -58,17 +58,19 @@ async function confirmKill() {
         return false
     }
 }
-async function stopREPL() {
+async function stopREPL(onDeactivate=false) {
     const config = vscode.workspace.getConfiguration('julia')
     if (Boolean(config.get('persistentSession.enabled'))) {
```
pfitzseb, line 63:
```suggestion
    if (Boolean(config.get('persistentSession.enabled')) && !onDeactivate) {
```
seems cleaner.

pfitzseb (approved):

PR #2815, opened 2022-03-29
'Go to Definition' for workspace items

pfitzseb:
image

This adds a 'Go to definition' button for modules and methods; it also adds all methods as children to their function.


davidanthoff (approved):
Very nice!

PR #2830, opened 2022-04-04
Fix a readonly problem

davidanthoff:
This might fix a crash report. It seems that accessing the `fsPath` file tries to manipulate the uri instance, but it is read only or something like that...


pfitzseb (approved):
Weird...

davidanthoff:
Yeah, I'm not actually entirely sure this will fix it, but lets hope ;)

PR #2831, opened 2022-04-04
Fix a case where missing introduces 3VL

davidanthoff:
I think technically we would actually need to special case handle _any_ case anywhere in our code base where we take any value from a user provided function and turn it into a boolean via some function that one would expect to return  a bool... But this at least fixes one case that showed up in crash reporting.

### DIFF
--- a/scripts/packages/VSCodeServer/src/trees.jl
+++ b/scripts/packages/VSCodeServer/src/trees.jl
@@ -277,7 +277,7 @@ function getvariables()
         !isdefined(M, n) && continue
         Base.isdeprecated(M, n) && continue
 
-        x = getfield(M, n)
+        x = coalesce(getfield(M, n), nothing)
         x in (
             vscodedisplay,
             VSCodeServer,
--- a/scripts/packages/VSCodeServer/src/trees.jl
+++ b/scripts/packages/VSCodeServer/src/trees.jl
@@ -277,8 +277,8 @@ function getvariables()
         !isdefined(M, n) && continue
         Base.isdeprecated(M, n) && continue
 
-        x = coalesce(getfield(M, n), nothing)
-        x in (
+        x = getfield(M, n)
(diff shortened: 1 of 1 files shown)


scripts/packages/VSCodeServer/src/trees.jl
```diff
@@ -277,7 +277,7 @@ function getvariables()
         !isdefined(M, n) && continue
         Base.isdeprecated(M, n) && continue
 
-        x = getfield(M, n)
+        x = coalesce(getfield(M, n), nothing)
         x in (
```
pfitzseb, line 281:
```suggestion
        x = getfield(M, n)
        x !== missing && x in (
```
is the correct fix here, no? Otherwise we show people's `missing` fields as `nothing` instead.

pfitzseb (changes requested):

scripts/packages/VSCodeServer/src/trees.jl
davidanthoff: Yep!

pfitzseb (approved):

KristofferC:
Classic `missing`.

pfitzseb:
Yeah... had to stop myself from ranting about `missing` multiple times today. Also fun that the docs only mention the 3VL logic in the prose instead of the "return type" annotation
```
help?> in
  in(item, collection) -> Bool
  ∈(item, collection) -> Bool

  Determine whether an item is in the given collection, in the sense that it is == to one of the values generated by iterating
  over the collection. Returns a Bool value, except if item is missing or collection contains missing but not item, in which
  case missing is returned (three-valued logic (https://en.wikipedia.org/wiki/Three-valued_logic), matching the behavior of
  any and ==).
  ```

davidanthoff:
> had to stop myself from ranting about `missing` multiple times today

Issue #2832, opened 2022-04-05
PyReverseDims crash in remote vscode


liupgd:
I use julia + remote vscode. The remote vscode run on a ubuntu system, native operating system is win10. 
# Code:
```julia
using PyCall
a_jl = ones(UInt8, (128, 128, 3));
a_np = PyReverseDims(a_jl);  # crashed
```
# Running Steps
* Use `Ctrl + Enter` to run julia codes and the REPL appears
* Run the code line by line. Crashed happened until PyReverseDims invoked.
* If the REPL is started manually in the terminal directly, that means julia REPL not started by `Ctrl+ENTER`, then this crash won't happen. 
* Run the same code with `Ctrl+Enter` in native win10, no crash happened. 
# Crash info:
The julia REPL quit immediately. I can only get the screen shot:
![image](https://user-images.githubusercontent.com/3350539/161770760-18ad6112-30a5-4fbe-a820-9a7da475e40a.png)
# version infos
* versioninfo()
> Julia Version 1.6.5
Commit 9058264a69 (2021-12-19 12:30 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Xeon(R) CPU E5-2678 v3 @ 2.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, haswell)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 

Crash happend with julia-1.7.0 also.
* vscode
![image](https://user-images.githubusercontent.com/3350539/161773096-ace49d3d-444e-48b4-ac12-1403c13bc6d8.png)

* package versions:
    - vscode remote development: 0.21.0 (newest)
    - PyCall package: v1.93.1(newest)
    - julia-vsocde plugin: 1.6.15(newest)

PR #2833, opened 2022-04-05
Fix getvariables check

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2832.


davidanthoff (approved):
Isn't that already fixed by https://github.com/julia-vscode/julia-vscode/pull/2831?

pfitzseb:
Nope. This fix isn't for missing but for `PyObject`s and is generally more robust.

Issue #2853, opened 2022-04-21
REPL hanging with nested progress logging in Transducers


sethaxen:
I was using Transducer's [example of nested progress logging](https://juliafolds.github.io/Transducers.jl/v0.4/reference/manual/#Transducers.withprogress) in the VS Code REPL, and it caused the REPL to hang and need to be stopped when ProgressLoading was also loaded. (it's fine in the standard REPL)

Here's the code:
```julia
julia> using Transducers

julia> xf = opcompose(
                    MapCat() do x
                        withprogress(1:x; interval=1e-3)  # nested progress
                    end,
                    Map() do x
                        sleep(0.5)
                        x
                    end,
                );

julia> foldxt(+, xf, withprogress(1:10; interval=1e-3); basesize=1)  # fine, no progress bars in VS Code

julia> using ProgressLogging

julia> foldxt(+, xf, withprogress(1:10; interval=1e-3); basesize=1)  # progress bars, but the process hangs
```

And here's a GIF of the behavior:
![Peek 2022-04-21 17-05](https://user-images.githubusercontent.com/8673634/164489280-84790e38-1929-4d81-b08b-444865633c81.gif)

## Version info

This is on v1.6.17 of the julia vs code extension, v1.63.2 of vs code, v0.4.73 of Transducers, and v0.1.4 of ProgressLogging, also:

```julia
julia> versioninfo()
Julia Version 1.7.2
Commit bf53498635 (2022-02-06 15:21 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: AMD Ryzen 9 5900X 12-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, znver3)
```

sethaxen:
> Interestingly enough this depends on the Julia version -- 1.6 works fine.

Yes, you're right, it works on 1.6. It also fails on the 1.8 beta.

PR #2854, opened 2022-04-22
Lock progress-notification

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2853.

### DIFF
--- a/scripts/packages/VSCodeServer/src/progress.jl
+++ b/scripts/packages/VSCodeServer/src/progress.jl
@@ -1,10 +1,18 @@
 struct VSCodeLogger <: Logging.AbstractLogger end
 
+const logger_lock = ReentrantLock()
 function Logging.handle_message(j::VSCodeLogger, level, message, _module,
     group, id, file, line; kwargs...)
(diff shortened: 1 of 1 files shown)


scripts/packages/VSCodeServer/src/progress.jl
```diff
@@ -1,10 +1,18 @@
...
     group, id, file, line; kwargs...)
     isprogress = try_process_progress(level, message, _module, group, id, file, line; kwargs...) do progress
-        JSONRPC.send_notification(conn_endpoint[], "repl/updateProgress", progress)
-        JSONRPC.flush(conn_endpoint[])
+        lock(logger_lock)
+        try
+            JSONRPC.send_notification(conn_endpoint[], "repl/updateProgress", progress)
+            JSONRPC.flush(conn_endpoint[])
+        catch err
+            # it's fine to drop this
```
davidanthoff, line 12: Hm, why would we want to drop errors here? That seems quite risky?

pfitzseb: Don't think so, but I've removed the catch block.

Issue #2849, opened 2022-04-15
Print versioninfo() on startREPL


xgdgsc:
Everytime I run start REPL I would wonder if I' m using the correct version or whether I updated julia successfully. And it seems to be common practice when starting such REPL in other language IDEs from my memory.

pfitzseb:
You can add `--banner=yes` to the `julia.additionalArgs` setting. We may end up changing the default in https://github.com/julia-vscode/julia-vscode/pull/2855.

PR #2855, opened 2022-04-22
Print banner by default

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2849.

Not sure if we want to change the default for this, since `julia.additionaArgs` can be used to change the behaviour.


davidanthoff:
Hm, I think I'm not too keen on this... Isn't there some new UI elements for showing more language specific information, maybe we could utilize that to show which Julia version is being used?

pfitzseb:
-> https://github.com/julia-vscode/julia-vscode/pull/2857

Issue #2802, opened 2022-03-25
REPL eval error from crash reporting


davidanthoff:
We have crash reports with this error message:
> The handler for the 'repl/runcode' request returned a value of type VSCodeServer.EvalErrorStack, which is not a valid return type according to the request definition.

This means that our handler inside the LS for that request is not always wrapping the return value in a `ReplRunCodeRequestReturn`. I tried to understand where that might happen, but the whole REPL execution framework is too complicated for me to understand right now :) @pfitzseb, this is probably for you.

xgdgsc:
During development of https://github.com/julia-vscode/julia-vscode/pull/3123 , I found when https://github.com/julia-vscode/julia-vscode/pull/3123/files#diff-78600594f1b69fa4e2f8213a0a3cf1aa6ebdaf7006b729337ea5e8c649ef11bdR261 anything inside `savePersistentResults` goes wrong it would return EvalErrorStack. So there could be some unhandled  error around.

PR #2856, opened 2022-04-22
Speculative fixes for repl_runcode_request

pfitzseb:
Will hopefully help debugging https://github.com/julia-vscode/julia-vscode/issues/2802.

### DIFF
--- a/scripts/packages/VSCodeServer/src/eval.jl
+++ b/scripts/packages/VSCodeServer/src/eval.jl
@@ -97,8 +97,8 @@ function add_code_to_repl_history(code)
 end
 
 ans = nothing
-function repl_runcode_request(conn, params::ReplRunCodeRequestParams)
-    return run_with_backend() do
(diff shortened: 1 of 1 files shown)


scripts/packages/VSCodeServer/src/eval.jl
```diff
@@ -117,10 +117,12 @@ function repl_runcode_request(conn, params::ReplRunCodeRequestParams)
         show_code = params.showCodeInREPL
         show_result = params.showResultInREPL
         show_error = params.showErrorInREPL
+        try
+            JSONRPC.send_notification(conn_endpoint[], "repl/starteval", nothing)
+        catch err
+            @debug "Could not send repl/starteval notification."
```
davidanthoff, line 123: These won't show up in crash reporting, right?

davidanthoff (approved):

scripts/packages/VSCodeServer/src/eval.jl
pfitzseb: Yes. It's really unlikely we'd get anything useful out of this even if it did though.

PR #2864, opened 2022-04-27
Feat/show inline all results

xgdgsc:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2507

For every PR, please check the following:
- [ ] End-user documentation check. If this PR requires end-user documentation, please add that at https://github.com/julia-vscode/docs.
- [ ] Changelog mention. If this PR should be mentioned in the CHANGELOG, please edit the CHANGELOG.md file in this PR.

PR #2866, opened 2022-04-27
feat/inline all results

pfitzseb:
Supersedes https://github.com/julia-vscode/julia-vscode/pull/2864.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1066,11 +1066,11 @@
                     "scope": "window",
                     "markdownDescription": "Save file before execution"
                 },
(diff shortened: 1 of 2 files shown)


xgdgsc:
What about make it default to true?

pfitzseb:
Changing the default behaviour feels much more breaking to users, so I'm hesitant to do that; especially since we don't really have a good way of communicating these kinds of changes at the moment.

xgdgsc:
I think removing something is breaking, adding some results display is not very breaking.

PR #2887, opened 2022-05-10
Workspace: Add toggle to hide modules

pfitzseb:
image

Ideally this would use a toggleable menu entry, but those aren't available to extensions atm (ref https://github.com/microsoft/vscode/issues/148633).


davidanthoff (approved):
Nice!

PR #2903, opened 2022-05-25
WIP: Live-share integration

### LINKED ISSUE #2895: Plots don't show up in Live Share
Plots don't show up for Live Share participants, it seems. Not sure if there's the Julia extension can do something about this, or if we have to wait for improvements in Live Share itself, though.

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2895.


pfitzseb:
Peek 2022-05-30 15-38

The extension is now live-share enabled, so this basic functionality works just fine now. Should probably take a look at properly synchronizing the plot pane, but maybe this is good enough for now.

PR #2923, opened 2022-06-15
Re-add check for valid Julia path

pfitzseb:
We lost that check in one of the path finding rewrites.

image


webpack.config.js
```diff
@@ -20,6 +20,8 @@ const config = {
     externals: {
         vscode: 'commonjs vscode', // the vscode-module is created on-the-fly and must be excluded. Add other modules that cannot be webpack'ed, 📖 -> https://webpack.js.org/configuration/externals/
         'applicationinsights-native-metrics': 'applicationinsights-native-metrics',
+        '@opentelemetry/instrumentation': '@opentelemetry/instrumentation',
+        '@azure/opentelemetry-instrumentation-azure-sdk': '@azure/opentelemetry-instrumentation-azure-sdk'
```
pfitzseb, line 24: This is unrelated and just gets rid of two errors/warnings when building.

davidanthoff (approved):

PR #2933, opened 2022-06-22
Use markdown show method for inline display popup

pfitzseb:
image (15)


korsbo:
Would it be possible to extend this to support and have `MIME"text/latex"` rendered on mouse hover? That could be super cool for LaTeXStrings, Latexify, Symbolics, ModelingToolkit, etc.

pfitzseb:
No. We can't inject anything to make latex render properly. You can of course render svgs and display them in the markdown :)

davidanthoff (approved):

PR #2935, opened 2022-06-25
Fix type for default in `julia.deleteJuliaCovFiles`

krassowski:
Tiny change - the type of `julia.deleteJuliaCovFiles`` is specified as boolean, but default was assigned to string.

- [x] End-user documentation check. If this PR requires end-user documentation, please add that at https://github.com/julia-vscode/docs. - not required
- [x] Changelog mention. If this PR should be mentioned in the CHANGELOG, please edit the CHANGELOG.md file in this PR. - not required

### DIFF
--- a/package.json
+++ b/package.json
@@ -1150,7 +1150,7 @@
                 },
                 "julia.deleteJuliaCovFiles": {
                     "type": "boolean",
-                    "default": "true",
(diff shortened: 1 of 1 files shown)


pfitzseb (approved):
Thanks!

PR #2955, opened 2022-07-15
No shell integration on windows

pfitzseb:
Shell integration is broken on Windows, so let's disable it for now.


pfitzseb:
May or may not be related to https://github.com/microsoft/vscode/issues/155315.

PR #2971, opened 2022-07-26
Added missing parentheses in "when" clauses of keybindings.

### LINKED ISSUE #2970: "Escape" keybinding interferes with the vim extension for .jmd and .md files
Trying to edit a .jmd file I noticed that vscode was ignoring `ESC` when trying to return to NORMAL mode from INSERT mode.

I'm pretty sure I've tracked the issue down to missing parentheses in a couple of the entries in the "keybindings" section of `package.json`.

I'll make a simple PR that I think should work, but I don't believe I have the setup in place to test it.

I can confirm, though, that the `language-julia.clearCurrentInlineResult` command is what is being triggered every time I press `ESC`.

twadleigh:
Fixes #2970. (Maybe. Untested.)

The scope of this PR extends a bit beyond #2970. It motivated the fix to the keybinding for `Escape`, but in the process revealed what is likely a similar error for other keybindings.


pfitzseb (approved):

pfitzseb:
This looks correct to me. I think you can just delete the second keybinding, FWIW.

twadleigh:
> This looks correct to me. I think you can just delete the second keybinding, FWIW.

Done.

PR #2975, opened 2022-07-27
Consistent context key logic and naming in when clauses

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2970.

### DIFF
--- a/package.json
+++ b/package.json
@@ -532,22 +532,22 @@
                 },
                 {
                     "command": "language-julia.clearAllInlineResultsInEditor",
-                    "when": "editorLangId =~ /^julia|juliamarkdown|markdown$/"
+                    "when": "editorLangId in julia.supportedLanguageIds"
                 },
                 {
                     "command": "language-julia.cdHere",
                     "group": "julia",
-                    "when": "editorLangId =~ /^julia|juliamarkdown|markdown$/"
+                    "when": "editorLangId in julia.supportedLanguageIds"
                 },
                 {
                     "command": "language-julia.activateHere",
                     "group": "julia",
-                    "when": "editorLangId =~ /^julia|juliamarkdown|markdown$/"
+                    "when": "editorLangId in julia.supportedLanguageIds"
                 },
                 {
                     "command": "language-julia.activateFromDir",
                     "group": "julia",
-                    "when": "editorLangId =~ /^julia|juliamarkdown|markdown$/"
+                    "when": "editorLangId in julia.supportedLanguageIds"
                 }
             ],
             "editor/title/run": [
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -749,17 +749,17 @@
             {
                 "command": "language-julia.clearCurrentInlineResult",
                 "key": "Escape",
-                "when": "editorTextFocus && !editorHasSelection && !findWidgetVisible && !markerNavigationVisible && !parameterHintsVisible && !inSnippetMode && !isInEmbeddedEditor && !suggestWidgetVisible && !onTypeRenameInputVisible && !renameInputVisible && juliaHasInlineResult && (editorLangId == julia || editorLangId == juliamarkdown || editorLangId == markdown)"
```
twadleigh, line 752: It makes sense that `markdown` can be safely included here because of the julia-specific guard `juliaHasInlineResult`, but I'm not sure it's a good idea to include `markdown` for the other keybindings. (Some included it before, some didn't.)

Best would be to restrict the binding to only when we are inside a proper julia code fence in the case of `markdown`. I recall from messing with syntaxes a few years ago that there is a way to get julia-specific code highlighting in markdown for julia code fences. I wonder if there is an analogous mechanism for the when clause.

pfitzseb: Unfortunately there's no context key for the injected grammar, so that's not an option.

All other keybindings already included markdown before; the only change here is that certain menu entries will now also appear in markdown files. Imho that's a reasonable behaviour.

twadleigh:
I can verify this branch fixes #2970.

package.json
davidanthoff: We could try to detect ourselves whether we are in a Julia code block and then use another custom context for that, I guess? But lets merge this as is for now.

davidanthoff (approved):

twadleigh:
Hmm, this branch may break the bindings for inline execution. I don't know why. I noticed yesterday that inline execution wasn't working (using a locally built vsix from this branch) for `juliamarkdown` and `markdown`, but I didn't have time to look deeper into it, so I moved on. Today, though, I'm having issues even in a regular `.jl` file. I just noticed it, and haven't yet investigated what's going on, but wanted to make a report here ASAP. Will follow up if I find learn anything new.

twadleigh:
I tried again, this time making sure I did the build in a completely clean repo. I'm seeing the same results.

twadleigh:
OK, now I'm _really_ confused. I uninstalled my locally built version and reinstalled the usual version. I'm seeing the same results. I checked all of my `settings.json` files, and they don't have anything unusual in them.
Is it possible that uninstalling a package via the GUI doesn't actually remove it from the cache, so that when I attempt to do a reinstall of the package with the same version number, it doesn't do a new download...so that I've actually been using the same broken version of the package all the while?

twadleigh:
OK, yes. Apparently I have to manually purge the `.vscode*/extensions` directory. Having done so I can report:
- The vsix built from this branch has broken keybindings for inline execution.
- The keybindings are fine for me with the current official release.

pfitzseb:
Care to try again? This should work fine now -- turns out I forgot to check in a file.

pfitzseb:
I'll just merge this because master being broken is annoying.

twadleigh:
OK, I can report that `master` (with this branch merged) has working code execution keybindings for `.jl` and `.jld` files, while also not interfering with `ESC` for the vim extension.

@pfitzseb, thanks for your effort in making those fixes.

o314:
Too much work to test pre-release vscode on my machine. I will alert if needed later
But the in op should do the fix since parenthesis are broken in when clause.
Thanks for the quick fix anyway.

PR #2985, opened 2022-07-30
Use one Julia project per Julia minor version

davidanthoff:
Ideally we would merge this by first running the old update script on `master`, then merge this one here and then run this new update script.


pfitzseb (approved):
LGTM.

PR #2986, opened 2022-07-31
simplify and optimze sorting a table view

LilithHafner:
When a user toggles a sort button on a table view, one semantic is to prepend that to the comparison function. i.e. sort by the most recently clicked sort breaking ties with less recently clicked sorts. This is equivalent to the semantic "when a user clicks a sort button, perform a stable sort according to that column". The latter phrasing lends itself better to eager sorting.

This can 
1) sort more quickly by using a simpler comparison function
2) sort more simply by storing history implicitly in the order of the table rather than explicitly

Somewhat orthogonally, this also fixes a minor issue where a filter change results in unsorted data being displayed despite having a sort selected.

It is probably possible to simplify `sortModel` because it now only needs to store the latest sort, but I couldn't find out where it is constructed. All I found was the declaration of `GetTableDataRequest`.

This PR is incompatible with lazy sorting via `partialsort`ing only what is displayed, but I don't think the perofrmance of partialsort is good enough to warrant doing that anyway.

### DIFF
--- a/scripts/packages/VSCodeServer/src/tables/filtering.jl
+++ b/scripts/packages/VSCodeServer/src/tables/filtering.jl
@@ -4,31 +4,12 @@ const SORTED_CACHE = Dict{UUID, UInt64}()
 col_access(row, col) = row[col]
 
 function generate_sorter(params, fixed_col_names)
-    lts = []
     for sortspec in params
-        let col = sortspec["colId"]
-            ind = findfirst(==(col), fixed_col_names)
-            if ind === nothing
-                continue
-            end
-            if sortspec["sort"] == "asc"
-                push!(lts, (r -> col_access(r, ind), (r1, r2) -> col_access(r1, ind) < col_access(r2, ind)))
-            else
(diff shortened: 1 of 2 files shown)


pfitzseb:
Makes sense to me. Not sure why I didn't do that in the first place...

> It is probably possible to simplify sortModel because it now only needs to store the latest sort, but I couldn't find out where it is constructed.

That's generated by ag-grid, so no, we can't really simplify that (nor would we really gain anything from that).

pfitzseb (approved):

LilithHafner:
Oops! I misunderstood how sortModel works. Thanks for the link to ag-grid. If one clicks a column header it sorts by that column, and then shift clicking another column header adds it as a lower priority key to sort by. This PR does not handle that correctly, and I don't see how to fix it.

lmk if you see a way forward here, but until then I don't think this approach is viable because we cannot assume anything about the relationship between the sortModel of one query and that of the previous queries.

pfitzseb:
Ah, that explains the "Not sure why I didn't do that in the first place..." part then. Thanks for double-checking!

PR #3032, opened 2022-08-28
Always use shell when we exec julia on Unix

davidanthoff:
I think whenever we start a new Julia process, we actually should use a shell on Linux/Mac because there might be some startup script that actually modifies the `PATH` that is needed.

My understanding is that `createTerminal` does this automatically, so in a way we are just mimic that behavior here. It might be that `createTerminal` also uses a shell on Windows, but not sure. But in any case, it seems to me that on Windows we generally don't want to add that extra overhead.

### DIFF
--- a/src/jlpkgenv.ts
+++ b/src/jlpkgenv.ts
@@ -82,7 +82,10 @@ export async function switchEnvToPath(envpath: string, notifyLS: boolean) {
                     println(false)
                 end`,
                 `${case_adjusted}`
-            ]
+            ],
+            {
+                shell: process.platform !== 'win32',
+            }
         )
 
         if (res.stdout.toString().trim() === 'false') {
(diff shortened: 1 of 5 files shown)


pfitzseb:
Mh, not particularly a fan of this because people should just properly set up their system. Since that might be a bit unrealistic, we should presumably just go with this...

`createTerminal` definitely does not go through a shell as it's currently used, so we'll need to adjust that as well (which may very well turn out to be a bit of a nightmare, because we need to handle escaping for ~arbitrary shells properly...).

davidanthoff:
Ah, you are right, I had misread the docs for `createTerminal`. Yeah, so let's not merge this here then.

The thing that is really confusing to me is that it seems we have systems where the LS and the REPL start (both of course launched not by us directly but rather by some VS Code internal code) and at the same time when we try to launch Julia for say the test process it seems to fail. That is how I am interpreting https://github.com/julia-vscode/TestItemRunner.jl/issues/19... If that is so, then there seems to be some difference in how we launch things relative to how say `createTerminal` launches things...

PR #3038, opened 2022-08-30
Typo fix in the Package Server setting description

mbauman:
`s/Set's/Sets` — don't need an apostrophe there.


pfitzseb (approved):
Thanks!

PR #3074, opened 2022-09-16
add zh cn translation

### LINKED ISSUE #2041: Add support for localization
Do we want to add support for localization.
With Jupyter, Python and some other extensions we just rely on the community to add support for location of other languges and the like.

Basically we just create a file named package.nls.json and most of the strings are hardcoded in there instead of having the text hardcoded in package.json.
E.g. the text `"title": "Julia: Restart Language Server"` would be hardcoded in another file.
And if we have a file named package.nls.zh-cn.json then based on locale VSCode will get the strings from tehre and merge into package.json.

As for strings added into the code base, that required a little bit more work, but once its setup, users can contribute to the commands.

We find this useful in the other extensions, and 99% of them are maintained by the community.

xgdgsc:
Fixes #2041 .


pfitzseb:
Very cool!
Is there a writeup of this somewhere in the VS Code docs? I'm kinda curious about how localizing other strings works.

davidanthoff:
Very nice! Two questions:
1. Can the translation files be moved into a sub folder? Root is already pretty busy and this has the potential to add a lot more files, right?
2. What happens if we add a new command and don't provide a translation for a given language? Will it automatically fall back to English? Primarily I'm wondering what the process would look for someone who wants to add a new feature but is not familiar with some of the languages that are supported in the extension. Can someone like that just add something in English and everything will work?

I think the final question is how we would maintain this long term. @xgdgsc do you plan to keep updating this and tracking new additions to the extension and then translating them?

xgdgsc:
1. https://github.com/microsoft/vscode-extension-samples/tree/main/i18n-sample found an sample of subfolder.
2. It would auto fallback to English.
3. I would like to maintain the zh cn translation in long term.

xgdgsc:
I couldn' t get the subfolder working because I didn' t find how ISO 639-3 works for simplified chinese.

xgdgsc:
Found an example https://github.com/microsoft/vscode-cmake-tools/tree/main/i18n , which requires gulp. But the translation doesn' t appear on my computer.

xgdgsc:
Found the bug in cmake tools repo that cause zh-cn translation not showing. Is it necessary to use the subfolder? I find that approach an additional complexity.

pfitzseb:
I don't mind these being in `/`. Let's merge this, @davidanthoff?

PR #3081, opened 2022-09-19
fix bug in extension.ts

kendonB:
Fixes bug where the user is pointed to the vscode environmentPath setting when I believe they should be sent to executablePath.

Not sure if this needs to be in the changelog. Can add if needed.


davidanthoff (approved):
This looks correct to me. @pfitzseb if you also agree, do you want to merge?

PR #3112, opened 2022-10-11
Implement `err` global

### LINKED ISSUE #3030: Support `err` global
Recently (for 1.8) Julia added an `err` global at the REPL to store the most recent exception not thrown from top-level. VSCode should incorporate it into its own REPL.

See the function containing this line:
https://github.com/JuliaLang/julia/blob/6cdc4ce6d3fe194965e269a9a7dc7ea500087115/stdlib/REPL/src/REPL.jl#L291

BioTurboNick:
Fixes #3030.

- [x] End-user documentation check. If this PR requires end-user documentation, please add that at https://github.com/julia-vscode/docs. (not necessary)

I don't know how to debug the extension personally, but this should be pretty close.


scripts/packages/VSCodeServer/src/eval.jl
```diff
@@ -174,13 +175,30 @@ function repl_runcode_request(conn, params::ReplRunCodeRequestParams)::ReplRunCo
...
+                        try
+                            istrivial = @static if isdefined(Base, :istrivialerror)
+                                Base.istrivialerror(errs.stack)
+                            else
+                                true
+                            end
+                            @static if @isdefined setglobal!
+                                istrivial || setglobal!(Main, :err, errs)
+                            else
+                                istrivial || ccall(:jl_set_global, Cvoid, (Any, Any, Any), Main, :err, errs)
```
pfitzseb, line 195: This probably shouldn't set `err` to the VSCodeServer specific error types, right? What's the Base behaviour for this?

BioTurboNick: It contains an `ExceptionStack`. So possibly it should be set to the contents of an EvalErrorStack?

scripts/packages/VSCodeServer/src/eval.jl
```diff
@@ -174,13 +175,30 @@ function repl_runcode_request(conn, params::ReplRunCodeRequestParams)::ReplRunCo
...
+                            istrivial = @static if isdefined(Base, :istrivialerror)
+                                Base.istrivialerror(errs.stack)
+                            else
+                                true
+                            end
+                            @static if @isdefined setglobal!
+                                istrivial || setglobal!(Main, :err, errs.stack)
+                            else
+                                istrivial || ccall(:jl_set_global, Cvoid, (Any, Any, Any), Main, :err, errs.stack)
+                            end
```
pfitzseb, line 196: Wanna put in some additional safe guards here like `errs isa EvalErrorStack`? I realize that's  never going to be false because `istrivialerror` was introduced later, but still :)

BioTurboNick: Let me know if that change makes sense - since the design is "try once, if it throws an exception that's okay, we just won't try anymore", throwing an error seemed easiest.

BioTurboNick:
Bump :-)

BioTurboNick:
@davidanthoff Maybe you can help?

pfitzseb (approved):

PR #3119, opened 2022-10-15
Make some more adjustments for new prerelease mech

davidanthoff:


davidanthoff:
@pfitzseb I'm going to merge now because I want to get an insider build out, but would be great if you would review after the fact.

PR #3122, opened 2022-10-16
Fix more errors in get module

davidanthoff:
Requires https://github.com/julia-vscode/LanguageServer.jl/pull/1176.

We are still getting crashes inside `getModuleForEditor` in crash reporting, apparently the check before the `sendRequest` whether the client is running didn't work properly. So this changes strategy to correctly use error codes to identify if the client wasn't running, which in general is a better way to handle this in an async world in any case. I also removed some nested try catch blocks that I didn't think make much sense and probably didn't work in any case as they did things like compare error message strings.

Easiest to review commit by commit, as this also adds additional error handling with crash report sending.

### DIFF
--- a/src/interactive/completions.ts
+++ b/src/interactive/completions.ts
@@ -3,6 +3,8 @@
 import * as vscode from 'vscode'
 import * as rpc from 'vscode-jsonrpc'
 import { Disposable, MessageConnection } from 'vscode-jsonrpc'
+import { handleNewCrashReportFromException } from '../telemetry'
+import { wrapCrashReporting } from '../utils'
 import { getModuleForEditor } from './modules'
 import { onExit, onInit } from './repl'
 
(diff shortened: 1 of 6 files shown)


src/interactive/completions.ts
```diff
@@ -43,19 +45,25 @@ function completionItemProvider(conn: MessageConnection): vscode.CompletionItemP
                 return
             }
             const completionPromise = (async () => {
```
pfitzseb, line 47: Can't we write this as
```
const completionPromise = wrapCrashReporting(async () => {})()
```

pfitzseb (commented):
Generally LGTM. Left one comment inline.

src/interactive/completions.ts
davidanthoff: No, `wrapCrashReporting` only works with non-`async` functions.

But I am realizing that in general we need two versions of all these error hooking functions, one async one non async. I'll fix that generically in a new PR.

pfitzseb (approved):

Issue #2705, opened 2022-01-29
show calculation results at endline on persistent session restore


xgdgsc:
Now on reconnect to a persistent session, the results shown at line endings are not shown. While the workspace of variables could be restored. Is it possible to show results at endline also from this workspace variables, if exists expressions like `Var=` and `Var` in workspace. Alghough it may not be 100% accurate as previous session (because of running order differences might not be saved for now to recover previous state), It  could be adequate for many applications.  Or you could just save the results variable name and matching line somewhere for later restore.

pfitzseb:
Definitely possible, but imho not worth the additional complexity.

Issue #2503, opened 2021-10-20
Feature Request: Link plots to code line that made them


oscarrutt:
The Julia quasi-notebook feature is fantastic. However, it lacks a valuable feature of native notebooks--plots become divorced from the line of code that generated them. In a notebook, the plot is directly below the line or cell that made it. In the julia-vscode case, all of the plots live in a single tab and you can't tell which plot belongs to which line of code. The request is to make a link back from each plot to the generating line of code. This could momentarily highlight the line while the eye follows the action from one tab to the other, or the highlight could be persistent, until another plot is linked back. This could be made to happen automatically or upon some user interaction such as clicking something--Command-clicking the plot, for example (Mac key convention used here) and/or a menu item in the plot pane. There will be instances where the user does not want to have the line of code displayed, perhaps because she/he is editing some other line and doesn't want the code pane to scroll. [Edit--adding to original post] Of course the linking would also work with plots displayed in the Plot Navigator.

Also, the extension provides tiny icons at the end of some lines of code for various things. To complement the above suggestion, make a tiny icon next to a line that generates a plot; clicking the icon takes you to the appropriate plot in the plot tab. [Edit--adding to original post] An additional use of the tiny icon would be a hover action whereby hovering on the icon causes a rendering of the plot to appear. The rendered plot might be a  miniature and might leverage off of the recent addition of the Plot Navigator which also uses miniatures.

pfitzseb:
Good ideas, although not trivial to implement.

oscarrutt:
> Good ideas, although not trivial to implement.

True, I suppose. But speaking from complete ignorance here, you can click on an error line in the terminal and it jumps you to the offending line of code. Maybe the line number could be embedded with the plot somehow.

PR #3123, opened 2022-10-17
Feat/persist inline results

xgdgsc:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2705 

 
I' ve encountered too many performance issues with jupyterlab and decided to do it with vscode.

### DIFF
--- a/scripts/packages/VSCodeServer/src/VSCodeServer.jl
+++ b/scripts/packages/VSCodeServer/src/VSCodeServer.jl
@@ -120,6 +120,7 @@ function serve(args...; is_dev=false, crashreporting_pipename::Union{AbstractStr
         msg_dispatcher = JSONRPC.MsgDispatcher()
 
         msg_dispatcher[repl_runcode_request_type] = repl_runcode_request
+        msg_dispatcher[repl_loadpersist_request_type] = repl_loadpersist_request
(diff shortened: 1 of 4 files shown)


davidanthoff:
Hm, I have to admit that I'm not really convinced that we should have something like this in the extension.

For one, this seems a very ad-hoc system to persist Julia session state, that to me seems feature creep for the extension. In particular, it seems to me that while a generic session persistence feature might be nice, it should be implemented outside of the extension. I also just remember _a lot_ of attempts to address this in some systematic way in the past in packages, none of the really worked out long-term, which suggests to me that this is a fairly tricky problem to solve well.

A second thing is that the saving of location information seems very brittle? A simple edit of the source file with some other editor would screw everything up, right?

And third, I'm also not really convinced that storing session state next to source files is a good idea...

Sorry to be so negative, but this is a pretty large-scale feature, and we are generally careful before we add something like this. Hope you understand! I'm also curious to hear what @pfitzseb thinks about this.

xgdgsc:
Implementation details could be discussed. I just need this to do my work for very long analysis notebooks where jupyterlab becomes unusable slow.

xgdgsc:
Added cleanup on start of loading persistent when code hash not matching. Added plot matching selection as fix #2503 . Added persistent of plots.

PR #3138, opened 2022-10-26
Fix unresolved promises

davidanthoff:
We had a gazillion unresolved promises in the codebase where we essentially didn't have the necessary `await` in the code. I now enabled some eslint rules to identify those situations, and then fixed them.

We can probably expect a bunch of new crash reports from this, previously we called a lot of functions in a fire-and-forget mode where any exception just disappears into the nothing :)


src/utils.ts
```diff
@@ -64,12 +64,24 @@ export function inferJuliaNumThreads(): string {
  * Same as `vscode.commands.registerCommand`, but with added middleware.
  * Currently sends any uncaught errors in the command to crash reporting.
  */
-export function registerCommand(cmd: string, f) {
-    const fWrapped = (...args) => {
+export function registerAsyncCommand<T1 extends unknown[],T2>(cmd: string, f: (...arg: T1) => Promise<T2>) {
```
pfitzseb, line 67: Thoughts on only using one function here which explicitly async-ifies everything?

davidanthoff: I also added an eslint rule that errors if a function is async when it isn't necessary. I guess we could remove that rule, but on the other hand it does seem good to not introduce unnecessary async functions?

PR #3139, opened 2022-10-26
bugfix: eval order

### LINKED ISSUE #3128: evaluation order guarantee
Usually after REPL restart I would rerun a lot of cells. But sometimes it happens the later cell are evaluated first , making a `Var undefined` error. 

I suspect it could be https://github.com/julia-vscode/julia-vscode/blob/588368a640e3fef4b8263c975fa8601d03c82d6e/src/interactive/repl.ts#L811 in this while loop,  code in next cell could send the message  to Julia backend before lower code inside previous cell. Is it the case? Should a queue be added?

xgdgsc:
Fixes #3128

Added a queue to ensure evaluation order on execution of multiple cells and cell-selection mixed execution of code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1495,6 +1495,7 @@
         "applicationinsights": "^2",
         "async-file": "^v2.0.2",
         "await-notify": "^1.0.1",
+        "fastq": "^1.13.0",
         "markdown-it": "^12.3.2",
         "markdown-it-footnote": "^3.0.3",
         "promised-temp": "^v0.1.0",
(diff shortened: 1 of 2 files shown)


Leena8686 (approved):

xgdgsc:
What' s the plan on this? Do you really use your extension to do anything serious? How can you not encounter this not at all during one day?

xgdgsc:
That only advances when you only run one cell and wait. But after a session crash like the just closed Endpoint issue, users would run multiple cells at once from top to bottom and wait for all data loading and transformations, etc, which is a habit from using jupyterlab notebooks. In this case when previous cell isn' t finished, the following cell already start to execute without a queue.

Issue #3153, opened 2022-11-12
Show unicode information on hover


dylanxyz:
Whenever i read other people's code and there is some unicode character that i want to know how to type it, i just copy and paste it in the REPL *help* mode. For example, if i want to know how to type the ⬲ character:

```julia-repl
help?> ⬲
"⬲" can be typed by \leftarrowonoplus<tab>
```

I think it would be useful to see this information when hovering over an unicode character instead of going in the process of copying the character → opening the REPL → typing `?` → pasting the character → pressing enter.

simonp0420:
The current behavior when hovering over a variable name is to display the line where the variable is assigned a value.  I like the current behavior and would not be in favor of eliminating it in favor of this proposal.  Perhaps there is some way to do both, which I would be fine with.

PR #3157, opened 2022-11-18
add cmd copy howto type

xgdgsc:
Add the command to copy how to type selected latex/emoji char to clipboard.

Fixes https://github.com/julia-vscode/julia-vscode/issues/3153

### DIFF
--- a/package.json
+++ b/package.json
@@ -436,6 +436,10 @@
             {
                 "command": "language-julia.hideModules",
                 "title": "Hide modules in Workspace"
+            },
+            {
+                "command": "language-julia.copyHowToType",
+                "title": "Julia: How to Type Selected Latex/Emoji to Clipboard"
             }
         ],
         "menus": {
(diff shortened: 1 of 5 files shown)


davidanthoff:
Seems to me this should be implemented in the language server, there is no good reason why this should only work with a running REPL, right?

xgdgsc:
I don' t know the statistics of how many people would run without REPL. If it' s in Language Server it would be better. I don' t know the performance impact if every hover tries to find this info versus on demand finding.

pfitzseb:
Agreed, this should happen in the LS. We can create a lookup table for this, much like the REPL code does: https://github.com/JuliaLang/julia/blob/88230581592d5a80c636c6a3ad592076cd25e0e2/stdlib/REPL/src/docview.jl#L391-L410

davidanthoff:
The main question is probably whether this info can just be added to the hover that is already returned? That seems the easiest and best solution?

pfitzseb:
It can, and yes, agreed :)

davidanthoff:
Ok, great, that should also actually make the implementation _much_ easier. @xgdgsc I think essentially you just need to make some changes to https://github.com/julia-vscode/LanguageServer.jl/blob/master/src/requests/hover.jl to get this working, without the need to define new messages, or make any other changes in the client itself.

PR #3162, opened 2022-11-24
Add some upcoming julia versions to executable search

fredrikekre:
This patch adds some upcoming julia versions to the executable search such that the extensions is already ready when those are released. Without doing this it seems like it is always outdated (at least that is my experience when trying to guide students to set this up). And I know juliaup might be preferred, but the standard installer should still IMO work, and the current user experience is really bad...


davidanthoff:
Yep, that is exactly how we should be doing it in theory. trying to be a few steps ahead. In practice we often forget, as you can tell ;) Thanks!

PR #3269, opened 2023-04-09
allow the user defined JULIA_LOAD_PATH to be used in the extension

tkuraku:
This change should allow the extension to use the same environment as starting the REPL from the command line.  Currently if a user defines a JULIA_LOAD_PATH to make local modules importable it doesn't not work in the extension. For example if in ~/.bashrc a user defines

```sh
export JULIA_LOAD_PATH="/path/to/my/julia/modules:$JULIA_LOAD_PATH"
```

It is ignored when starting a REPL session in vscode. I think the REPL in the vscode extension should have the same behavior as the REPL started from the command line. This facilitates local utility modules that probably don't warrant being made into their own full on package.


davidanthoff:
There is nothing in the extension that is preventing the REPL from picking up a `JULIA_LOAD_PATH`, that should just work out-of-the-box, I think.

The change in this PR here is _only_ changing the environment in which the language server is running (has nothing to do with the REPL), and for that process we are very careful to _not_ pick up things like `JULIA_LOAD_PATH`. We have a long-standing issue here somewhere to teach the language server about custom `JULIA_LOAD_PATH` settings, but that is a much more involved change.

tkuraku:
Thanks for the feed back. However, this doesn't seem to be working. I have the following in my bashrc

```sh
export JULIA_LOAD_PATH="/path/to/my/julia/modules:$JULIA_LOAD_PATH"
```

starting the REPL outside vscode and running `Base.load_path()` has the desired results, but running `Base.load_path()` in the REPL started with the vscode extension shows that `/path/to/my/julia/modules` is missing. So the REPL is behaving differently in vscode.

PR #3291, opened 2023-05-05
Fix/tmux session name

### LINKED ISSUE #2851: Failed to close tmux session: can't find session: julia_vscode_
I have `"julia.persistentSession.tmuxSessionName": "julia_vscode_${workspaceFolderBasename}",`

set.
Only on one of my machines it happens to when I do `stop REPL` the error of 
`Failed to close tmux session: can't find session: julia_vscode_` shows. While `start REPL ` works correctly with the `${workspaceFolderBasename}` resolved on the same machine ( its a remote session). On all other machines with the same version of vscode(1.66.2) and julia-vscode(v1.6.17) there' s no this issue. I' ve tried remove `.vscode-server/bin` and killing all code processes on this machine and reinstall but with same results. 

It' s a minor issue that affects only one machine. But it' s really weird.

xgdgsc:
Fixes #2851 .

 And fix when using ` "julia.persistentSession.tmuxSessionName": "julia_vscode_${workspaceFolderBasename}",`   when `workspaceFolderBasename` contains `.` (like in `StaticCompiler.jl`) , in newer tmux versions (like version  3.2a in Ubuntu 22.04) it would fail to attach to a tmux session.


src/interactive/repl.ts
```diff
@@ -195,8 +206,8 @@ async function startREPL(preserveFocus: boolean, showTerminal: boolean = true) {
             shellArgs = [
                 <string>config.get('persistentSession.shellExecutionArgument'),
                 // create a new tmux session, set remain-on-exit to true, and attach; if the session already exists we just attach to the existing session
-                `tmux new -d -s ${sessionName} "${juliaAndArgs}" && tmux set -q remain-on-exit && tmux attach -t ${sessionName} ||
-                tmux send-keys -t ${sessionName}.left ^A ^K ^H '${connectJuliaCode}' ENTER && tmux attach -t ${sessionName}`
+                `sessionNameSanitized=\`echo "${sessionName}" | sed "s/\\./_/g" \` && tmux new -d -s $sessionNameSanitized "${juliaAndArgs}" && tmux set -q remain-on-exit && tmux attach -t $sessionNameSanitized ||
```
pfitzseb, line 209: Let's do the replacement in `parseSessionArgs` (or at least on `sessionName` directly).

src/interactive/repl.ts
```diff
@@ -65,7 +64,19 @@ async function stopREPL(onDeactivate=false) {
...
+                    JULIA_EDITOR: getEditor()
+                }
+                vscode.window.createTerminal({
+                    name: `Julia REPL killed`,
+                    shellPath:  shellPath,
+                    shellArgs: shellArgs,
+                    isTransient: true,
+                    env: env,
+                    hideFromUser: true,
+                } as any)
```
pfitzseb, line 79: Can you explain why this is necessary? Invoking `tmux` directly should work fine, no?

src/interactive/repl.ts
xgdgsc: Is that possible? `workspaceFolderBasename` seems to be still not resolved to path name after `parseSessionArgs`.

src/interactive/repl.ts
xgdgsc: Invoking tmux with `exec` would see an empty `workspaceFolderBasename`.

PR #3328, opened 2023-06-13
Extensible Inlay Hints

Zentrik:
This  works the same way the diagnostics displaying works, this enables packages like Cthulhu to show types with the source code.

Example:
```jl
struct YourType end

Base.show(io::IO, ::MIME"application/vnd.julia-vscode.inlayHints", t::YourType) = Dict("C:/Users/gag/Documents/Code/julia-vscode/testing.jl" => [(
    position = (3, 12), # line, column (0 indexed)
    label = "::R",
    kind = 1, # optional; 1: Type, 2: Parameter, nothing: undefined
    tooltip = "test", # optional
    paddingLeft = false, # optional
    paddingRight = false # optional
)])

YourType()
```
image


pfitzseb:
This is awesome. Thanks for the great contribution!

PR #3378, opened 2023-08-22
if string comparison in @test fails, show as diff

### LINKED ISSUE #3121: Use diff UI feature for failed tests

brychcy:
Fixes #3121.

### DIFF
--- a/scripts/packages/VSCodeTestServer/src/VSCodeTestServer.jl
+++ b/scripts/packages/VSCodeTestServer/src/VSCodeTestServer.jl
@@ -160,13 +160,42 @@ function run_testitem_handler(conn, params::TestserverRunTestitemRequestParams)
 
             return TestserverRunTestitemRequestParamsReturn(
                 "failed",
-                [TestMessage(sprint(Base.show, i), Location(filepath2uri(string(i.source.file)), Range(Position(i.source.line - 1, 0), Position(i.source.line - 1, 0)))) for i in failed_tests],
+                [ create_test_message_for_failed(i) for i in failed_tests],
                 elapsed_time
             )
         else
             rethrow(err)
         end
     end
+end
+
+function create_test_message_for_failed(i)
+    (expected, actual) = extract_expected_and_actual(i)
+    return TestMessage(sprint(Base.show, i),
+        expected,
+        actual,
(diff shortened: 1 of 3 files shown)


scripts/packages/VSCodeTestServer/src/VSCodeTestServer.jl
```diff
@@ -169,6 +169,33 @@ function run_testitem_handler(conn, params::TestserverRunTestitemRequestParams)
...
+                if m !== nothing
+                    expected = unescape_string(m.captures[1])
+                    actual = unescape_string(m.captures[2])
+                    return (expected, actual)
+                end
+            end
+        end
+    catch err
+        # if anything goes wrong (e.g. unescaping failed), just return nothing (meaning no diff will be shown)
+    end
```
davidanthoff, line 195: I'm generally not a fan of try catch blocks that blanket hide errors, I think I would prefer that we just fail if things go wrong, and then hopefully telemetry will tell us about it and we can fix it? Unless this is handling a specific failure mode that you know about, but in that case I think we should also try to handle that more narrowly.

davidanthoff (changes requested):
This is great, thanks so much!

I have one small request for a change.

We eventually need to port this change over to https://github.com/julia-vscode/TestItemServer.jl, I'm in the middle of moving the current code here in the julia-vscode repo over into that separate repo, but I can handle that once we have merged this here, so feel free to ignore.

scripts/packages/VSCodeTestServer/src/VSCodeTestServer.jl
brychcy: I wasn't sure if  result.data exists in older julia versions, but I've now checked, and it exists at least since 1.0.

the only thing that could happen is what is already mentioned in the comment: unescaping might fail if the user registers a weird failure result, e.g.
@testitem "#bad_escape" begin
    Test.record(Test.get_testset(), Test.Fail(:test,"" , "\"\\,\" == \"\"", false, LineNumberNode(2, @__FILE__)))
end

I've pushed an updated commit that narrow the try...catch to those unescape statements.

davidanthoff (approved):
Excellent, thanks so much :)

PR #3384, opened 2023-08-27
Run CI on Julia 1.10

davidanthoff:


davidanthoff:
Hm, it looks as if this is not actually creating a job for Julia 1.10? But why?

PR #3417, opened 2023-09-23
Support single quoted and pythoncall python syntax highlighting

LilithHafner:
This enables python syntax highlighting for `py"labmda x: x+1"`, `@pyeval "import math"`, `@pyexec "1+1"`, and  `@pyexec """1+1"""`. Previously this was only enabled for triple quoted calls to the `py` string macro. I think it's worth giving special syntax highlighting to PythonCall in addition to PyCall.

Before:
Screenshot 2023-09-23 at 2 58 37 PM

After:
Screenshot 2023-09-23 at 2 58 45 PM

### DIFF
--- a/syntaxes/julia_vscode.json
+++ b/syntaxes/julia_vscode.json
@@ -509,7 +509,7 @@
           ]
         },
         {
-          "begin": "(py)(\"\"\")",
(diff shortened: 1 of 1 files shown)


pfitzseb:
Please open this against https://github.com/JuliaEditorSupport/atom-language-julia.

PR #3428, opened 2023-10-04
Allow setting parent logger in VSCodeLogger

pepijndevos:
This allows writing stuff like
```julia
with_logger(VSCodeLogger(ConsoleLogger(Debug))) do
    @debug "hello world"
    @logmsg LogLevel(-1) "Progress" progress=0.5
end
```

Untested for now, just firing off a quick suggestion from the web

### DIFF
--- a/scripts/packages/VSCodeServer/src/progress.jl
+++ b/scripts/packages/VSCodeServer/src/progress.jl
@@ -1,4 +1,7 @@
-struct VSCodeLogger <: Logging.AbstractLogger end
+struct VSCodeLogger <: Logging.AbstractLogger
+    parent::Logging.AbstractLogger
+end
+VSCodeLogger() = VSCodeLogger(Logging.current_logger())
 
 const logger_lock = ReentrantLock()
 function Logging.handle_message(j::VSCodeLogger, level, message, _module,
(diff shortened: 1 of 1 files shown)


pfitzseb:
Pretty sure something like this was the original implementation of `VSCodeLogger`, but this stops working correctly when someone sets a new global logger in their code, which will never be used because the `VSCodeLogger` is only instantiated once per executed code  block.

What issue are you trying to solve here?

pepijndevos:
Pretty much what I wrote in the example, contextually log debug messages without setting the global logger.

We have a sort of context struct, with configurable logging options, and things running in the context of that struct should use the configured logging options. We log progress updates, so when we're running inside VSCode we want to use its logger. However, there is no way to specify the log level on the VSCode logger without updating the global logger.

In a regular terminal we use TerminalLoggers, which does support a loglevel.

One hacky workaround would be to just set a new global logger with Debug. Another workaround would be to ONLY sent progress updates to VSCodeLogger via some LoggingExtras filtered logger.

pfitzseb:
Alright, then let's go with this but make `parent::Union{Nothing,Logging.AbstractLogger}`.

Issue #3320, opened 2023-06-06
Update JuliaFormatter?


gdalle:
I realized while testing packages that the Julia extension doesn't use the latest version of JuliaFormatter, is there a way to force this behavior?

hyrodium:
x-ref: https://github.com/julia-vscode/julia-vscode/issues/3226#issuecomment-1451803980

PR #3453, opened 2023-10-31
Update dependencies

davidanthoff:
Fixes https://github.com/julia-vscode/julia-vscode/issues/3320 (and updates lots of other things as well).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -212,9 +212,9 @@
             }
         },
         "node_modules/@eslint-community/regexpp": {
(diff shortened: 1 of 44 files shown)


davidanthoff:
It is already pretty automated, i.e. there is a node script that updates everything. I don't think we want to make it even more automatic, there are too many different things that depend on each other, so I think we still want to keep the manual trigger of updating things.

hyrodium:
I see. Thank you for the explanation!

PR #3472, opened 2023-11-23
Fix displaying of inlay hints when plot pane disabled

Zentrik:
Fixes https://github.com/JuliaDebug/Cthulhu.jl/issues/517


pfitzseb (approved):
Good catch, thanks

Issue #1781, opened 2020-11-19
environmentPath can't handle configuration variables


mortenpi:
I am trying to set `environmentPath` to something like `"julia.environmentPath": "${workspaceFolder:Foo}/subdir"` that relies on [VS Code's configuration variables](https://code.visualstudio.com/docs/editor/variables-reference#_configuration-variables). This seems to lead to:

```
mainThreadExtensionService.ts:66 TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be one of type string, Buffer, or URL. Received type undefined
	at Object.watchFile (fs.js:1345)
	at /home/mortenpi/.vscode/extensions/julialang.language-julia-1.0.10/dist/extension.js:1
	at Generator.next (<anonymous>)
	at s (/home/mortenpi/.vscode/extensions/julialang.language-julia-1.0.10/dist/extension.js:1)
```

with the language server being just stuck on "Starting Julia Language Server..." on the taskbar. I presume the variable is not expanded early enough.

Interestingly, I can still start the REPL and it actually starts in the correct environment, so somewhere down the line it looks like the variable does get expanded.

marius311:
I think`"."` is another thing it can't handle, needs an absolute path.

davidanthoff:
Closing: this was fixed in v1.223.0 — the client now resolves the environment path itself, expanding `${env:...}` and `${config:...}`, and relative paths are accepted.

Please reopen or file a new issue if you still see it on a current release.

PR #3477, opened 2023-11-30
Support more VSCode variables in `environmentPath`

tpgillam:
Some more progress towards #1781.

This adds support for:
 *  `${userHome}`
 *  `${workspaceFolder}`
 *  `${workspaceFolder:<FOLDER_NAME>}`  (to support [multi-root workspaces](https://code.visualstudio.com/docs/editor/workspaces#_multiroot-workspaces))
 *  `${pathSeparator}`

The implementation is suboptimal, in that it effectively duplicates functionality [present in core VSCode](https://github.com/microsoft/vscode-python/blob/main/src/client/common/variables/systemVariables.ts), but this seems unavoidable until https://github.com/microsoft/vscode/issues/46471 is solved properly.

**Notes** 
For `${workspaceFolder}`, the behaviour I've implemented is different in the case of a multi-root workspace. The standard behaviour will take the workspace of the current file (IIUC [this](https://github.com/microsoft/vscode-python/blob/bffc9b36e414e8f4919f09060f95196e1dabe05f/src/client/common/variables/systemVariables.ts#L109-L110) correctly). Two reasons for this:
  - I couldn't easily see how to gain access to the currently open file at this point
  - I suspect that if using this for `julia.environmentPath` then the behaviour could be unintuitive, since the julia environment is only activate once... so the behaviour would depend on which file the user happened to have open first.

Personally I have not used a multi-root workspace in my regular workflow, so not sure what behaviour would be most desirable here. But requiring explicit selection seems like a sensible initial choice to me.

In the case of failure (e.g. setting `{"julia.environmentPath": "${workspaceFolder:non-existent-directory}"}`, currently failure will be silent, and the global julia project will be activated. Maybe we should log somewhere at least? Not sure on the standard approach here, comments welcome.

btw, I'm not sure if there's anywhere to add tests for this? I had a quick look and couldn't find anything obvious.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -90,22 +90,34 @@ export function wrapCrashReporting(f) {
 }
 
 export function resolvePath(p: string, normalize: boolean = true) {
-    p = parseEnvVariables(p)
-    p = parseWorkspaceFolder(p)
+    p = parseVSCodeVariables(p)
     p = p.replace(/^~/, os.homedir())
     p = normalize ? path.normalize(p) : p
     return p
 }
 
-function parseEnvVariables(p: string) {
-    return p.replace(/\${env:(.*?)}/g, (_, variable) => {
+/**
+ * Parse a subset of VSCode 'variables' in `p`, and return a string with the replacements.
+ *
+ * Specifically, we support:
+ *  - ${userHome}
+ *  - ${workspaceFolder}
+ *  - ${pathSeparator}
+ *  - ${env:<ENVIRONMENT_VARIABLE>}
+ *
+ * See https://code.visualstudio.com/docs/editor/variables-reference for definitions of the
+ * above.
+ */
+function parseVSCodeVariables(p: string) {
+    p = p.replace(/\${userHome}/g, os.homedir())
+    p = p.replace(/\${workspaceFolder}/g, (_) => {
+        const workspace_folders = vscode.workspace.workspaceFolders
+        // We do not support multi-root workspaces.
+        return workspace_folders.length == 1 ? vscode.workspace.workspaceFolders[0].uri.fsPath : null;
+    })
+    p = p.replace(/\${pathSeparator}/g, path.sep);
+    p = p.replace(/\${env:(.*?)}/g, (_, variable) => {
         return process.env[variable] || ''
     })
-}
-
-function parseWorkspaceFolder(p: string) {
-    return p.replace(/\${workspaceFolder}/g, (_) => {
-        const workspace_folders = vscode.workspace.workspaceFolders
-        return workspace_folders.length > 0 ? vscode.workspace.workspaceFolders[0].uri.fsPath : null;
-    })
+    return p
 }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -91,6 +91,7 @@ export function wrapCrashReporting(f) {
 
 export function resolvePath(p: string, normalize: boolean = true) {
     p = parseEnvVariables(p)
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -90,14 +90,51 @@ export function wrapCrashReporting(f) {
...
+function parseVSCodeVariables(p: string) {
+    p = p.replace(/\${userHome}/g, os.homedir())
+
+    const workspace_paths = vscode.workspace.workspaceFolders.map((folder) => {
+        return folder.uri.fsPath
+    })
+    p = p.replace(/\${workspaceFolder}/g, (_) => {
+        // We do not support multi-root workspaces here.
+        // In that case the user should use the ${workspaceFolder:<FOLDER_NAME>} syntax.
+        return workspace_paths.length === 1 ? workspace_paths[0] : null
```
pfitzseb, line 125: Elsewhere in the extension we're defaulting to the first workspace folder, which I think we should do here as well (could use `workspaceFolder1` for that to be explicit, I guess).

tpgillam: OK, let's default to the first one here too then. (Introducing a new name seems like it could add more confusion... especially if it were only done here and not in the other places!)

tpgillam: (Done)

tpgillam:
Thanks for the review @pfitzseb ! I added a line to the changelog too just now. Anything else that'd be useful from my side?

pfitzseb (approved):

xgdgsc:
I guess this breaks the use  of the extension when there's no folder/workspace open.

tpgillam:
> I guess this breaks the use of the extension when there's no folder/workspace open. Error at
> 
> https://github.com/julia-vscode/julia-vscode/blob/18b3f6f5ea238115d978c69255f3c8daa686506b/src/utils.ts#L119
> 
> ```
> stack trace: TypeError: Cannot read properties of undefined (reading 'map')
>     at parseVSCodeVariables (/data/Documents/oss/julia-vscode/src/utils.ts:119:63)
>     at resolvePath (/data/Documents/oss/julia-vscode/src/utils.ts:93:9)
>     at /data/Documents/oss/julia-vscode/src/jlpkgenv.ts:38:49
>     at Generator.next (<anonymous>)
>     at /data/Documents/oss/julia-vscode/dist/extension.js:86561:71
>     at new Promise (<anonymous>)
>     at __webpack_modules__../src/jlpkgenv.ts.__awaiter (/data/Documents/oss/julia-vscode/dist/extension.js:86557:12)
>     at switchEnvToPath (/data/Documents/oss/julia-vscode/dist/extension.js:86597:12)
>     at Object.<anonymous> (/data/Documents/oss/julia-vscode/src/jlpkgenv.ts:272:11)
>     at Generator.next (<anonymous>)
>     at fulfilled (/data/Documents/oss/julia-vscode/dist/extension.js:86558:58)
>     at processTicksAndRejections (node:internal/process/task_queues:95:5)
> ```

Ah thanks for the catch. I think @pfitzseb has already merged another PR that builds on this so reversion is probably non-trivial at this point. So maybe open a new issue?

Anyway, afraid I don't have time to make a PR right now, but looks like it should be a fairly simple fix. Maybe this?

```typescript
const workspace_paths = (vscode.workspace.workspaceFolders ?? []).workspace_folders.map(<snip, as before>)
```

but I'm not 100% convinced that this is the correct syntax!

pfitzseb:
Thanks, good catch .
-> https://github.com/julia-vscode/julia-vscode/pull/3490

PR #3485, opened 2023-12-06
Run the Juliaup detection in a shell on Mac/Linux

davidanthoff:
We had a couple of reports where folks launch VS Code on MacOS from the graphical finder UI, and that seems to end up with a VS Code process that doesn't have the Juliaup `PATH` modifications. This seems one easy way here to fix this. I haven't tested it, though...


pfitzseb (approved):
Seems to work fine.

PR #3491, opened 2023-12-14
feat: fix env handling and improve tmux behaviour

pfitzseb:


pfitzseb:
Accidentally pushed this to main directly in https://github.com/julia-vscode/julia-vscode/commit/2d33ccf860e31c1f5850388291b9164b5b38fa51.

PR #3494, opened 2023-12-15
machine-overridable of closeStrategy

xgdgsc:
Should this be machine-overridable instead?


pfitzseb (approved):
Sure, I don't see why not. Wanna put a mention into the changelog?

Issue #3505, opened 2024-01-05
Unnecessary Activation of Julia Extension in Jupyter Notebooks where Julia is not involved


singularitti:
I've encountered an issue with the Julia extension in Visual Studio Code that I hope can be addressed. Whenever I open a Jupyter notebook within VSCode, the Julia extension automatically activates and starts a Julia process. This has become a significant concern for me as it consumes a considerable amount of memory.

The primary issue is that my work in VSCode is solely focused on Python, and I do not require Julia for my current projects. The automatic activation of the Julia process is therefore unnecessary for my use case and impacts the performance of my system.

Could you please guide me on how to disable this automatic activation behavior of the Julia extension when opening Jupyter notebooks? A setting or a workaround to prevent the extension from starting up unless explicitly required would be greatly beneficial.

PR #3510, opened 2024-01-08
chore: clean up activation events

pfitzseb:
and remove `onNotebook:jupyter-notebook` event

Fixes https://github.com/julia-vscode/julia-vscode/issues/3505.


davidanthoff:
> and remove `onNotebook:jupyter-notebook` event

Can we do that? Wouldn't that mean that if someone opens an empty folder, creates a new notebook, that they won't even be shown the Julia kernels, unless they first do something else to activate the extension?

pfitzseb:
Yes, indeed. Setting the cell language to Julia will trigger activation though, which might be good enough.

davidanthoff:
I think what we really need is a state where the extension is loaded, but things like the LS aren't started... Some kind of lightweight state that the extension can be in when there isn't any Julia file in the workspace yet.

pfitzseb:
Yes, but the question is what we want to do in the meantime. Imho this trade-off is worth it.

davidanthoff:
> Imho this trade-off is worth it.

I don't know, that really degrades the notebook experience if one first has to select the cell language and then select a kernel. In my mind the flow really is to select the kernel first, and that then sets the language for the cells...

singularitti:
Is it possible to make it configurable then?

davidanthoff:
> Is it possible to make it configurable then?

I don't think so...

singularitti:
> that they won't even be shown the Julia kernels, unless they first do something else to activate the extension?

I wouldn't mind it. Usually when I want to activate the Julia notebook, there are already some Julia files in it. I could just open one of them. In fact, I don't think we should by default have the Julia extension activated whenever we have a notebook: sometimes I just want to write Python code and the Julia language server bites a lot of memory even I'll never use it. What about people who use other languages? Jupyter supports lots of languages besides Python. I think they don't want to see the same thing either.

davidanthoff:
> Usually when I want to activate the Julia notebook, there are already some Julia files in it.

Certainly not for me. I _very_ often create a new empty folder, start VS Code in it and create an empty notebook and select a Julia kernel.

I think at the end of the day this is just a bit of a design problem in the notebook API of VS Code. They should have had a way to declare that an extension can offer kernels, so that loading of the extension can be delayed further.

I'm all for us changing things so that the LS doesn't _always_ start up when the extension gets loaded. I'm not sure how difficult that is, but maybe it would actually be not super complicated?

spinicist:
I think this should be added.

I mainly use VS Code for C++ development, with some Python notebooks. I've recently started trying out Julia. It's highly annoying for the Julia environment to keep loading itself when I'm working in my other projects that don't use the language. I'm going to disable to extension for now and only re-enable it when I go into my Julia project.

PR #3519, opened 2024-01-19
feat: static inlay hints

pfitzseb:


Zentrik:
Do you have any thoughts on how to integrate this with Cthulhu's use of inlay hints. The parameter name hints seem to work fine in conjunction with Cthulhu but the variable type hints cause duplicated hints (both the Language Server and Cthulhu print a hint).

My initial thought is for Cthulhu to disable the language server providing variable hints when Cthulhu is providing them (If we can it would be nice to only disable the language server specifically on the code for which Cthulhu is providing hints).

pfitzseb:
Yeah, this is a bit tricky. We have much less control over the LS inlay hints than the runtime hints, because their in-editor display is not in our hands.

pfitzseb:
LS.jl got updated, so I'll go ahead and merge this too so people can control inlay hints properly.

Issue #3526, opened 2024-02-01
`LoadError: UndefVarError: `is_GotoIfNot` not defined`


singularitti:
I'm constantly seeing this error when starting the REPL in VSCode with `Option+J Option+O`, which only happen recently:

```julia
[ Info: Precompiling VSCodeServer [9f5989ce-84fe-42d4-91ec-6a7a8d53ed0f]
[ Info: Precompiling Revise [295af30f-e4ad-537b-8983-00126c2a3abe]
ERROR: LoadError: UndefVarError: `is_GotoIfNot` not defined
Stacktrace:
 [1] include
   @ Base ./Base.jl:495 [inlined]
 [2] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt128}}, source::Nothing)
   @ Base ./loading.jl:2216
 [3] top-level scope
   @ stdin:3
in expression starting at ~/.vscode/extensions/julialang.language-julia-1.66.2/scripts/packages/Revise/src/Revise.jl:1
in expression starting at stdin:3
```

```julia
julia> versioninfo()
Julia Version 1.10.0
Commit 3120989f39b (2023-12-25 18:01 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 10 × Apple M1 Pro
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, apple-m1)
  Threads: 11 on 8 virtual cores
Environment:
  JULIA_NUM_THREADS = auto
  JULIA_EDITOR = code
  JULIA_PKG_SERVER_REGISTRY_PREFERENCE = eager
  JULIA_IMAGE_THREADS = 8
```

Extension version: 1.66.2

VSCode:

```
Version: 1.86.0
Commit: 05047486b6df5eb8d44b2ecd70ea3bdf775fd937
Date: 2024-01-31T10:29:11.933Z
Electron: 27.2.3
ElectronBuildId: 26495564
Chromium: 118.0.5993.159
Node.js: 18.17.1
V8: 11.8.172.18-electron.0
OS: Darwin arm64 23.3.0
```

PR #3541, opened 2024-02-14
Remove imports of removed JuliaInterpreter functions

visr:
In https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/598 `is_GotoIfNot` and `is_ReturnNode` were removed. With #3531 we started using JuliaInterpreter 0.9.28 which includes this change. This breaks the Julia test server:

```
 Info: Starting the Julia Test Server
ERROR: LoadError: UndefVarError: `is_GotoIfNot` not defined
Stacktrace:
 [1] include(mod::Module, _path::String)
   @ Base .\Base.jl:495
 [2] include(x::String)
   @ VSCodeTestServer c:\Users\visser_mn\.vscode\extensions\julialang.language-julia-1.71.1\scripts\packages\VSCodeTestServer\src\VSCodeTestServer.jl:1
```

With this change, that is fixed. These two functions are not used in this repo or its submodules.

The removal of these functions also caused #3526 but I think that has already been fixed by using a newer Revise version, so I think that issue can be closed as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 ### Fixed
 * Static inlay hints are now automatically disabled when runtime hints are displayed ([#3539](https://github.com/julia-vscode/julia-vscode/pull/3539))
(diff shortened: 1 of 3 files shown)


pfitzseb (approved):

pfitzseb:
Thanks!

Issue #3542, opened 2024-02-15
Can't load Cthulhu inside vs-code REPL


oxinabox:
Some how VS-Code is putting an funky (old?) version of JuliaSyntax on the load path,
and it is breaking Cthulhu.
I wouldn't be shocked if this was something i setup at one point and forgot about.

This is the error i see:
```
julia> using Cthulhu

Precompiling Cthulhu
        Info Given Cthulhu was explicitly requested, output will be shown live 
ERROR: LoadError: UndefVarError: `parsestmt` not defined
Stacktrace:
  [1] tsn_and_mappings(m::Method, src::Core.CodeInfo, rt::Any, sourcetext::SubString{String}, lineno::Int32; warn::Bool, strip_macros::Bool, kwargs::@Kwargs{})
    @ TypedSyntax ~/.julia/packages/TypedSyntax/2dDMg/src/node.jl:44
  [2] tsn_and_mappings(m::Method, src::Core.CodeInfo, rt::Any; warn::Bool, strip_macros::Bool, kwargs::@Kwargs{})
    @ TypedSyntax ~/.julia/packages/TypedSyntax/2dDMg/src/node.jl:39
  [3] get_typed_sourcetext(mi::Core.MethodInstance, src::Core.CodeInfo, rt::Any; warn::Bool)
    @ Cthulhu ~/.julia/packages/Cthulhu/SWgb3/src/reflection.jl:354
  [4] find_callsites(interp::Cthulhu.CthulhuInterpreter, CI::Core.CodeInfo, stmt_infos::Vector{Core.Compiler.CallInfo}, mi::Core.MethodInstance, slottypes::Vector{Any}, optimize::Bool, annotate_source::Bool, pc2excts::Nothing)
    @ Cthulhu ~/.julia/packages/Cthulhu/SWgb3/src/reflection.jl:37
  [5] _descend(term::REPL.Terminals.TTYTerminal, interp::Cthulhu.CthulhuInterpreter, curs::Cthulhu.CthulhuCursor; override::Nothing, debuginfo::Symbol, optimize::Bool, interruptexc::Bool, iswarn::Bool, hide_type_stable::Bool, verbose::Nothing, remarks::Bool, with_effects::Bool, exception_type::Bool, inline_cost::Bool, type_annotations::Bool, annotate_source::Bool, inlay_types_vscode::Bool, diagnostics_vscode::Bool, jump_always::Bool)
[...]

PR #3543, opened 2024-02-15
Always popfirst the load path, even if using fails

oxinabox:
Avoids #3542 from happening again.

I just editted this in Github so this will definately need squashing.


oxinabox:
idk if i needed to change all of these.

scripts/terminalserver/terminalserver.jl
```diff
@@ -6,8 +6,11 @@ let distributed = Base.PkgId(Base.UUID("8ba89e20-285c-5b6f-9357-94700520ee1b"),
         Distributed.remotecall_eval(Main, 1:Distributed.nprocs(), :(popfirst!(LOAD_PATH)))
```
pfitzseb, line 6: This needs a try/finally too.

pfitzseb (approved):

PR #3562, opened 2024-03-07
Handle \r in outputs

### LINKED ISSUE #3561: Handle `\r` in the notebook
Running
```julia
print("Hello World")
print("\r")
print("1234")
```
with a Jupyter/IJulia kernel results in output
```
1234o World
```
but running it in the extension Julia kernel results in
```
Hello World
1234
```

If we just handled the `\r` like the Jupyter/IJulia kernel did, things like ProgressMeter would look a lot better, at least in some basic use-cases.

@DonJayamanne and @amunger any suggestion on how we might want to do that? Is there special handling in the Jupyter extension for this that we could just replicate somehow? The code that currently handles this in the extension is https://github.com/julia-vscode/julia-vscode/blob/696ce5099862668a242ba6bc0a3cd22de1401698/src/notebook/notebookKernel.ts#L255, presumably we would want to change that somehow?

DonJayamanne:
Fixes #3561

### DIFF
--- a/src/notebook/notebookKernel.ts
+++ b/src/notebook/notebookKernel.ts
@@ -250,27 +250,28 @@ export class JuliaKernel {
                     }
                 })
 
+                const outputsPerExecution = new WeakMap<vscode.NotebookCellExecution,{output: vscode.NotebookCellOutput | undefined, name: 'stdout' |'stderr'}>()
                 this._msgConnection.onNotification(
                     notifyTypeStreamoutput,
                     ({ name, data }) => {
-                        if (name === 'stdout') {
-                            const execution = this._currentExecutionRequest
-                            if (execution) {
-                                execution.appendOutput([
-                                    new vscode.NotebookCellOutput([
-                                        vscode.NotebookCellOutputItem.stdout(data),
-                                    ]),
-                                ])
-                            }
-                        } else if (name === 'stderr') {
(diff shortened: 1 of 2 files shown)


src/notebook/notebookKernel.ts
```diff
@@ -250,27 +250,28 @@ export class JuliaKernel {
...
-                                    ]),
-                                ])
-                            }
+                        const execution = this._currentExecutionRequest
+                        if (!execution) {
+                            return
+                        }
+                        if (name === 'stdout' || name === 'stderr') {
+                            // Ensure \r is in a seprate line.
+                            data.split(/(\r)/).forEach((line) => {
```
DonJayamanne, line 263: The reason we need to split them is because the assumption with notebooks is 
* \r would come in a separate output from a process
E.g. a progress message (bar) would display some line,
Then the next stream output would contain the `\r`.
Thus the first line of the next stream would be `\r`
Its very unlikely that `\r` would be in the middle of the string, if thats the case, the code that wrote the message might as well update the string and then write the final string to the stream.

I guess we get a combined string here because the julia code is combining them at some point.

pfitzseb:
Can you add a quick note to our changelog please?

davidanthoff (approved):
Fantastic, thanks so much :)

PR #3585, opened 2024-04-14
CI: Run CI on Julia 1.11, and migrate to `julia-actions/install-juliaup`

DilumAluthge:
Alternative to #3584.

Fixes #.


pfitzseb:
What's the advantage of using the setup-juliaup action?

davidanthoff:
We can start to put a version like 1.11 into the mix as soon as there is an alpha build. For us it essentially makes it easy to add testing on upcoming betas.

Issue #858, opened 2019-09-12
Execute Code Cell: Add Codelens


bolducke:
It would be nice to have codelens similar to the python extension with Jupyter Notebook.

![image](https://user-images.githubusercontent.com/26026137/64813324-17ddaa80-d56f-11e9-99cf-3e69610413ab.png)

otobrzo:
Or maybe change background color for active cell, similarly to MATLAB.
In present state, it is hard to see what exactly I will run.

phlavenk:
Also, on my system execute cell does not move cursor to next cell. That would be great to have as well as a shortcut to move to cell up/down with cursor.

davidanthoff:
Great suggestion, I moved that to a separate issue, because implementing these two is independent.

brenhinkeller:
+1 for this, and generally anything that can help make code cells more visible/intuitive to beginners!

RoyiAvital:
I find the cell experience in VS Code interactive mode to be great: https://github.com/microsoft/vscode-jupyter/issues/273.
I wish the Julia VS Code extension would embrace it as it is literally the https://github.com/mwouts/jupytext standard.

PR #3586, opened 2024-04-15
Codelens Support in `julia` Files

huangyxi:
This PR attempts to support CodeLens for `julia` files (#858), enabling users to explore and execute code cells directly in the editor.

image

The initial commit only demonstrates the availability of CodeLens for `julia` files. To implement the complete functionality, there are two more tasks to be done.

1. Currently, the `julia.executeCell` command is used to execute code cells. However, the range of the code cell is reliant on the pointer position, where the trigger of CodeLens actions may not enclose the current pointer position.
https://github.com/julia-vscode/julia-vscode/blob/9b009459727161f9292d2890d6c5433ec6932960/src/interactive/repl.ts#L856-L857

To address this issue, I propose two separate approaches:
   - **Workaround**: Modify the `julia.executeCell` command to execute either the code cell enclosing the pointer position or the code cell enclosing the CodeLens trigger position.
   - **General solution**: Strip and refactor all source code related to code cell execution to a more general approach used in the vscode-R extension (https://github.com/REditorSupport/vscode-R/blob/master/src/rmarkdown/chunks.ts). The new approach will statically analyze delimiters of all code cells and then dynamically locate the active code cell based on the executed CodeLens or pointer position. Though this approach may require more computation resources, it may provide a more robust and flexible solution for code cell execution.

I would like to hear the community's opinion on this issue. Either way, I'm willing to implement the solution.

2. To introduce a new command `julia.executeAboveCells` to execute all code cells above the current pointer position if the second solution above is adopted, which is also a common feature in Jupyter notebooks and vscode-R extension.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -12,6 +12,7 @@ import { LanguageClient, LanguageClientOptions, RevealOutputChannelOn, ServerOpt
 import * as debugViewProvider from './debugger/debugConfig'
 import { JuliaDebugFeature } from './debugger/debugFeature'
 import * as documentation from './docbrowser/documentation'
+import { CodelensProvider } from './interactive/codelens'
 import { ProfilerFeature } from './interactive/profiler'
 import * as repl from './interactive/repl'
 import { WorkspaceFeature } from './interactive/workspace'
@@ -71,6 +72,7 @@ export async function activate(context: vscode.ExtensionContext) {
                 decreaseIndentPattern: decreaseIndentPattern
             }
         })
+        vscode.languages.registerCodeLensProvider('julia', new CodelensProvider())
 
         const profilerFeature = new ProfilerFeature(context)
         context.subscriptions.push(profilerFeature)
--- a/src/interactive/codelens.ts
+++ b/src/interactive/codelens.ts
@@ -1,0 +1,54 @@
+import * as vscode from 'vscode';
+
+export class CodelensProvider implements vscode.CodeLensProvider {
+
+    private codeLenses: vscode.CodeLens[] = [];
+    private cellDelimiters: string[] = [];
+    private regex: RegExp;
+    private _onDidChangeCodeLenses: vscode.EventEmitter<void> = new vscode.EventEmitter<void>();
+    public readonly onDidChangeCodeLenses: vscode.Event<void> = this._onDidChangeCodeLenses.event;
+
+    constructor() {
+        vscode.workspace.onDidChangeConfiguration((_) => {
+            this._onDidChangeCodeLenses.fire();
+        });
+    }
+
+    public provideCodeLenses(document: vscode.TextDocument, token: vscode.CancellationToken): vscode.CodeLens[] | Thenable<vscode.CodeLens[]> {
+        this.cellDelimiters = vscode.workspace.getConfiguration("julia").get<string[]>("cellDelimiters"); // ['^##(?!#)', '^#(\\s?)%%', '^#-']
+        if (this.cellDelimiters === undefined || this.cellDelimiters.length === 0) {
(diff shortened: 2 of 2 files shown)


pfitzseb:
I didn't look at the linked implementation in detail, but that second approach seems preferable.

RoyiAvital:
Any chance to push this forward?

janckerchen:
related https://github.com/julia-vscode/julia-vscode/issues/3732

mkoculak:
Hi, this looks great!
Not sure if it is in scope of CodeLens integration, but with similar function in python files, there are additional lines bordering the focused code cell. From your screenshots its seems they are not there. Is it something that could be added?

huangyxi:
image

@mkoculak implemented!

pfitzseb:
> Having said that, I think it should probably be implemented via the language server, rather than in the extension. The LSP provides all the required stuff to add code lenses, and that would have many benefits:

This is a bit of a weird case though -- generally, the LSP doesn't specify anything related to code execution and other clients don't have the infrastructure in place for it. 

Also, IIRC dealing with cells was pretty annoying with CSTParser, because these kinds of comments don't get their own nodes and are instead attached as trailing trivia to the previous node, which means you need to do a lot of extra work to figure out proper positioning etc. I guess we're somewhat moving towards JuliaSyntax nowadays, which probably deals with that issue in a much cleaner way.

davidanthoff:
> the LSP doesn't specify anything related to code execution

But the LSP does allow definition of custom commands that are executed on the client, so we could easily define our own commands, and then other clients could decide whether they want to support that or not.

> I guess we're somewhat moving towards JuliaSyntax nowadays

JuliaWorkspaces.jl would have an easy way to hook into the JuliaSyntax nodes. But I'm also not really sure whether those are what we will eventually land on.. All a bit unclear to me at the moment, and maybe actually best to not have too much stuff depend on these new JuliaSyntax data structures for now...

I'll defer to @pfitzseb for a final decision re merging, you would have to do the review of this part in any case, you are much more familiar with the whole code execution stuff :)

huangyxi:
Looking forward to your review @pfitzseb

huangyxi:
Ping @pfitzseb

huangyxi:
Hi @pfitzseb , just checking in. Is there anything I can do to help move this PR forward?

huangyxi:
> We've started putting these mostly self-contained pieces of code into classes, so it would be nice to have this in a e.g. `CellEvaluationFeature` class. This isn't crucial, but helps a bit with lifecycle management, abstraction, and code separation.

The original `CodeLens`, `FoldingRange`, and `JuliaCell` execution logic has been abstracted into `CodeCellFeature`.

In addition, `selectJuliaBlock` and `evaluateBlockOrSelection` from `repl.ts` have been merged. Note that the original `evaluateBlockOrSelection` (now `executeSelectionOrCell`) behaves slightly differently from the existing `executeCell`: the former executes each non-empty selection, or the intersecting cell when no selection is present. These targets are collected into a **list** *(preserving order and allowing duplicates)* and executed in **selection order**, preserving the original behavior. In contrast, `executeCell` treats all targets as a **set** *(deduplicated)* of intersecting cells and executes them in **document order**.

Whether to keep both implementations or consolidate them into one can be discussed later.

huangyxi:
> Seems like this gets called a lot, so I wonder if it's worth adding some caching for this based on document.uri and document.version. Let's not add that additional complexity until we know it helps though.

Meanwhile, `docCells` is now cached based on `document.uri`, `document.version`, and `document.languageId` (to account for rare cases where the language changes between Julia and Julia Markdown), improving performance.

PR #3618, opened 2024-06-13
fix: don't require size for <:AbstractArrays

### LINKED ISSUE #3617: Variables of type JuMP.Containers.SparseAxisArray do not show up in list of Julia REPL workspace variables
I attempted to run a Julia script by executing the active file in REPL to find all the variables the script creates in the workspace.
However, variables of subtypes to JuMP.Containers.SparseAxisArray do not show up in the list of variables in the Julia REPL in the vscode plugin. All other created variables are shown in the list.

julia vscode version: v1.79.2
julia version: 1.9.4
JuMP version: 1.22.1

pfitzseb:
Fixes https://github.com/julia-vscode/julia-vscode/issues/3617.


davidanthoff:
Shouldn't we check `Base.IteratorSize` really?

pfitzseb:
Yeah, I suppose we can do that and look at the iterator interface instead of the AbstractArray interface, which does in fact requires `size` to be implemented: https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array.

davidanthoff:
Ah, I didn't see that they must implement `size`.

I have to say, I'm kind of reluctant to work around bugs in packages... Clearly they are not correctly implementing the abstract array interface, shouldn't we just show an error in that case?

PR #3626, opened 2024-07-05
Add `#+` code seperator for Weave support

kylebutts:


kylebutts (commented):
Weave.jl `convert_doc` produces output with `#+` cell separators (and not `# %%`). This PR makes these work in VSCode

PR #3629, opened 2024-07-07
Remove sysimage feature and fix REPL startup on Julia 1.11

davidanthoff:
We need to fix the REPL startup so that the user project only gets activated _after_ we load `VSCodeServer`, otherwise things break badly on Julia 1.11 if the user has a project that was created e.g. on Julia 1.10.

But if we only activate the project after startup, that doesn't seem to play well with the idea to load a project specific sys image...

I have no evidence, but my best guess is that we have exactly zero users actually using the sysimage feature... Plus, with recent Julia versions and native code caching for packages that feature is also no longer really needed.

Soooo, this PR removes the entire sysimage compile and load feature, and changes the REPL startup code to only activate the user project after we have loaded the extension specific code.

@pfitzseb this needs a careful review from you before merge, a lot of that startup code was quite unfamiliar to me.


scripts/terminalserver/terminalserver.jl
```diff
@@ -1,7 +1,12 @@
-
+import Pkg
```
pfitzseb, line 1: I feel like this will only be guaranteed to work if we start the process in our own env. The global one (which this process starts in implicitly, afaict) doesn't necessarily contain Pkg.

davidanthoff: I started implementing that, but there is a funny chicken-egg situation: if we can't import `Pkg` without having it in a project, then we cannot activate the project :) We can of course activate the project via command line flags, but overall this all seems a bit weird to me... It just generally seems that `Pkg.activate` is functionality that should be available always, completely regardless of any project loaded or not etc

pfitzseb: Mhm, yes. The whole Pkg/extensions story has made me pretty unhappy in that past.

This functionality is also available in `Base.set_active_project` (for 1.8 and up) or `Base.ACTIVE_PROJECT[] = ...`.

davidanthoff: Ok, new version pushed. But thinking about it, we should probably get rid of the `LOAD_PATH` modification and just put the VSCodeServer into the project we activate at the beginning...

pfitzseb: Yes, we should do that if we're having to mess with the active project.

scripts/terminalserver/terminalserver.jl
```diff
@@ -1,28 +1,45 @@
...
+    else
+        Base.set_active_project(joinpath(version_specific_env_path,"Project.toml"))
+    end
+
+
+    args = [popfirst!(Base.ARGS) for _ in 1:9]
+    conn_pipename, debug_pipename, telemetry_pipename, project_path = args[1:4]
+
     if haskey(Base.loaded_modules, distributed) && (Distributed = Base.loaded_modules[distributed]).nprocs() > 1
+        # TODO Move this logic over to the project loading mechanism
```
davidanthoff, line 25: @pfitzseb Not sure what we need to do here. Maybe do a `remotecall_eval` to activate the project everywhere, then do the `using` call, then do another `remotecall_eval` to activate the user project everywhere? I assume activating projects in the main process doesn't automatically propagate to the child workers?

davidanthoff:
Another problem is that now if one tries `versioninfo()` in the REPL it doesn't find it. Somehow with this change we seem to have lost the `InteractiveUtils`, but I don't understand why...

pfitzseb (approved):
LGTM.
Fixes the InteractiveUtils/Sockets issues.

PR #3632, opened 2024-07-09
Fix #3631

### LINKED ISSUE #3631: Open Package Directory in New Window Not Working with Custom Dev Path
The command `Open Package Directory in New Window` does not work correctly when using a custom `dev` path.

The directory is read here:

https://github.com/julia-vscode/julia-vscode/blob/10cad7b3ea835d753f2023cb9a426db517f0bcc6/src/openpackagedirectory.ts#L17

This calls the following function:

https://github.com/julia-vscode/julia-vscode/blob/10cad7b3ea835d753f2023cb9a426db517f0bcc6/src/packagepath.ts#L13-L35

The function reads the path returned by `Pkg.depots()`, but I think it should first check `ENV["JULIA_PKG_DEVDIR"]`, similar to [what Pkg does](https://pkgdocs.julialang.org/v1/managing-packages/#developing).

Also, it ould be nice if it read from the startup file because the environment variable is usually written there.

VEZY:
Fixes #3631

### DIFF
--- a/src/packagepath.ts
+++ b/src/packagepath.ts
@@ -17,10 +17,9 @@ export async function getPkgPath() {
         const res = await execFile(
             juliaExecutable.file,
             [
-                '--startup-file=no',
                 '--history-file=no',
                 '-e',
-                'using Pkg; println(Pkg.depots()[1])'
+                'using Pkg;haskey(ENV, "JULIA_PKG_DEVDIR") ? println(ENV["JULIA_PKG_DEVDIR"]) : println(joinpath(Pkg.depots()[1], "dev"))'
             ],
             {
                 env: {
@@ -29,7 +28,7 @@ export async function getPkgPath() {
                 }
             }
         )
-        juliaPackagePath = join(res.stdout.toString().trim(), 'dev')
+        juliaPackagePath = res.stdout.toString().trim()
     }
     return juliaPackagePath
 }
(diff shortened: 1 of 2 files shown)


VEZY:
The CI is not passing for one action: julia 1.0 windows-latest. I think it is linked to a connection issue on Github's side. Could you re-run this action manually? I don't have the right to do that. 
Note: the previous commit was not passing either, with a lot of connection issues, but making a new commit seemed to fix the issue for most actions.

src/packagepath.ts
```diff
@@ -17,10 +16,9 @@ export async function getPkgPath() {
         const res = await execFile(
             juliaExecutable.file,
             [
-                '--startup-file=no',
                 '--history-file=no',
                 '-e',
-                'using Pkg; println(Pkg.depots()[1])'
+                'using Pkg;haskey(ENV, "JULIA_PKG_DEVDIR") ? println(ENV["JULIA_PKG_DEVDIR"]) : println(joinpath(Pkg.depots()[1], "dev"))'
```
davidanthoff, line 21:
```suggestion
                'using Pkg;println(get(ENV, "JULIA_PKG_DEVDIR", joinpath(Pkg.depots()[1], "dev")))'
```

Here is a slightly more succinct version :) But also happy to merge the original one.

VEZY: I was wondering how I could idiomatically retrieve the values from the environment, but I couldn't find it anymore. Thanks!

davidanthoff (approved):

PR #3637, opened 2024-07-13
feat: explicitly start current Julia version in debug mode

pfitzseb:


davidanthoff:
Hm, do we really want this? Not clear to me what (good) scenario this would enable :)

pfitzseb:
The scenario where I can develop the extension for 1.11 with juliaup. I thought about just making it a dev mode toggle though, which is probably better :shrug: Not entirely sure *how* to query whether the extension is being debugged though.

davidanthoff:
Ha, well, that does seem important, didn't think of that ;) But yes, I think we should handle that via some non-user facing option. We could just add an env var to the launch config, but there is probably a better way to detect debug mode...

pfitzseb:
We're using `process.env.DEBUG_MODE` elsewhere, so I'll push a PR with that instead.

davidanthoff (approved):

PR #3690, opened 2024-09-03
Use stable VS Code API to edit notebook metadata

DonJayamanne:
For https://github.com/microsoft/vscode/issues/227419

- [N/A] End-user documentation check. If this PR requires end-user documentation, please add that at https://github.com/julia-vscode/docs.

@davidanthoff Given that the Notebook API in VS Code is now stable, we can start using that instead of the private extension specific API added.
The plan is to get this merged, and once you have shipped a new version of the Julia extension we can then remove the other internal API `vscodeIpynbApi.setNotebookMetadata` (perhaps in another month or two, giving enough time for Julia extension to adopt and publish the changes)

### DIFF
--- a/src/notebook/notebookFeature.ts
+++ b/src/notebook/notebookFeature.ts
@@ -26,9 +26,6 @@ type JupyterNotebookMetadata = Partial<{
 }>
 
 function getNotebookMetadata(notebook: vscode.NotebookDocument): JupyterNotebookMetadata['metadata'] | undefined{
-    if (notebook.metadata && 'custom' in notebook.metadata){
-        return notebook.metadata.custom?.metadata;
-    }
     return notebook.metadata?.metadata as any;
 }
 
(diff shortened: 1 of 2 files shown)


src/notebook/notebookFeature.ts
```diff
@@ -26,9 +26,6 @@ type JupyterNotebookMetadata = Partial<{
 }>
 
 function getNotebookMetadata(notebook: vscode.NotebookDocument): JupyterNotebookMetadata['metadata'] | undefined{
-    if (notebook.metadata && 'custom' in notebook.metadata){
```
DonJayamanne, line 29: This old approach of storing metadata is also no longer required as VS Code has completely migrated to using the new format

src/notebook/notebookFeature.ts
```diff
@@ -342,8 +335,17 @@ export class JuliaNotebookFeature {
             },
         }
 
-        if (this.vscodeIpynbApi && !isEqual(getNotebookMetadata(notebook), metadata)) {
-            this.vscodeIpynbApi.setNotebookMetadata(notebook.uri, metadata)
+        if (!isEqual(getNotebookMetadata(notebook), metadata)) {
+            const edit = new vscode.WorkspaceEdit();
```
DonJayamanne, line 339: This is the exact same code from this PR https://github.com/microsoft/vscode/pull/132298/files

davidanthoff (approved):
Awesome, thanks for taking care of it :)

PR #3780, opened 2025-02-25
Allow saving for Makie png/html plots and improve clipboard behavior

jkrumbiegel:
The "save plot" button was not working for Makie plots ever since Makie started to show them via html MIME types, as the png MIME type doesn't show them at the intended pixel density. This PR wraps html content in an id'ed div so the plot detection code can pick that up and extract a possible single `` tag. With that, the existing code works for the Makie case as well.

Also, I had trouble with the "copy plot to clipboard" functionality for a long time, so while I was at it, I looked into improving that as well. It seems that the focus check for clipboard access often fails because of a race condition inherent in this system, so apparently adding a small timeout improves that https://stackoverflow.com/questions/77465342/how-do-i-ensure-that-the-website-has-focus-so-the-copy-to-clipboard-can-happen. Anecdotally, this feels much better on my system. With the 0.05 timeout I seem to almost never get the "not focused" error while before I was getting that all the time. But would probably be good for others to test this, too, because the error is already nondeterministic, so the "fix" kind of is as well.

I didn't see these two functions in tests, but maybe I overlooked where the behavior is currently tested. If so, I could add tests there.


jkrumbiegel:
Anything I can do to move this one forward?

pfitzseb (approved):

PR #3804, opened 2025-04-21
Correctly highlight operator-style macros

MilesCranmer:
Fixes #3803 

Here's my testing which shows it correctly highlights operator-style macros, while still skipping invalid macros like `@123` and `@.+`

image

PR #3805, opened 2025-04-21
Create test suite for TextMate grammar

MilesCranmer:
I could very well be mistaken (?), but could not find any test suite for the syntax itself that is used for code highlighting. So I made one with a few basic checks. You can use it with

```
npm run verify-grammar
```

I also found an issue during testing: the composition operator was being parsed as `keyword.operator.arithmetic.julia` rather than  `keyword.operator.compose.julia`. So this fixes that too.

Should be merged after #3804 since it has some of the same commits (and includes a test for operator-style macros).


pfitzseb:
Feel free to add these to https://github.com/JuliaEditorSupport/atom-language-julia/blob/master/test/test.js, but I don't want to duplicate the tests here.

MilesCranmer:
Ah, there’s a test suite! I didn’t see it. Why is it in the atom repo? Why not centralise it here, since atom Julia is deprecated?

MilesCranmer:
Does that test suite run in the CI somehow?

pfitzseb:
> Why is it in the atom repo? Why not centralise it here, since atom Julia is deprecated?

That repo is the source of truth for the VS Code and GH syntax highlighters and predates the VS Code extension. In my mind it kinda makes sense to keep it separate, but I guess we should make it clearer that we pull the grammar in from there.

> Does that test suite run in the CI somehow?

Yes, it does.

MilesCranmer:
Since it sounds like the `julia_vscode.json` is an artifact rather than a source of truth, I don't think it should be tracked in version control. Perhaps you should just have a SHA pointing to the source grammar and generating the file during CI or release?

PR #3821, opened 2025-05-28
Fix @profview to correctly include interactive threads

### LINKED ISSUE #3820: @profview does not show all threads when using interactive threads
When calling @profview after starting Julia with `--threads="n,1"` - which will become the default meaning of `--threads=n` in 1.12 - the profile viewer incorrectly displays the interactive thread as thread 1, and then misses out the last worker thread (even in "all" view). That is:

## Steps to reproduce
1. Set `"julia.NumThreads": "n,1",` in `settings.json`, for the `n` of your choice (or `"auto,1"`).
2. Set up some multithreaded code (example given below)
3. Open the REPL using alt-J alt-O, and include it.
4. Run this code using `@profview`
<details>

<summary>Example code</summary>

```julia
using LinearAlgebra

function work()::Float64
    # Simulate some work, e.g., matrix multiplication
    A = rand(1000, 1000)
    B = rand(1000, 1000)
    C = A * B
    return norm(C)
end

function test_threads(n::Int)::Nothing

    # Workload for testing threading performance. Choose an embarassingly parallel task.
    results = Vector{Float64}(undef, n)
    Threads.@threads :greedy for i in 1:n
        results[i] = work()
    end

    @show results

    return nothing
end
```

</details>

## Expected results
I see profiling results for all n+1 threads

## Actual results
I see profiling results for the interactive thread and first n-1 worker threads. This last thread is also not included in the "all" view.

## Further details

The main errors would seem to be in these lines of code:

https://github.com/julia-vscode/julia-vscode/blob/7dd2799f1723e9fe0120a707468321a9c4fb017b/scripts/packages/VSCodeServer/src/profiler.jl#L16

https://github.com/julia-vscode/julia-vscode/blob/7dd2799f1723e9fe0120a707468321a9c4fb017b/scripts/packages/VSCodeServer/src/profiler.jl#L42

`Threads.nthreads()` gives the number of *worker* threads only; in the presence of interactive threads (again, the default in 1.12), this will give the incorrect count.

This should be a relatively simple fix. In fact, making the edit:

```julia
Threads.nthreads() => (Threads.nthreads(:interactive)+Threads.nthreads(:default))
```

seems to work perfectly on my machine.

jaredjeya:
@profview does not show information from all worker threads because it fails to count the total number of threads correctly, incorrectly only counting the number of worker threads. However, since Julia orders interactive threads before worker threads, we miss out on the last worker thread (assuming one interactive thread). Since Julia will by default have 1 interactive thread starting with v1.12, it's important to fix this before that releases.

Fixes #3820.

I've made the minimal possible change here and tested that this works on my own machine. I've also made a small GUI change: the dropdown on the profiler now lists "All", "Interactive: x", "Worker: y" instead of just numbering the threads. I've made the choice that worker threads start at 1, instead of `nthreads(:interactive) + 1`, but this can of course be changed.

I don't think this requires a documentation update, it should be self-explanatory.

### DIFF
--- a/scripts/packages/VSCodeServer/src/profiler.jl
+++ b/scripts/packages/VSCodeServer/src/profiler.jl
@@ -27,7 +27,14 @@ function view_profile(data = Profile.fetch(); C=false, kwargs...)
     data_u64 = convert(Vector{UInt64}, data)
     for thread in threads
         graph = stackframetree(data_u64, lidict; thread=thread, kwargs...)
-        d[string(thread)] = make_tree(
+        threadname = if thread == "all"
+            "All"
+        elseif thread <= Threads.nthreads(:interactive)
+            "Interactive: $(thread)"
+        else
+            "Worker: $(thread - Threads.nthreads(:interactive))"
+        end
+        d[threadname] = make_tree(
             ProfileFrame(
                 "root", "", "", 0, graph.count, missing, 0x0, missing, ProfileFrame[]
             ), graph; C=C)
--- a/scripts/packages/VSCodeServer/src/profiler.jl
+++ b/scripts/packages/VSCodeServer/src/profiler.jl
@@ -13,7 +13,7 @@ function view_profile(data = Profile.fetch(); C=false, kwargs...)
     d = Dict{String,ProfileFrame}()
 
     if VERSION >= v"1.8.0-DEV.460"
-        threads = ["all", 1:Threads.nthreads()...]
+        threads = ["all", 1:(Threads.nthreads(:interactive)+Threads.nthreads(:default))...]
     else
         threads = ["all"]
     end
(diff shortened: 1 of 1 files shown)


jaredjeya:
I've updated this to make no assumptions about thread ordering and to be a little more robust. Note also that I've not included the foreign threadpool (this was never included, to be clear). In the future there could be a keyword argument to the macro that decides whether to also include foreign threads?

Also, it seems like changing the name "all" to "All threads" (or in fact, anything but "all") means that option gets sorted last in the profile viewer.  That's particularly annoying as it no longer gets selected by default. I'm going to assume there's some code in there that grabs it and puts it at the top? So you might want to change it back to "all" before merging this - OR, change the code in whatever actually displays the profile to match. Unfortunately I can't figure out where that is!

pfitzseb:
> Also, it seems like changing the name "all" to "All threads" (or in fact, anything but "all") means that option gets sorted last in the profile viewer. That's particularly annoying as it no longer gets selected by default. I'm going to assume there's some code in there that grabs it and puts it at the top?

That's at https://github.com/pfitzseb/jl-profile.js/blob/a13f2ef7852bc1782ec180e3efd4ccef2da7ba6d/profile-viewer.ts#L146C17-L151. Would be nice if you could just change it back to `all` for now so we can get the PR merged sooner rather than later.

jaredjeya:
Ah, well that explains why nothing I changed in `src/interactive/profiler.ts` made any difference to which one got selected! I've changed it back now.

PR #3822, opened 2025-05-29
feat: support variable substitution for `environmentPath` setting

mtiller-jh:
I was mainly concerned with supporting `${workspaceFolder}`, but you can find the substitution rules for many other patterns here:

https://github.com/default-writer/filetestonsave/blob/main/src/extension.ts

Note, the tooltip help for the `environmentPath` setting already states that `${workspaceFolder}` is supported
which is **not** the case.  So this at least makes this tooltip (shown below) slightly more honest:

image

### DIFF
--- a/src/jlpkgenv.ts
+++ b/src/jlpkgenv.ts
@@ -20,6 +20,14 @@ let g_resolved_path_of_environment: string = null
 
 let g_juliaExecutablesFeature: JuliaExecutablesFeature = null
 
+function getEnvironmentPathConfig() {
+    const section = vscode.workspace.getConfiguration('julia')
+    const rawConfigValue = section.get<string>('environmentPath')
+    const workspaceFolderPath =
+        vscode.workspace.workspaceFolders?.[0]?.uri?.fsPath // Use the first workspace folder if it exists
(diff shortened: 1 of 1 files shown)


src/jlpkgenv.ts
```diff
@@ -20,6 +20,14 @@ let g_resolved_path_of_environment: string = null
 
 let g_juliaExecutablesFeature: JuliaExecutablesFeature = null
 
+function getEnvironmentPathConfig() {
+    const section = vscode.workspace.getConfiguration('julia')
+    const rawConfigValue = section.get<string>('environmentPath')
+    const workspaceFolderPath =
+        vscode.workspace.workspaceFolders?.[0]?.uri?.fsPath // Use the first workspace folder if it exists
+    return rawConfigValue.replace('${workspaceFolder}', workspaceFolderPath)
```
pfitzseb, line 28: Won't this put a literal `undefined` into the path if no workspace folders are open? That seems like pretty bad fallback behavior. Would be better to explicitly return a `null` here and check the return value at both call sites.

mtiller-jh: Actually, I just realized that the extension already had a function for doing this.  So I just changed my PR so that I use the existing function.

pfitzseb (approved):

PR #3852, opened 2025-07-22
Highlight docstring in .jl files using "string.docstring.julia"

quachpas:
Hello!

This is a rehash of [juliacodeblock.json](https://github.com/julia-vscode/julia-vscode/blob/main/syntaxes/juliacodeblock.json) but injecting in `L:string.docstring.julia`.

From my limited testing, it seems to work correctly with no leaks to the main source. Obviously, nested docstring in the codeblock does not work, in that case the syntax highlighting seems to fail completely.

image

If there is any additional testing I can do on my side, I'll be glad to do so!

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,13 @@
                 "embeddedLanguages": {
                     "meta.embedded.block.julia": "julia"
                 }
+            },
+            {
+                "scopeName": "source.docstring.julia",
+                "path": "./syntaxes/juliadocstring.json",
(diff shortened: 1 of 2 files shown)


pfitzseb:
LGTM. Can you add a mention to the changelog please? :)

quachpas:
@pfitzseb done! Thank you

pfitzseb (approved):

PR #3878, opened 2025-10-08
feat(api): added api to get `juliaup` executable

Hetarth02:
This PR aims to add a way to fetch the information regarding `juliaup` installation on the user's machine.


Hetarth02:
I don't know if this requires documentation or not.

PR #3902, opened 2025-11-17
feat: added option to select specific version for julia repl

Hetarth02:

### DIFF
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -133,7 +133,11 @@ function parseSessionArgs(name: string) {
 }
 
 // FIXME: refactor this!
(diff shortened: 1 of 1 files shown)


src/interactive/repl.ts
```diff
@@ -276,6 +311,46 @@ async function startREPL(preserveFocus: boolean, showTerminal: boolean = true) {
...
+        title: 'Start REPL with specific version',
+    })
+
+    if (select) {
+        await startREPL(false, true, select.Name)
+    }
+}
+
+async function getInstalledJuliaVersions() {
+    const juliaup = await g_juliaExecutablesFeature.getActiveJuliaupExecutableAsync()
```
pfitzseb: What happens if juliaup is not installed? Do we show an error message in that case?

PR #3914, opened 2025-11-24
feat: restart language server command now takes an optional `envPath` arg

Hetarth02:


src/extension.ts
```diff
@@ -337,6 +337,9 @@ async function startLanguageServer(juliaExecutablesFeature: JuliaExecutablesFeat
     }
 
     let jlEnvPath = ''
+    if (envPath && envPath !== '') {
```
pfitzseb: Let's just do `if (envPath)` here and below. Otherwise we have this half-baked input validation that accepts some invalid paths but not others.

PR #3926, opened 2025-12-10
feat: add suggestion to install revise if it is not present

Hetarth02:
When you execute `Julia: Start REPL` or `Julia: Start REPL with specific version` it will suggest to install & setup `Revise.jl` if it is not present and `Use Revise` setting is `true`.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Fix a bug that caused Julia test detection to fail ([#3935](https://github.com/julia-vscode/julia-vscode/pull/3935))
 
+### Added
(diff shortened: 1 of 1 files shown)


src/interactive/repl.ts
```diff
@@ -1426,7 +1427,68 @@ export function activate(
...
-        ),
+
+            connection.onNotification(notifyTypeCheckRevise, (hasRevise: boolean) => {
+                const config = vscode.workspace.getConfiguration('julia')
+                const useRevise = config.get('useRevise')
+
+                if (useRevise && !hasRevise) {
+                    const install = 'Install & Setup Revise'
+                    const turnOff = 'Turn off'
+                    const turnOffGlobally = 'Turn off globally'
```
pfitzseb:
```suggestion
                    const turnOff = 'Disable (workspace)'
                    const turnOffGlobally = 'Disable'
```

PR #4000, opened 2026-01-22
Custom pane title

jkrumbiegel:
The custom pane feature is useful, but a little limited by the `id` being used for the title of the pane as well. We'd like to show fit progress information in these panes and because this runs in the background, the pane title is a very useful thing that would be a bit wasted on a static id.

In order to support this, I had to parse multiple parameters correctly from the MIME string in unquoted or quoted style (complex titles might contain double quotes, backslashes, semicolons). I used https://www.rfc-editor.org/rfc/rfc9110?utm_source=chatgpt.com#section-5.6.6 as a reference and Claude helped me translate that into a parsing function.

Tests were written completely by Claude but looked sensible to me.

Here's an example, I didn't use a plot because I didn't have Makie loaded but the principle is the same:

https://github.com/user-attachments/assets/363b7996-3b37-4376-be75-d26802e8841e

### DIFF
--- a/src/interactive/repl.ts
+++ b/src/interactive/repl.ts
@@ -323,7 +323,7 @@ function makeTerminalName(juliaExecutable: JuliaExecutable) {
     const channelName = juliaExecutable.juliaupChannel?.name
     let name = `Julia REPL (v${version}`
 
-    if (channelName && version.includes(channelName)) {
+    if (channelName === undefined || version.includes(channelName)) {
         name += ')'
     } else {
         name += `, +${channelName})`


scripts/packages/VSCodeServer/src/display.jl
```diff
@@ -63,44 +75,184 @@ function with_no_default_display(f; allow_inline = false)
...
+    # parameter = token "=" ( token / quoted-string )
+    # token = 1*tchar where tchar excludes delimiters
+    # quoted-string = DQUOTE *( qdtext / quoted-pair ) DQUOTE
+
+    params = Dict{String, String}()
+    pos = 1
+
+    # tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." /
+    #         "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
+    token_re = r"[!#$%&'*+\-.0-9A-Z^-z|~]+"
```
pfitzseb: Shouldn't we allow unicode here? Not sure if it's RFC compliant, but might be useful. If not, would be good to specify "alphanumeric ASCII" in the docstring.

pfitzseb: Regardless of the decision here, the tests should either check that non-ASCII mime types get rejected or handle properly.

jkrumbiegel: I'm allowing unicode now only in the quoted string style. For unquoted tokens I find it too annoying to handle weird whitespace characters that would definitely look wrong etc. Added to the docstring and added tests.

Issue #2894, opened 2022-05-21
Showing cell headers in outline


artemsolod:
Is it possible to display some sort of section information in Outline pane? I've tried using cell syntax but it seems to be treated just like a comment and ignored. Can support for this be added within `julia-vscode` or is it upstream territory? Or may be there is syntax to make some sort of anchors in outline? Thanks!

![image](https://user-images.githubusercontent.com/1939268/169646765-51f0338e-2518-4dba-a192-141f1836548d.png)

donboyd5:
Just seconding this - would love to see it.

kylebutts:
I agree that this would be super useful! It also would enable collapsing code chunks :-)

pmottishaw:
This would be extremely useful. It is difficult to navigate long files.

dsanchezp18:
Also would be helpful!!

jared711:
Ditto!

kellertuer:
This would indeed be very nice.

santilopezmoskovits:
Yes! This is very much needed

PR #4003, opened 2026-01-23
feat: contents outline option

artemsolod:
Fixes https://github.com/julia-vscode/julia-vscode/issues/2894. This is mostly written with an LLM though I've reviewed and tested it.

Adds an opt-in Contents outline for markdown-style section headers. `julia.outline.contents` is false by default.

Results in outline pane having 2 sections: one for symbols and one for contents.
image

### DIFF
--- a/src/interactive/codecells.ts
+++ b/src/interactive/codecells.ts
@@ -677,7 +677,10 @@ class CodeCellExecutionFeature extends JuliaCellManager {
     }
 }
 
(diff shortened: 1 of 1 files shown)


huangyxi:
Since you’re already good at leveraging LLMs for coding tasks, maybe we could try prompting the LLM to extract headers from `buildDocCells` in `JuliaCellManager` using a slightly modified `JuliaCell`. With that in place, we could then implement `vscode.DocumentSymbolProvider` inside `CodeCellFeature`.

PR #4033, opened 2026-02-19
fix: install correct winget package

pfitzseb:


davidanthoff:
I don't think this will work, the winget community registration for Juliaup points to the appx installer (which we already try as the second option), see https://github.com/microsoft/winget-pkgs/blob/master/manifests/j/Julialang/Juliaup/1.19.9.0/Julialang.Juliaup.installer.yaml.

PR #4082, opened 2026-04-16
fix: defer test feature init if julia not available

pfitzseb:


dr14-make:
I can confirm works on fresh linux

davidanthoff (approved):

PR #4085, opened 2026-04-18
Fix a small strict error

davidanthoff:
This fixes a small error in TypeScript strict mode.


pfitzseb:
This looks wrong. We can just use `private exitCode: number | void = undefined` here.

PR #4088, opened 2026-04-20
Handle LS errors better

davidanthoff:
My hope is that this will reduce unnecessary telemetry crash reports by a lot.

### DIFF
--- a/src/languageClient.ts
+++ b/src/languageClient.ts
@@ -4,7 +4,15 @@ import * as net from 'net'
 import * as os from 'os'
 import * as path from 'path'
 import * as vscode from 'vscode'
-import { LanguageClient, LanguageClientOptions, RevealOutputChannelOn, ServerOptions } from 'vscode-languageclient/node'
+import {
+    LanguageClient,
+    LanguageClientOptions,
+    RevealOutputChannelOn,
+    ServerOptions,
+    State,
+    StateChangeEvent,
+} from 'vscode-languageclient/node'
+import { ErrorCodes, LSPErrorCodes, ResponseError } from 'vscode-languageserver-protocol'
 
 import * as jlpkgenv from './jlpkgenv'
 import * as telemetry from './telemetry'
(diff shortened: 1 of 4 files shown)


src/languageClient.ts
```diff
@@ -302,19 +387,22 @@ export class LanguageClientFeature {
...
+            if (!isLanguageServerError(err)) {
+                vscode.window.showErrorMessage('Failed to refresh the language server cache.', {
+                    detail: err,
+                })
+            }
         }
     }
 
     async restartLanguageServer(envPath?: string, autoInstall?: boolean) {
+        this._intentionalStop = true
```
pfitzseb: Can we factor this out into a `stop` method that can be used here and in `dispose` below?

src/languageClient.ts
```diff
@@ -30,8 +68,33 @@ export class LanguageClientFeature {
...
+            this._onDidChangeStateEmitter.fire(state)
+            this.updateStatusBarForState()
+        }
+    }
+
+    private updateStatusBarForState() {
+        if (this._state === 'crashed') {
+            this.statusBarItem.text = '$(warning) Julia Language Server Crashed'
+            this.statusBarItem.backgroundColor = new vscode.ThemeColor('statusBarItem.warningBackground')
+            this.statusBarItem.command = 'language-julia.restartLanguageServer'
```
pfitzseb, line 92: This needs to be reset to `undefined` in `startServerInner`.

src/languageClient.ts
```diff
@@ -63,17 +126,18 @@ export class LanguageClientFeature {
...
+            return callbackOnHandledErr ? callbackOnHandledErr(err) : undefined
         }
 
         try {
             return await callback(this.languageClient)
         } catch (err) {
-            if (err.message === 'Language client is not ready yet') {
-                return callbackOnHandledErr(err)
+            if (isLanguageServerError(err)) {
+                return callbackOnHandledErr ? callbackOnHandledErr(err) : undefined
```
pfitzseb, line 153: Should maybe `await` these as well for consistency's sake, even though it doesn't matter for our current use cases.

davidanthoff:
@pfitzseb Ok, I think comments are addressed?

pfitzseb (approved):

PR #4089, opened 2026-04-22
Fix telemetry bugs

davidanthoff:
Three fixes:

One trivial try catch to surpress an error.

And then a lot of cleanup around `async` wrapping. A lot of these cases were ok before because functions happened to return a promise (even though not marked as `async`), but this makes it all more explicit, and there were at least a handful of cases were we ended up with unhandled promises also.

And I also narrowed the telemetry handler back to what it originally was, we had a crash report about the new form.

### DIFF
--- a/src/docbrowser/documentation.ts
+++ b/src/docbrowser/documentation.ts
@@ -65,11 +65,13 @@ export function activate(context: vscode.ExtensionContext, languageClientFeature
     const provider = new DocumentationViewProvider(context, languageClientFeature)
 
     context.subscriptions.push(
-        registerCommand('language-julia.show-documentation-pane', () => provider.showDocumentationPane()),
-        registerCommand('language-julia.show-documentation', () => provider.showDocumentation()),
-        registerCommand('language-julia.browse-back-documentation', () => provider.browseBack()),
-        registerCommand('language-julia.browse-forward-documentation', () => provider.browseForward()),
-        registerCommand('language-julia.search-word', (params) => provider.findHelp(params)),
+        registerCommand('language-julia.show-documentation-pane', async () => await provider.showDocumentationPane()),
+        registerCommand('language-julia.show-documentation', async () => await provider.showDocumentation()),
+        registerCommand('language-julia.browse-back-documentation', async () => provider.browseBack()),
+        registerCommand('language-julia.browse-forward-documentation', async () => provider.browseForward()),
+        registerCommand('language-julia.search-word', async (params: { searchTerm: string }) =>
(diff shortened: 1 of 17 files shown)


src/docbrowser/documentation.ts
```diff
@@ -65,11 +65,13 @@ export function activate(context: vscode.ExtensionContext, languageClientFeature
     const provider = new DocumentationViewProvider(context, languageClientFeature)
 
     context.subscriptions.push(
-        registerCommand('language-julia.show-documentation-pane', () => provider.showDocumentationPane()),
-        registerCommand('language-julia.show-documentation', () => provider.showDocumentation()),
-        registerCommand('language-julia.browse-back-documentation', () => provider.browseBack()),
-        registerCommand('language-julia.browse-forward-documentation', () => provider.browseForward()),
-        registerCommand('language-julia.search-word', (params) => provider.findHelp(params)),
+        registerCommand('language-julia.show-documentation-pane', async () => await provider.showDocumentationPane()),
```
pfitzseb, line 68: Isn't this essentially a no-op? From the outside you can await the result of both exactly the same.

src/utils.ts
```diff
@@ -65,21 +65,27 @@ export function inferJuliaNumThreads(): string | undefined {
...
         try {
-            const result = f(...args)
-            if (result && typeof result.then === 'function') {
-                return result.then(undefined, (err) => {
-                    handleNewCrashReportFromException(err, 'Extension')
-                    throw err
-                })
-            }
-            return result
+            return await f(...args)
```
pfitzseb, line 75: I think we can get away with a single definition for both sync and async functions with something like
```
try {
  const res = f(...args)
  if (res.catch) {
    res.catch(err => handleNewCrashReportFromException(err, 'Extension'))
  }
  return res
} catch (err) {
  handleNewCrashReportFromException(err, 'Extension')
  throw err
}
```

src/docbrowser/documentation.ts
davidanthoff: Yeah, I just find it easier to understand what is going on?

src/utils.ts
davidanthoff: I think that is similar to what we tried before, right? But maybe that previous attempt was just not very correct...

To me the separate functions are just easier to understand.

pfitzseb: Oh, yeah, I guess I just didn't look at the left side of the diff earlier :D

pfitzseb (approved):

PR #4091, opened 2026-04-23
Handle Julia not found situations better

davidanthoff:
Right now a lot of exceptions bubble through to our telemetry crash reporting that shouldn't.

@pfitzseb would be good if you could review this carefully, it was done largely by Claude and I don't have a good sense of the current breadth and scope of that area of the extension :)

### DIFF
--- a/src/debugger/debugFeature.ts
+++ b/src/debugger/debugFeature.ts
@@ -1,6 +1,6 @@
 import * as vscode from 'vscode'
 import * as jlpkgenv from '../jlpkgenv'
-import { ExecutableFeature } from '../executables'
+import { ExecutableFeature, JuliaNotFoundError } from '../executables'
 import {
     generatePipeName,
     getCustomEnvironmentVariables,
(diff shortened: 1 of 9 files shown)


pfitzseb (commented):
Adding a new error type to executables.ts makes a lot of sense to me, but imho it would be cleaner to filter this error type out in `handleNewCrashReportFromException`; that way we don't have to annotate every call site. 

However, there's also some value in not throwing exceptions everywhere, so I'd like to add the deferred feature init pattern from https://github.com/julia-vscode/julia-vscode/pull/4082 wherever it makes sense. That should probably happen in a follow up PR though so we can get crash reports clean sooner rather than later.

davidanthoff:
Yeah, I really like the approach you used for the test feature, agreed we should adopt that everywhere. I think if we merge this PR as is we would already know all the places where we need to do that, Claude tried to find all call sites that could receive the exception. I guess filtering it out in the telemetry stage would have been easier, but on the other hand now we at least know all the places where this could be a problem?

pfitzseb (approved):

PR #4100, opened 2026-05-06
Compat Julia 1.8: @profview calls undefined threadpooltids and threadpool

cihga39871:
Fixes `@profview` error in Julia v1.8.5: threadpooltids and threadpool are undefined but called. Add those functions in `__init__`

### DIFF
--- a/scripts/packages/VSCodeServer/src/VSCodeServer.jl
+++ b/scripts/packages/VSCodeServer/src/VSCodeServer.jl
@@ -274,31 +274,4 @@ end
 
 _precompile_()
 
-function __init__()
-    # patch at runtime to avoid precompilation issues
-    @static if v"1.8.0-DEV.460" <= VERSION < v"1.9"
-        # Compat: @profview calls threadpooltids and threadpool, but they are not defined in Julia v1.8
-        if !isdefined(Threads, :threadpooltids)
-            Core.eval(Base.Threads, :(function threadpooltids(x)
-                # Get the thread IDs in the current thread pool
-                if x === :default
-                    return Threads.nthreads() == 1 ? [1] : collect(1:Threads.nthreads())
-                else
-                    return Int[]
-                end
-            end))
-        end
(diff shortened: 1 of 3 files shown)


scripts/packages/VSCodeServer/src/VSCodeServer.jl
```diff
@@ -274,4 +274,31 @@ end
 
 _precompile_()
 
+function __init__()
+    # patch at runtime to avoid precompilation issues
+    @static if v"1.8.0-DEV.460" <= VERSION < v"1.9"
```
pfitzseb: This is the wrong fix imo. We should instead add another case at the callsite and fall back to the [previous implementation](https://github.com/julia-vscode/julia-vscode/commit/e2cce14886303ce6ff06baa183471ca1a4ceb3d6) for 1.8.x releases.

pfitzseb (changes requested):

scripts/packages/VSCodeServer/src/VSCodeServer.jl
cihga39871: unfortunitely the previous implementation is still not working on 1.8.5. I don't know why julia-vscode think julia 1.8 has the concept of interactive and default thread pools. I tried to change `v"1.8.0-DEV.460"` in profiler.jl but it still fails.

pfitzseb: I meant the previous implementation which doesn't use thread pools at all:
```
threads = ["all", 1:(Threads.nthreads(:interactive)+Threads.nthreads(:default))...]
```

cihga39871:
Threads.nthreads(::Symbol) is still undefined


________________________________

pfitzseb:
Then we should simply use `Threads.nthreads()` I think.

cihga39871:
Thank you for your time. The new fix uses similar statements as previous implementations.

pfitzseb (approved):

PR #4123, opened 2026-06-17
fix: add timeout to Julia availability/version checks so activation can't hang indefinitely

### LINKED ISSUE #4119: Set timeout for julia availability check
Ref https://discourse.julialang.org/t/juliaup-configuration-is-locked-by-another-process-waiting-for-it-to-unlock/121366.

This can cause extension activation to hang indefinitely. While the actual issue is handled upstream now, it seems reasonable to add a timeout to our checks as well.

mvanhorn:
Fixes #4119

- [ ] End-user documentation check. This change is internal hardening with no user-facing behavior to document, so no docs update is required.

## Summary

The extension probes for a usable Julia executable during activation by running `julia --version` and related `juliaup` commands, awaiting each child process with no timeout. Several of these `execFile` probes in `src/executables.ts` can block indefinitely if the child never returns.

This adds a conservative timeout (plus `killSignal`) to those bare availability probes, defined as a single named constant `JULIA_PROBE_TIMEOUT_MS` so it is easy to tune:

- `JuliaupExecutable.run` (the `juliaup` invocation)
- `JuliaExecutable.rootFolder` (the `Sys.BINDIR` probe)
- `ExecutableFeature.tryGetJuliaVersion` (the `--version` probe)
- `ExecutableFeature.getJuliaupExecutableNoCache` (the `juliaup --version` discovery probe)

On timeout `execFile` kills the child and rejects with an `ETIMEDOUT`-style error, which the existing `catch` blocks already translate into the graceful "executable not available / failed to run" path, so no new control flow is required and a genuinely missing or erroring executable still reports unavailable with its real error.

## Why this matters

As described in #4119, a probe can hang forever, for example when a `juliaup` call waits on a configuration lock held by another process (see the linked Discourse thread). With no bound on the probe, extension activation hangs indefinitely with no recovery. The underlying lock case is handled upstream, but bounding the extension's own availability checks makes activation resilient to any future probe that blocks.

## Testing

- `npm run check-types` (tsc --noEmit) passes.
- `eslint --ext .ts src/executables.ts` passes.
- `prettier --check src/executables.ts` passes.
- A fast-responding `julia --version` still resolves to the parsed version well under the timeout (no behavior change); a probe exceeding the timeout is killed and surfaces through the existing "failed to run" branch instead of hanging.


src/executables.ts
```diff
@@ -15,6 +15,16 @@ const execFile = promisify(child_process.execFile)
 
 const juliaVersionPrefix = 'julia version '
 
+/**
+ * Upper bound (ms) for the short-lived `julia --version` / juliaup availability probes
+ * in this file. These probes run during activation; without a bound a child process that
+ * blocks indefinitely (e.g. a juliaup call waiting on a configuration lock held by another
+ * process) would hang extension activation forever. On timeout `execFile` kills the child
+ * and rejects with an ETIMEDOUT-style error, which the existing `catch` blocks already
+ * translate into the graceful "executable not available / failed to run" path.
```
pfitzseb: Imho it's a bit misleading if this runs into the exact same error handling as when there actually is no executable. Can you add a branch that checks the error type and shows a different message to the user?

PR #4141, opened 2026-07-16
Use the configured Julia executable for test items instead of hardcoding the release channel

dr14-make:
## What this does

`JuliaTestController.start()` currently has two paths for picking the Julia executable that runs the test item controller:

- With `DEBUG_MODE` set, it calls `executableFeature.getExecutable()`, which respects the user's configuration (`julia.executablePath`, then the selected juliaup channel, then the juliaup default, then `julia` on the `PATH`).
- In the normal path, it requires juliaup and then hardcodes `new JuliaExecutable(await juliaup.getChannel('release'))`, ignoring the configured executable entirely and refusing to run unless a `release` channel exists.

This PR unifies both paths on `getExecutable()`, keeping the existing `JuliaNotFoundError` handling. The `hasJuliaup()` check, the `getJuliaupExecutable` call, the `getChannel('release')` special case, and their two error message prompts become dead code and are removed.

## Why

The test runner should use the same Julia as the rest of the extension. The hardcoded `release` channel breaks `@testitem` running for anyone whose Julia isn't installed as a juliaup `release` channel: a custom `julia.executablePath`, a non-`release` channel selection, or a bundled/offline juliaup setup that has no `release` channel at all. `getExecutable()` already implements a robust fallback chain, so the special case is redundant.

## Behavior change

Previously, test items always ran on the juliaup `release` channel regardless of configuration. With this change they run on whatever Julia the user has configured/selected, same as the REPL and language server. If pinning test runs to `release` was intentional (e.g. to guarantee a controller-compatible Julia version), happy to adjust the approach — feedback welcome.

### DIFF
--- a/src/testing/testFeature.ts
+++ b/src/testing/testFeature.ts
@@ -120,41 +120,13 @@ export class JuliaTestController {
 
         let juliaExecutable: JuliaExecutable | null
 
-        if (process.env.DEBUG_MODE) {
-            try {
-                juliaExecutable = await this.executableFeature.getExecutable()
-            } catch (err) {
-                if (err instanceof JuliaNotFoundError) {
-                    return true
-                }
-                throw err
-            }
-        } else {
-            if (!(await this.executableFeature.hasJuliaup())) {
-                vscode.window.showErrorMessage(
-                    'You must install Julia using Juliaup to use the test item functionality.'
-                )
+        try {
+            juliaExecutable = await this.executableFeature.getExecutable()
+        } catch (err) {
+            if (err instanceof JuliaNotFoundError) {
                 return true
             }
-
-            let juliaup
-            try {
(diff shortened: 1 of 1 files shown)


pfitzseb:
Needs a changelog entry, looks good otherwise. Note that this won't be released soon since main is currently very unstable.

davidanthoff:
No, this is not the right approach :) The test item controller should always run on the release channel.

There is already extensive infrastructure for using different Julia versions for the test processes themselves (which run the test items). Currently that is exposed via the UI (one can pick a Juliaup channel on which the tests should run). We could be more clever in auto picking Julia versions based on manifests etc but there are some subtleties that I am still thinking through.

dr14-make:
Sure, @davidanthoff  that's why I've asked, I am ok with any possible solution which will allow the override of hardcoded 'release' channel, any pointers how u see it can be done in correlation to your design env var or something?

PR #4158, opened 2026-08-05
Add per-button cell CodeLens toggles

moyner:
Attempt at fixing #4047 by adding separate toggles for each codelens item.


pfitzseb (approved):

pfitzseb:
Thanks!

PR #4168, opened 2026-08-17
Load bundled packages from their unmodified entry files

fredrikekre:
This adds `include_with_private_deps(mod, path)`, which includes a package's regular entry file into `mod` by stripping the top level `module` wrapper, rewriting `using`/`import` of bundled dependencies to relative imports (so they resolve to the private copies instead), and applying the same rewriting to nested `include`s. This removes the need for packages to adopt the `packagedef.jl` file structure. The second commit converts all bundled packages in VSCodeServer and VSCodeDebugger to this mechanism.

Closes fredrikekre/Runic.jl#209.

### DIFF
--- a/scripts/include_with_private_deps.jl
+++ b/scripts/include_with_private_deps.jl
@@ -1,0 +1,91 @@
+# Include the unmodified entry file of a bundled package (e.g. `Foo/src/Foo.jl`) into
+# `mod`, stripping the top level `module Foo ... end` wrapper and rewriting
+# `using`/`import` statements of bundled dependencies to relative imports so they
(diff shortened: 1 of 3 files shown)


davidanthoff:
This would break the entire analysis stack of the LS, it can't follow these kind of dynamic include things...

fredrikekre:
Why does it need to follow internal deps?

davidanthoff:
We are dogfooding our own tools, i.e. we want to be able to use the LS on the code that powers the extension, and that means there are certain dynamic things we can't do in the extension Julia code because the static analysis engine wouldn't be able to follow those includes and then none of the features in JuliaWorkspaces.jl work.

PR #4219, opened 2026-09-17
Stop reporting an ordinary controller stop as a crash

davidanthoff:
Crash report **c28**, seen on insider 1.243.1 — `Julia test item controller exited with code none, signal SIGTERM`, 3 events from 1 user on 2026-09-16.

The test item controller's exit handler reported every exit that carried a signal. `kill()` and the timeout fallback in `shutdown()` both call `ChildProcess.kill()` with its default `SIGTERM`, so the extension stopping *its own* controller — closing VS Code, using the kill action in the process view — arrived here as `signal SIGTERM` and filed a crash report with no crash in it.

[#4217](https://github.com/julia-vscode/julia-vscode/pull/4217) (c26) gave the language server `isOsKillSignal` for exactly this, but the controller's copy of the logic never got it. The exit decision now lives in `unexpectedControllerExit`, the counterpart of `unexpectedServerExit`, and applies the same rule:

- `SIGTERM`/`SIGKILL` are a stop, not a crash — whether they come from us or from an out-of-memory killer.
- Native crash signals such as `SIGSEGV` are still reported.
- Code 1 stays excluded, because the controller's own crash logger has already reported it with the right cloud role.
- The Windows session-teardown codes stay excluded, as before.

Covered by seven new cases in `src/test/suite/testFeature.test.ts`.

Found by reviewing crash telemetry; opened by Claude.


davidanthoff:
David asked whether this was signature-only, and whether that risks masking kills we do want to hear about. It was, and it did — pushed a second commit that fixes it.

As originally written this decided on the exit signature alone, so an unasked-for `SIGTERM`/`SIGKILL` was indistinguishable from the extension stopping its own controller. That is strictly weaker than the language server, which has had `_intentionalStop` all along and only gained `isOsKillSignal` on top of it in #4217. The case that would have been lost is an out-of-memory kill of the controller — the likeliest way a test run dies, since the runs are what use the memory.

`JuliaTestController` now records `_intentionalStop` while it is being stopped, set by `kill()` and `shutdown()` and cleared in `start()`, and `unexpectedControllerExit` takes it as its first check — the same shape as `unexpectedServerExit`.

A kill we did not ask for is still not filed as a crash report, on #4217's reasoning that it describes the machine rather than a fault here. But it is no longer silent: it is counted as a `ticoskill` usage event with the same memory figures as the server's `lsoskill`, which is what tells an out-of-memory kill apart from a container stop.

Tests updated for the new signature, with cases pinning both the intentional-stop path and the unasked-for kill.

davidanthoff:
> A OOM for the TIC process should definitely be reported to the user and presumably also in telemetry, it might be because we have a memory leak...

Agreed, and that undercuts the reasoning I inherited from #4217. That PR's argument for the language server was that an OS kill "describes the machine rather than a fault here" — but the controller and the test processes it supervises are ours, and the test runs are what use the memory. A run that grows until the kernel steps in is a plausible leak on our side, so filtering it out is filtering out the evidence.

Third commit changes a kill nobody here asked for from *counted* to *reported*:

**To the user.** An error notification naming the signal, saying the likely cause is memory, with a `Show Logs` button. Until now the tests simply stopped and the Test Explorer went quiet — the runs were ended and the output channel got a one-line exit note, but nothing told the user why.

**To the output channel**, so the detail survives the notification.

**To crash telemetry**, as `TestItemControllerOsKill` under the `Test Item Controller` cloud role, carrying what tells a leak apart from a short machine:

```
Julia test item controller was killed with SIGKILL
Memory: 256.0 MiB free of 16.0 GiB, cgroup limit none
Test processes alive: 7
```

The live test process count is the leak-relevant part — a controller dying with a handful of processes on a big machine is a very different story from one dying at a cgroup ceiling. The `ticoskill` usage event stays, now with the process count too, so the rate is still countable separately from the reports.

`unexpectedControllerExit` still returns `null` for these signals, but that is now only what stops the same death being reported twice; the dedicated report strictly supersedes the bare exit line.

`osKillReport` is a pure function with tests, including one asserting it carries no paths or package names — only numbers and the signal.

One thing I deliberately did **not** do: the equivalent change for the language server in #4217. The same argument may well apply there — an OOM kill of the LS could equally be our leak — but that is a merged decision of yours and a separate PR, so I left it alone rather than reversing it here. Say the word if you want it revisited.

